PR #16, opened 2019-12-12
Welfare module

johanneskoch94:
I compared the results before/after the refactoring and there are slight differences in the prices (see attachment). Is this problematic?
[compScen-2019-12-12_16.01.04.pdf](https://github.com/remindmodel/remind/files/3956462/compScen-2019-12-12_16.01.04.pdf)

### DIFF
--- a/config/default.cfg
+++ b/config/default.cfg
@@ -13,7 +13,7 @@ cfg <- list()
 #### Main settings ####
 
 #short description of the actual run
-cfg$title <- "default"
+cfg$title <- "welfare_mod_namechange"
 
 # path to the submodel to be used relative to main model folder (e.g."standalone/template.gms")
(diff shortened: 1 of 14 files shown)


johanneskoch94:
I think it looks that way because of the scale... the global difference is very small compared to the order of magnitude in the regions.

I haven't run any policy scenarios yet. Gonna run some now.

giannou:
> I think it looks that way because of the scale... the global difference is very small compared to the order of magnitude in the regions.

of course, sorry for not noticing

johanneskoch94:
Here's a Budget950 run comparison. I think it looks good.
[Budg950_vs_welf_mod_Budg950.pdf](https://github.com/remindmodel/remind/files/3972241/Budg950_vs_welf_mod_Budg950.pdf)

PR #28, opened 2020-01-08
add to model output tutorial

fschreyer:
added to and restructured model output tutorial


fschreyer:
Ok. Deleted cluster mounting section.

PR #33, opened 2020-01-15
Added input folders of new modules

johanneskoch94:
Had to add the input/files with add --force, to override the .gitignore. Hope it works now!


johanneskoch94:
ok, one thing I just noticed, my "files" files don't end with a newline... should I fix that quickly?

PR #35, opened 2020-01-16
add testOneRegi option to refactored start scripts

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
- testOneRegi runs are useful for fast development iterations ("do these
  changes break the model, or does it still run"), so setting them up
  should be fast as well

- supplying the parameter `--testOneRegi` to start.R will ignore any
  scenario_config file provided and modify the default configuration to
  start a testOneRegi run in the directory output/testOneRegi with no
  output processing
- the old oneRegi.R script and oneRegi.cfg configuration are deleted

- existing output/testOneRegi directories are overwritten


0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
@dklein-pik Sorry about the white space changes.  That's not me, but RStudio.  You can hide them from the options (gear icon) in the "Files changed" view.  Or append `?w=1` to the URL.

PR #40, opened 2020-01-20
change name of log file from $JOBNAME.out into log.txt to reactivate …

LaviniaBaumstark:
…rs-command on cluster


dklein-pik:
Close because has been addressed in my previous pull request.

PR #65, opened 2020-02-11
Add path_gdx_bau to scenario_config.csv to avoid warning message in s…

dklein-pik:
…tart.R: In is.na(isettings$path_gdx_bau) : is.na() applied to non-(list or vector) of type 'NULL'


dklein-pik:
> Question: `path_gdx_ref` is the "reference" gdx that is used for calculating Consumption Losses and the like, while `path_gdx_bau` is the gdx used for fixing all (relevant) variables before `cm_startyear`?

It's the other way round, I think.

> So if `cm_startyear` is 2015 for both runs, shouldn't there also be a gdx for those periods? My understanding (from reading `start.R`, I think) is that `./config/input_bau.gdx` is used as a default gdx for that, but that's not at all clear from this file.

I don't get your point, sorry. `cm_startyear` should be 2005 for the bau run. The reference run is fixed on the bau run until say 2010, so `cm_startyear` for the reference run is 2015. The policy run is fixed on the refernce run until 2015, so `cm_startyear` for the policy run is 2020. Why would you need another gdx?

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
1) There is a gdx to which variables are fixed before `cm_startyear`.
2) There is a different gdx based on which Consumption Losses etc. are calculated.
3) The two can be different files, therefore there are two paths for them.

I don't know which gdx is which (`path_gdx_bau` v `path_gdx_ref`).  That part of the tutorials seems to be "in the making".  But it should be documented somewhere.

In any case, if both example runs (`RCP37` and `RCP26`) are fixed until `cm_startyear`, they will need both gdxes (for fixing variables and calculating stuff relative to a baseline) to be meaningful examples, right?

So, why is the column `path_gdx_bau` empty in `./config/scenario_config.csv`?  Wouldn't it be a better example if there was something?

PR #72, opened 2020-02-18
Documentation changes

johanneskoch94:
Let me know what you think. The second commit is definitely optional I think, but I though I'ld try to rewrite it...
Also, am I considered an author of REMIND yet? because if yes, my name doesn't appear ;)


johanneskoch94:
Open the file in "edit" mode to enable soft wrap in the upper right corner.

PR #81, opened 2020-02-27
Convergence in time for conversion FE to ES. Affects the reporting only.

MariannaR:
Convergence in time across regions for the conversion between final energy and energy services for LDVs, passenger non LDVs, freight. The parameter is used only in the reporting. Convergence occurs in 2200.

### DIFF
--- a/modules/35_transport/complex/datainput.gms
+++ b/modules/35_transport/complex/datainput.gms
@@ -13,6 +13,8 @@ Parameter
         entrp   1.5
           fetf  0.8
   /
+
+  p35_valconv                "temporary parameter used to set convergence between regions"
 ;
 pm_cesdata_sigma(ttot,in)$p35_cesdata_sigma(in) = p35_cesdata_sigma(in);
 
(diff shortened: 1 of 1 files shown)


modules/35_transport/complex/datainput.gms
```diff
@@ -29,6 +31,25 @@ p35_pass_nonLDV_ES_efficiency(ttot,regi)$(ttot.val ge 2005)  = f35_transp_eff(re
 p35_passLDV_ES_efficiency(ttot,regi)$(ttot.val ge 2005)      = f35_transp_eff(regi,"Eff_Pass_LDV");
 p35_freight_ES_efficiency(ttot,regi)$(ttot.val ge 2005)      = f35_transp_eff(regi,"Eff_Freight");
 
+
+p35_valconv = smin((regi),f35_transp_eff(regi,"Eff_Pass_nonLDV"));
```
Loisel: I'm not quite sure if it is a good Idea to have efficiencies converge to the least efficient public transport system

modules/35_transport/complex/datainput.gms
```diff
@@ -29,6 +31,25 @@ p35_pass_nonLDV_ES_efficiency(ttot,regi)$(ttot.val ge 2005)  = f35_transp_eff(re
...
+
+p35_pass_nonLDV_ES_efficiency(ttot,regi)$(ttot.val ge 2005) = (p35_pass_nonLDV_ES_efficiency(ttot,regi)*(2200-ttot.val)+p35_valconv*(ttot.val-2005))/(2200-2005);
+
+p35_valconv = smin((regi),f35_transp_eff(regi,"Eff_Pass_LDV"));
```
Loisel: For LDVs this might be an even tougher claim.

PR #105, opened 2020-03-12
Add check of R library to guarantee package loads

johanneskoch94:
Hey, so I'm 99% sure that a check for a "^00LOCK.*" folder is enough to figure out if a package is being updated. 
What I've implemented now is a bit heavy-handed, since as soon as any single package is being updated, all packages are loaded from the snapshot library. One could probably have something that only gets the package-being-updated from the snapshot, but oh well...

Do you think this is ok?

### DIFF
--- a/.Rprofile
+++ b/.Rprofile
@@ -2,6 +2,19 @@
 # DO NOT EDIT THIS LINE!
 if(file.exists("~/.Rprofile")) source("~/.Rprofile")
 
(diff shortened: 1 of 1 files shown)


johanneskoch94:
Uh don't approve just yet! I forgot to add the existence check of the snapshot library.

johanneskoch94:
Ok done. 

Before merging, we should test it somehow... like simultaneously update a package and start R on the cluster...

PR #111, opened 2020-03-23
Add output.R script to compute policyCosts

johanneskoch94:
Ok guys, here's the script. 
I don't have write access to the validation folder on the cluster so I couldn't recreate Björn's plots quite yet.. Maybe we should test the script there before merging. 

Also Björn, the relative GDP loss doesn't appear here yet. Since I'm using remind::reportPolicyCosts to compute the other policy costs, I would add the computation there, if that's ok.

Let me know if there is something that should be changed/added!

PS: There is a funny little bug that I don't quite understand. Calling remind::reportPolicyCosts on the cluster only works, if I load the magclass package before (loading the remind package would also work of course, but it's slightly overkill). Locally, and theoretically, I don't/ shouldn't have to. Do you know what's going on here?

PPS: Here's an example output pdf, created with runs lying around in my output folder (completely nonsensical). Just to get an impression.
[PolicyCost.pdf](https://github.com/remindmodel/remind/files/4371257/PolicyCost.pdf)

### DIFF
--- a/scripts/output/comparison/policyCosts.R
+++ b/scripts/output/comparison/policyCosts.R
@@ -1,0 +1,197 @@
+# |  (C) 2006-2019 Potsdam Institute for Climate Impact Research (PIK)
+# |  authors, and contributors see CITATION.cff file. This file is part
+# |  of REMIND and licensed under AGPL-3.0-or-later. Under Section 7 of
+# |  AGPL-3.0, you are granted additional permissions described in the
+# |  REMIND License Exception, version 1.0 (see LICENSE file).
+# |  Contact: remind@pik-potsdam.de
+
+# policyCosts.R
+#
+# This script produces a pdf in which the policy costs of policy-runs with 
+# respect to specified reference runs are displayed.
+# It can be called via the output.R script. If that is the case, the order 
+# in which the runs are selected is important:
+#   1: policy run 1
+#   2: reference run 1
+#   3: policy run 2
+#   4: reference run 2
+# and so on... 
+# 
+
+suppressPackageStartupMessages(library(tidyverse))
+
+# TODO: Loading magclass shouldn't actually be necessary... but running on the 
+# cluster without doing so throws an error...
+suppressPackageStartupMessages(library(magclass)) 
+
+# Function defintions
+get_run_names <- function(filepaths) {
+  # The run names are found between a "/" and the time-stamp of the run.
+  run_names <- str_match(filepaths, pattern = "^.*\\/(.*)(_....-..-..*)$")[,2]
+  
+  # If for some reason there isn't a time-stamp, return the name of the folder
+  if (any(is.na(run_names))) {
+    run_names <- str_match(filepaths, pattern = "^.*\\/(.*)\\/.*$")[,2]
+  }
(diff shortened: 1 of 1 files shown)


bs538:
Thanks for putting this together! A few quick thoughts:
- Using reportPolicyCosts.R as a backend is definitely a good idea. I only realised afterwards that my script had duplicated some existing infrastructure.
- Adding relative GDP loss to reportPolicyCosts.R is also a good idea. X% of GDP in 2050 is easier to grasp than YYY billion $.

On the code itself:
Have you tried your script with different reference runs? I am sceptical whether this will work:
```r
tmp_policy_costs <- lapply(pol_gdxs, remind::reportPolicyCosts, ref_gdxs) 
```
lapply() only iterates over the first argument, so if "ref_gdx" is a character vector with multiple paths I guess it will break. Suggestion: create the vectors with the reference paths beforehand by replicating if necessary, and then use mapply().
Also: might be worth adding an assertion that ensures that ref_gdx is either length 1, or the same length as pol_gdx

About the R-libraries: I'm not surprised that magclass needs to be loaded somewhere along the chain of calls. If you build remind (R) locally, I think it might be loaded already - hence the difference between cluster and local.

johanneskoch94:
Thank you for the great feedback!!
I'm not in the "piam" group yet. I've sent a mail to IT requesting access. As soon as I get it, I'll test the script on the validation runs. 

You're right with the lapply() problem! I've tried it with different ref runs, and it works (doesn't through an error), but after closer examination, it always takes the first ref run in the list! Thanks for catching this mistake so early!

I don't understand the magclass loading issue though... What exactly do you mean by  "build" remind locally?

bs538:
Ah sorry, you haven't actually changed something in the remind R-library so far, so forget what I wrote...

johanneskoch94:
OK :) mapply() does the trick, thx again Björn. I've also added a check on the "outputdirs" to make sure that there are always pol-ref pairs. 
As it is now, there is no way to say "compute all these policy costs with respect to this single reference run".  This could be added of course.

Still waiting on the permissions to "piam".

I see! You were referring to the package build (sry I didn't get it). Nah I didn't update the remind package yet. But even if I had (which I will do now) do you expect the installed package to affect the loaded namespace? Because I do not load the remind package. I just call the single function reportPolicyCosts...

bs538:
Looks good, thanks!
Regarding the multiple policy runs, single reference run: the easiest is probably to check if length(pol_gdxs) > 1 & length(ref_gdxs) == 1, and if yes, simply replicate ref_gdx to the same length. I think I had that in my script. Sounds like a useful functionality to me to make it user-friendly.

Not sure about the package question, probably a question for RSE.

johanneskoch94:
great :) I updated the remind package to compute reference gdp loss so that should work as well now.

Thanks for approving. Let's wait for the validation tests before merging ok?

bs538:
yes, feel free to merge on your own whenever it's done. thanks a lot for putting this together!

johanneskoch94:
Final additions:
- time_stamp the file
- some description
- deletion of pdf helper-files and folder
- adjustment of ggplot parameters

Going to merge now.

PR #125, opened 2020-03-26
Harmonization of UE passenger share of HDV transport

Loisel:
This is required due to high ES demand reported, see this validation issue:

https://gitlab.pik-potsdam.de/REMIND/validation/issues/55

Old behavior:
Screenshot from 2020-03-26 11-12-23

New behavior:
Screenshot from 2020-03-26 20-06-26


Loisel:
> you re not allowed to use region names in the code like "EUR". Please find a more general rule.

Good point. I replaced the value by the average share for developed countries.

PR #156, opened 2020-04-24
Add a switch for green hydrogen

Loisel:
The switch `c_greenh2` defines the share of green hydrogen, i.e., hydrogen made from electricity within overall SE level hydrogen production. To that end

- an equation and a variable is added to `core`,
- the switch and the corresponding documentation is added to `config/default.cfg`.
- the outcomes are tested in a nash run and seem consistent with the reporting.


Loisel:
Addressed the issues, please check again.

PR #159, opened 2020-04-29
Pullreq1

MariannaR:
The pull request contains the latest changes to EDGE-Transport and the realization `edge_esm `of the transport module. EDGE-T alternative decarbonization scenarios are available. The market of LDVs updates endogenously. Bounds on synthetic fuel use and biofuels are included for transportation.


johanneskoch94:
Perhaps squashing commits would be appropriate here? (really just a suggestion!)

Loisel:
I agree that the commit history is messy. Actually, this PR should rather consist of 4 PRs which could be squashed without loosing too much history. Alternatively, one could squash manually some of the commits related to specific features within the PR, to allow for cherry picking.

Loisel:
Please be aware also that *blaming*, which is frequently used in this code base to find out who did what will be made difficult by squashing.

johanneskoch94:
> Please be aware also that _blaming_, which is frequently used in this code base to find out who did what will be made difficult by squashing.

a reason more to do it ;)

PR #177, opened 2020-05-27
change NDC2018: introduce maximum carbon price in region SSA (if exis…

christophbertram:
…tent): 7.5$ in 2020, 30$ in 2025 and 45$ in 2045 (leading to roughly 300Mt higher emissions in 2030


bs538:
@LaviniaBaumstark see also discussion here.

PR #179, opened 2020-05-28
put a minimum on EC shares and modified a constraint on parameter equ…

piklev:
…ation


silviamade:
Not sure what I am supposed to write in my review :)

PR #188, opened 2020-06-12
Revert the formulation of biofuel limit to apply only to the transport sector

Loisel:
The other formulation which applied to all liquids is more consistent with the reporting (on blend serves them all), but to only constrain the biofuel supplies to the transport sector leaves more flexibility to REMIND.

Sorry for the back-and-forth.


Loisel:
By default this switch should not do anything: the upper bound on the share is 1 by default, leaving all freedom to REMIND. The problem that occurred last time was triggered by the wrong parameter name used in the bounding equation (the one for synthetic liquids if I remember correctly, which is quite low).

Loisel:
@LaviniaBaumstark forgot to add the actual commits :)

PR #201, opened 2020-07-01
:sparkles:  Add special request features

johanneskoch94:
Hey Björn, this should enable the addition of transfers. It should be self-explanatory how to use ;) Let me know how it goes and what we still need to change.


bs538:
Thanks a lot! I'll merge this for now and we can add other burden sharing schemes later

PR #211, opened 2020-07-31
fix missing offset_quantity for fehei under fixed_shares introduced in 167d0f20

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:


modules/29_CES_parameters/calibrate/datainput.gms
```diff
@@ -323,6 +323,16 @@ loop ((t_29hist(t),regi_dyn29(regi))$(
   pm_cesdata(t,regi,"fehe_otherInd","offset_quantity")
   = -pm_cesdata(t,regi,"fehe_otherInd","quantity");
 );
+$else
```
Loisel, line 326: Should it be `$else.subsectors` or does it not matter?

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q: Works without, [might not work with](https://www.gams.com/mccarlGuide/ifthen_iftheni_ifthene_conditi.htm?zoom_highlightsub=ifthen):
> One may add a tag to the IFTEN and ENDIF conditions to force them the match up

PR #220, opened 2020-08-12
Tax conv

fschreyer:
This would introduce a convergence check for tax revenues in the Nash algorithm. In addition to the prior criteria, Nash only stops now if abs(vm_taxrev) < 1e-3 for all regions and time steps. This 1.) helps detecting problems when you implement taxes as tax convergence was not automatically checked before. 2.) It lets all tax revenues converge to a 1e-3 which is  ~1% of the smallest GDPs we now have with REMIND EU. It ensures that taxes do not create or eat money.  The precision may be up for debate as there is a trade-off between longer run time and better tax convergence. In my test runs (see below) iterations increased, but only in one case up to 100 (SDP900). 
image

image

### DIFF
--- a/modules/80_optimization/nash/postsolve.gms
+++ b/modules/80_optimization/nash/postsolve.gms
@@ -186,7 +186,7 @@ if(sm_fadeoutPriceAnticip gt 1E-4, s80_bool = 0);
 ***additional criterion: did taxes converge?
 loop(regi,
   loop(t,
-    if( (abs(vm_taxrev.l(t,regi)) gt 1E-4),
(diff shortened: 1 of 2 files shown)


fschreyer:
Can we also add this case in the rs script? Otherwise, people might get confused. How/where can I change that script?

giannou:
The resolution of the images you uploaded is too low to allow me to take a look at them. But ok I understand what you mean. The `rs` script is found here: `/p/projects/rd3mod/tools/rs`, what would you like to include in it?

PR #221, opened 2020-08-24
change tax convergence criterion for nash runs such that tax revenues need to be below one percent GDP

fschreyer:
Change tax convergence criterion for nash runs. Now tax revenues in all regions and time steps need to be below one percent of GDP. In a suite of SSP2 and SDP standard test runs, this did not significantly increase the number of iterations relative to the version without tax convergence check. 

Would be nice if sombody from RSE could add this criterion to the rs script. So that if p80_messageShow("taxconv") = YES, i.e. taxes did not converge, this also shows up in the rs.

### DIFF
--- a/modules/80_optimization/nash/postsolve.gms
+++ b/modules/80_optimization/nash/postsolve.gms
@@ -186,7 +186,7 @@ if(sm_fadeoutPriceAnticip gt 1E-4, s80_bool = 0);
 ***additional criterion: did taxes converge?
 loop(regi,
   loop(t,
-    if( (abs(vm_taxrev.l(t,regi)) gt 1E-3),
+    if( (abs(vm_taxrev.l(t,regi)) / vm_cesIO.l(t,regi,"inco")) gt 1E-2,
      s80_bool = 0;
      p80_messageShow("taxconv") = YES;
     );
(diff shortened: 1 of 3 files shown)


giannou:
can you send me a link to the place where you did the runs?

fschreyer:
/p/tmp/schreyer/Modeling/REMIND_H12/TaxConv_Check/output

giannou:
can you run codeCheck, fix the errors, and correct your PR?

giannou:
Hi Felix, another question: do you have a run where everything other than the taxes has converged? It seems to me that rs captures non-convergence regardless of the reason it happens, that's why I am not sure how to include tax convergence in it.

fschreyer:
Yes, in this run /p/tmp/schreyer/Modeling/REMIND_H12/TaxConv_Check/output/SDP-PkBudg900_2020-08-11_12.54.12. 
All markets converged, but taxes did not. Rs shows it as green. However, note that this was my old tax convergence criterion two weeks ago, which I now changed. So, by normalizing by GDP it should not happen so often that iterations are increased only because of tax convergence criterion.

fschreyer:
Also, I should probably know this but...how do I run codeCheck again?

giannou:
> Also, I should probably know this but...how do I run codeCheck again?

`gms::codeCheck(strict=TRUE)`

fschreyer:
Ok, done! Why do we have to add variables to the not_used.txt?

giannou:
> Ok, done! Why do we have to add variables to the not_used.txt?

Thanks! Now the tests are successful :) The modularization of REMIND is based on strict interface definition principles, that's why we need to monitor all objects appearing in modules.

giannou (approved):

giannou:
`rs` is now marking as red the runs where the taxes haven't converged. I did a couple more improvements, feedback is welcome!

fschreyer:
Cool thx! It says "Failing markets: GeneralAlgebraModelingSystem" for the non-converged tax run for me. Could we make the error message more specific like "Taxes did not converge, see vm_taxrev level"?

giannou:
yes, I know.. working on it.. the script is under continuous improvement

PR #225, opened 2020-08-28
overwrite the 2025 and 2030 lower bounds for nuclear in NDC2018 with …

robertpietzcker:
…zero for nucscen=5 (= no new nuclear construction after 2020)


giannou (approved):
Everything looks OK to me but I also started a cascade of test runs with these changes just to be sure.

PR #234, opened 2020-09-11
Added a differentiated coal phase-out depending on 2010 GDP per capita. This allows different phase-out stringencies for CHN and other developing regions.

rauner:


christophbertram (approved):
Looks good, "more rich" or next time "richer"??

PR #237, opened 2020-09-15
rescaling of prices in the calibration and additions to the buildings module for the computation of shares

piklev:
When the CES function contains a putty-clay part and/or complementary factors, we cannot infer the ```en``` quantity directly from the prices and quantities at the ppfen_29 level. Therefore, we unfortunately cannot rescale prices before all quantities are computed. 
I put the rescaling of prices in case the share of ```lab``` is too low after the computation of quantities. Quantities are computed anew in case of a rescaling.


MariannaR (approved):

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
> Quantities are computed anew in case of a rescaling.

But then they might be inconsistent with the exogenously prescribed efficiencies in `industry/subsectors`, which is why I moved this to the top in the first place.  🤔

I will do some testing …

piklev:
@0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q could you make the tests? Can I merge?

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
Sorry, I didn't get around to testing.
Commit.  Either it works, or I have to come up with a solution in any case. ;)

piklev:
The pull request now also includes improvements in the buildings detailed modules. The FE-UE technologies now have vintages and the shares are now computed based on a model, which should make it easier to impose restrictions

PR #238, opened 2020-09-16
Tax convergence check update

fschreyer:
This again modifies the tax convergence check for nash iterations. The criterion is now that tax revenues in all regions must be below 0.01% of GDP in time steps before 2100. Hope this meets the desired precision for regional policy cost etc. I tested SSP2 and SDP runs and it does not significantly increase iterations. If ever this criterion bothers, one can switch it off by setting ``cm_TaxConvCheck`` to 0.


LaviniaBaumstark (approved):

LaviniaBaumstark (changes requested):
please run codeCheck for testing if you follow the coding etiquette

LaviniaBaumstark (approved):

Issue #180, opened 2020-05-29
Reporting error: Fortran runtime Error in the Magicc Reporting


Loisel:
When `reporting.R` is executed, the magicc reporting parts fail with
```
[1] "Executing reporting.R"                                                                                                                                                                                   
/p/tmp/aloisdir/remind-trunk/output/SDP-calibrate_2020-05-27_16.57.02/magicc                                                                                                                                  
At line 1299 of file MAGICC_6005_v2_forJessicaGunnarElmar.F90 (unit = 42, file = 'MAGCFG_NMLYEARS.CFG')                                                                                                       
Fortran runtime error: End of file                                                                                                                                                                            
sh: climate_reporting_template.txt: No such file or directory                                                                                                                                                 
sed: can't read ../../../output/SDP-calibrate_2020-05-27_16.57.02/REMIND_climate_SDP-calibrate.mif: No such file or directory                                                                                 
cat: ../../../output/SDP-calibrate_2020-05-27_16.57.02/REMIND_climate_SDP-calibrate.mif: No such file or directory 
```

Consequentially the EDGE Transport reporting is not executed (it comes after the magicc reporting in `reporting.R`) but a MIF file is produced so without checking the logs or looking for EDGE-T or MAGICC reporting, no error is evident.

giannou:
@LaviniaBaumstark had asked me to remove the magicc files from the model when we went open-source. Not sure what the status is right now

johanneskoch94:
I'm trying to run remind with magicc, and it's not working.  I get the same error as @Loisel showed above. @0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q  is helping me with a workaround. If the magicc files can't be open-sourced, is there a standard set-up for running magicc?

PR #247, opened 2020-09-28
add symlinks to MAGICC files

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
I'm still of the opinion that we can just add the MAGICC files, since you also get them on their website, but here goes:

In order to have a REMIND with on less source of constant failure (#180), we can add symlinks to the `./core/magicc/` directory.  The symlinks only contain the information where the files are to be found ([e.g.](https://github.com/0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q/remind/blob/feature/symlink_magicc/core/magicc/EDGAR_CH4_EMIS_LANDUSE.IN)), but no actual data.  When the run is assembled by `./scripts/start/prepare_and_run.R`, the `./core/magicc/` directory [is copied](https://github.com/remindmodel/remind/blob/3e4fc335245fc37c4ed0572eb4aee960a8c29182/scripts/start/prepare_and_run.R#L403) using `file.copy()` [which copies the targets of the links, not the links](https://stat.ethz.ch/R-manual/R-devel/library/base/html/files.html): 
> On a POSIX filesystem the targets of symbolic links will be copied rather than the links themselves

so we end up with proper MAGICC in the output directory, without any actuall MAGICC in the repository.

If this is agreeable, pick a proper place for the MAGICC files, not my home directory ;)

### DIFF
--- a/core/magicc/EDGAR_CH4_EMIS_LANDUSE.IN
+++ b/core/magicc/EDGAR_CH4_EMIS_LANDUSE.IN
@@ -1,0 +1,1 @@
+/home/pehl/swap/magicc/EDGAR_CH4_EMIS_LANDUSE.IN
--- a/core/magicc/EDGAR_NOX_EMIS_FOSSIL_IND.IN
+++ b/core/magicc/EDGAR_NOX_EMIS_FOSSIL_IND.IN
@@ -1,0 +1,1 @@
+/home/pehl/swap/magicc/EDGAR_NOX_EMIS_FOSSIL_IND.IN
--- a/core/magicc/EDGAR_NOX_EMIS_LANDUSE.IN
+++ b/core/magicc/EDGAR_NOX_EMIS_LANDUSE.IN
@@ -1,0 +1,1 @@
+/home/pehl/swap/magicc/EDGAR_NOX_EMIS_LANDUSE.IN
--- a/core/magicc/GISS_BCB_OT.IN
+++ b/core/magicc/GISS_BCB_OT.IN
@@ -1,0 +1,1 @@
+/home/pehl/swap/magicc/GISS_BCB_OT.IN
--- a/core/magicc/GISS_BCB_RF.IN
+++ b/core/magicc/GISS_BCB_RF.IN
@@ -1,0 +1,1 @@
+/home/pehl/swap/magicc/GISS_BCB_RF.IN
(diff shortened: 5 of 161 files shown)


giannou (commented):
Looks good to me, thanks Michaja. I would also very very strongly prefer to just bring back the MAGICC files. @LaviniaBaumstark

giannou:
I put the files here: /p/projects/rd3mod/magicc
Shall we change the targets of the symbolic links to point there?

giannou:
The MAGICC developers have a no-further-sharing restriction in their webpage: http://www.magicc.org/about/legalnotice
This means we can not upload the files here after all, right?

giannou:
> I put the files here: /p/projects/rd3mod/magicc
> Shall we change the targets of the symbolic links to point there?

Also, RSE recommends to change the implementation in a way that no PIK-cluster-specific paths are found in the code. How about a config setting called `magicclocation` which contains the path to the local MAGICC files? In a more general note: is it necessary to copy the whole magicc folder into each run folder?

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
> This means we can not upload the files here after all, right?

We would not upload files, only symlinks.

> Also, RSE recommends to change the implementation in a way that no PIK-cluster-specific paths are found in the code. How about a config setting called `magicclocation` which contains the path to the local MAGICC files?

Possible, yet pointless, since we cannot share "our" MAGICC files with anyone, and they can't get "our" MAGICC files in any other way, and dropping any other MAGICC files in there would produce different results, which we didn't test and aren't willing to accept as "REMIND results".

giannou:
> > This means we can not upload the files here after all, right?
> 
> We would not upload files, only symlinks.
I was referring to the previous discussion about uploading the MAGICC files in the REMIND repository, of course I was not referring to the links
> 
> > Also, RSE recommends to change the implementation in a way that no PIK-cluster-specific paths are found in the code. How about a config setting called `magicclocation` which contains the path to the local MAGICC files?
> 
> Possible, yet pointless, since we cannot share "our" MAGICC files with anyone, and they can't get "our" MAGICC files in any other way, and dropping any other MAGICC files in there would produce different results, which we didn't test and aren't willing to accept as "REMIND results".

PR #251, opened 2020-10-01
improve realism of BEV upscaling in USA in 2025: increase deltacapoff…

christophbertram:
…set by  3Mio vehicles

### DIFF
--- a/core/datainput.gms
+++ b/core/datainput.gms
@@ -733,6 +733,9 @@ $offdelim
 /
 ;
 p_adj_deltacapoffset("2015",regi,"tnrs")= 1;
+***special case: BEV in USA in 2025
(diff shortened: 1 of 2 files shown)


Loisel (approved):

LaviniaBaumstark (changes requested):
there should be no regions called explicitly in the GAMS code. If you want this change, you have to adjust the input data generation

christophbertram:
I thought that the construction "("2025",regi,"apCarElT")$(sameas(regi,"USA"))" is allowed, as it would just mean that no region value is changed in case that "USA" is not part of the regi set. We have this in some modules as well...

LaviniaBaumstark (changes requested):
well, we only allow for region names in the NPI/NDC modules, all the rest of the code should be clean

PR #253, opened 2020-10-03
improve realism of BEV upscaling in USA in 2025: increase deltacapoff…

LaviniaBaumstark:
… set by 3Mio vehicles, new input data revision


christophbertram (approved):
in main.gms, cm_OILRETIRE is set to "on", to be in line with default.cfg setting. A future commit should also correct the "!! def = off" to "on"...

PR #255, opened 2020-10-05
add flexibility tax and switch H2 subsidy by default

fschreyer:
This adds a flexibility tax to REMIND main. Currently, the flexibility tax provides a subsidy to electrolysis with increasing VRE share. It is based on the idea that with high wind and solar shares, electrolysis can operate flexibly (at lower capacity factors) and face lower-than-average electricity prices. This first implementation reduces the capacity factor of elh2 from 0.9 to 0.5 and provides a subsidy that roughly halves the electricity price that electrolysis sees. This follows results of more detailed power system models. The flexibility subsidy is switched on by default. 

In addition, the switch ``cm_FlexTaxFeedback`` can be switched on (it is off by default). This is work in progress and would enable the model to choose the capacity factor of electrolysis endogenously and includes a feedback of flexible operation of electrolysis on inflexible electricity demands. The lower-than-average electricity prices of electrolysis are balanced (load-weighted) by higher-than-average electricity prices for industry and buildings (tdels).


fschreyer:
@LaviniaBaumstark Kannst du Falko mal dem github hinzufügen? Dann hole ich ihn hier als Reviewer mit rein.

fschreyer:
Here, how the flexibility tax (which so far only provides a subsidy to electrolysis) pushes H2 in the current REMIND:

Right is a standard Peak Budget 900 run, center is flexibility subsidy for H2 switched on, left is flexibility subsidy + CCU switched on. 

image

LaviniaBaumstark (changes requested):
please, make sure that codeChecks runs without any error

FalkoUeckerdt:
Thanks so much Felix!

1) Biomass and CCS are not limited here, are they?
2) Is CCU on by default? I suggest, it should be given their importance in the debate. 
3) The FE graph does show synfuels as part of liquids and gases and only shows hydrogen that is directly used. So synfuels and h2 for synfuels are not shown here, right (in the CCU on version)?
4) Why is the transition to h2 use so slow, or only significant after 2050? Especially synfuels are only used after 2050.
5) The regional distribution seems very adequate. Those regions with high VRE potential and demand side challenges go to hydrogen. REMIND might actually see trade potential here.
6) The EUR H2 results are still far less ambitious than the EU H2 strategy
7) Could you also show H2 and synfuel supply costs and FE liquid and gas prices for the different scenarios?
8) No gaseous synfuels anymore. Makes sense in general, and might change with more constraints on CCS and biomass.
9) How did you parameterize DAC now? Is fossil CCU possible and used in the mid-term (despite there being now PtL in the mid-term yet)?

FalkoUeckerdt:
I made these bottom-up cost graphs for the e-fuel paper. I also calculate required CO2 prices. It shows that e-fuels can become a backstop with further learning.

image
image

It would be good to see how your parameterization and REMIND costs relate to this.

FalkoUeckerdt (approved):

FalkoUeckerdt:
Thanks Felix! The parameterization is a good approximation. Still, let's have a closer look sometime soon. The 0.5 at 100% VRE seems a bit pessimistic. I see 0.5 at ~50% VRE based on price data for South Australia already.

LaviniaBaumstark (approved):

PR #268, opened 2020-10-29
Bunkers calculation based on EDGE-T data for complex transport realization

MariannaR:
This PR contains:

- the updated calculation of bunker shares on total `fedie` FE demand based on `EDGE-T` values from `mrremind`.

- a cache update, to allow for the new bunker shares data to be processed by REMIND.


christophbertram (approved):
All good

PR #270, opened 2020-11-04
Most importantly: add automatic copying of magicc files

dklein-pik:


LaviniaBaumstark (changes requested):
the folder magicc_template should be a new argument in the default.cfg which can be changed by the user

LaviniaBaumstark (approved):

PR #283, opened 2020-11-27
add a temporary note in README informing people about pending input data issues

giannou:


LaviniaBaumstark (changes requested):
please, try not to write that you can not run REMIND without input data but more that you need input data

PR #285, opened 2020-12-01
synfuel update

fschreyer:
This fixes synfuels to zero in baseline runs and moves the minmial lower bounds of synfuel technologies to after 2030 (avoids infeasibilities in certain cases). Also, a switch to force in synthetic gas was added in analogy to the one for synthetic liquids. These switches now force in synfuels starting in 2035 (more realistic than from 2025 onwards after discussion with industry guy from Sunfire).


fschreyer:
Can somebody please approve this?

Loisel:
Ok maybe I have some misunderstanding but could you at least comment on my comment? I'm happy to approve this if you are confident it works out.

Loisel (approved):

PR #290, opened 2020-12-18
update scripts to future proof against R 4.0.0+ updates

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
- https://stat.ethz.ch/pipermail/r-announce/2020/000653.html
    > R now uses a stringsAsFactors = FALSE default, and hence by
    > default no longer converts strings to factors in calls to
    > data.frame() and read.table().

  - this breaks all instances of levels() being used on data.frames
    read using read.csv()

  - switching to unique() solves this

  - unique() used on factors (R < 4.0.0, currently on the cluster) still
    returns the same values as levels(), but as a factor, which does not
    cause problems (but is a little ugly in interactive use)

### DIFF
--- a/scripts/start/prepare_and_run.R
+++ b/scripts/start/prepare_and_run.R
@@ -8,7 +8,7 @@ library(lucode, quietly = TRUE,warn.conflicts =FALSE)
 library(dplyr, quietly = TRUE,warn.conflicts =FALSE)
 require(gdx)
 
-################################################################################################## 
+##################################################################################################
 #                             function: getReportData                                            #
 ##################################################################################################
 
@@ -27,7 +27,7 @@ getReportData <- function(path_to_report,inputpath_mag="magpie",inputpath_acc="c
     out["JPN",is.na(out["JPN",,]),] <- 0
     dimnames(out)[[3]] <- NULL #Delete variable name to prevent it from being written into output file
     write.magpie(out[notGLO,,],paste0("./modules/30_biomass/",inputpath_mag,"/input/p30_pebiolc_pricemag_coupling.csv"),file_type="csvr")
-  }  
+  }
   .bioenergy_costs <- function(mag){
     notGLO <- getRegions(mag)[!(getRegions(mag)=="GLO")]
     if ("Production Cost|Agriculture|Biomass|Energy Crops (million US$2005/yr)" %in% getNames(mag)) {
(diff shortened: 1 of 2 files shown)


0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
N.B.: I wasn't bored.  This of course broke REMIND on my machine after updating to R 4.0.2.

giannou (approved):

giannou:
the checks have failed (but for a random reason, it seems) so for me the PR is ok

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
> the checks have failed (but for a random reason, it seems) so for me the PR is ok

Not so random.

```
* installing *source* package â€˜gertâ€™ ... 
** package â€˜gertâ€™ successfully unpacked and MD5 sums checked 
** using staged installation 
Using PKG_CFLAGS= 
Using PKG_LIBS=-lgit2 
----------------------------- ANTICONF ------------------------------- Configuration failed to find libgit2 library. Try installing: 
* brew: libgit2 (MacOS) 
* deb: libgit2-dev (Debian, Ubuntu, etc) 
* rpm: libgit2-devel (Fedora, CentOS, RHEL) 
If libgit2 is already installed, check that 'pkg-config' is in your PATH and PKG_CONFIG_PATH contains a libgit2.pc file. 
If pkg-config is unavailable you can set INCLUDE_DIR and LIB_DIR manually via: R CMD INSTALL --configure-vars='INCLUDE_DIR=... LIB_DIR=...' 
-------------------------- [ERROR MESSAGE] --------------------------- <stdin>:1:18: fatal error: git2.h: No such file or directory compilation terminated. 
---------------------------------------------------------------------- 
ERROR: configuration failed for package â€˜gertâ€™
```

You need to update your Travis config.

giannou:
looks like the error is indeed random: I [changed](https://github.com/remindmodel/remind/pull/292) the travis config of REMIND to be exactly the same as the one of MAgPIE (which [works](https://travis-ci.com/github/magpiemodel/magpie/builds/210029807)) but still get the same [error](https://github.com/remindmodel/remind/pull/292/checks?check_run_id=1589585217). it looks as though the machine where the checks of MAgPIE run have `devtools` installed already whereas the machine or REMIND does not. We have to [move away from travis](https://news.ycombinator.com/item?id=25338983) anyway and switch to [Github actions](https://docs.github.com/en/free-pro-team@latest/actions/quickstart) so not sure if this is worth looking further into.

giannou:
Update: by tweaking the config a little bit ([1](https://github.com/remindmodel/remind/pull/292/commits/72fe060cf40b3706ca237e8e9a3926b820c6d643),[2](https://github.com/remindmodel/remind/pull/292/commits/b2be337aa5ce2d987ec8f1e253fb776ee7f9b05c)) the build works.

PR #300, opened 2021-01-15
Updated cache; reporting adapted to regional EDGE-T.

MariannaR:
This PR contains:

- the updated cache number which contains the regional-level input data for EDGE-T
- the fixes to the reporting of transport for regional-level EDGE-T


config/default.cfg
```diff
@@ -22,7 +22,7 @@ cfg$title <- "default"
 cfg$regionmapping <- "config/regionmappingH12.csv"
 
 #### Current input data revision (<mainrevision>.<subrevision>) ####
-cfg$revision <- 5.964
+cfg$revision <- 5.971
```
Loisel, line 25: wow this looks like a big jump

Loisel (approved):

config/default.cfg
MariannaR: there were many errors in mrremind, and the in-between broken caches are not to be used.

PR #303, opened 2021-01-22
Magpie coupling:  Change the letter used to split the list of MAgPIE scenarios into individual scenarios from "," to "|" to avoid accidental splitting by text editors that use the comma as a column separator.

dklein-pik:


jhilaire (approved):
Looks good to me!

PR #305, opened 2021-01-27
Temporarily removed learning.

MariannaR:
This PR contains:

- cleaning of deprecated function from EDGE-T iterative.
- temporarily removed learning due to a bug - it is on the TODO list to fix it.


Loisel (approved):
Better no learning than broken learning.

PR #311, opened 2021-02-12
REMIND-EU merge request

Renato-Rodrigues:
This pull request includes all code changes necessary to run REMIND with an alternative 21 regions formulation with emphasis on European countries and regions, also known as REMIND-EU.


LaviniaBaumstark (changes requested):
why is in defautl.cfg cm_TaxConvCheck turned off?
why is the list of outputs restricted to c("reporting","reportCEScalib")?
And cm_OILRETIRE default changed for "off" to "on". Is this intended?
In main.gms the following sets are different from the default:
*** EU bioenergy switches
cm_BioSupply_Adjust_EU = 3; !! def 1
cm_BioImportTax_EU = 1; !! def 0.2

Renato-Rodrigues:
About three weeks ago cm_TaxConvCheck had problems with smaller regions (REMIND-EU case). Felix recommended me to turn it off as he was also doing the same for ARIADNE scenarios. As this check was added by him just recently, I would keep it off until we are sure that it is consistent for REMIND runs.

The list of automatically created outputs can be extended again as soon as the alternative reportings are tested. I kept the contents to a minimum as I was not able to test all combinations by myself.    

cm_OILRETIRE default value should be on. This was discussed in a REMIND meeting about two/three months ago. This was already changed in many config scenarios, but it should be the default value for any REMIND run.

I believe the cm_BioSupply_Adjust_EU and cm_BioImportTax_EU values reflect the current ARIADNE assumptions. @fschreyer would know better about the correct values for that. I would keep whatever is used in ARIADNE as the default value for REMINd runs.

LaviniaBaumstark:
I agree regarding  cm_TaxConvCheck, cm_OILRETIRE and the list of reportings. And also regarding the default settings for cm_BioSupply_Adjust_EU  and cm_BioImportTax_EU. If we change settings we should do it not by accident but decide actively. That's why I ask you.

Renato-Rodrigues:
I completely agree with you! And it is also nice to have a second pair of eyes to check for any of my mistakes :) 
By the way, Felix just confirmed me that these are indeed the current default values for the cm_BioSupply_Adjust_EU and cm_BioImportTax_EU.

LaviniaBaumstark:
fine,thanks, please adjust the default values

LaviniaBaumstark (approved):

PR #319, opened 2021-03-01
Update 2_RunningREMIND.md

cchrisgong:
some addition to the tutorial on restarting REMIND


Renato-Rodrigues:
@cchrisgong 
Dear Chris, thk you for adding the detailed description to this point. 
Could I ask you to just make a small change before accepting it?
I would remove the last sentence as it is very specific and can add confusion to the reader ("Be careful convergence indicators such as p80_surplusMax2100 might not be as close to convergence as the previous unconverged run until 28 or 29 iterations.").

Renato-Rodrigues (approved):

PR #320, opened 2021-03-02
merge ariadne branch mainly with updates to H2 td, refactoring of FE share bounds, fix to EDGE-T to run after merge

fschreyer:
This merges the ariadne branch. Please take a closer look as this originates from my old REMIND EU branch and I pulled the recent main develop branch into it.  

Main changes are:
* introduce capacities for transmission and distribution of H2 in buildings and industry to prevent jumping of H2 between these sectors
* refactoring of Renato's implementation of FE share bounds that is switched on optionally in some scenarios and inclusion of ariadne switches
* Marianna fixed EDGE-T to run with new merged REMIND-EU
* updates ariadne scenario config and REMIND-EU EDGE-T input gdx

PS: Ok, I completed my checks now. I tested the H12 SSP2 config and ARIADNE EU config.


LaviniaBaumstark (commented):
what will happen to the EU-specific switches? In general, they are not only REMIND-EU specific, but they mainly need some specific countries as single regions, right? 
Did you also adjust the R-package remind2 if necessary (e.g. renamed variables)?

fschreyer:
Yes, the EU-specific switches are only active if ``EUR_regi`` exists and only for those regions in that region group. At one point we should make this fully region flexible, see this [issue](https://github.com/remindmodel/development_issues/issues/27). 

The variables I changed names of do not appear in the reporting. And the remind2 reporting runs for both setups now (finally!), see [here](https://github.com/pik-piam/remind2/pull/16).

LaviniaBaumstark (approved):

PR #322, opened 2021-03-03
Update call to edgeTrpLib function readREMINDdemand.

MariannaR:
This PR contains the necessary fixes to remove the dependency on `mrremind` of the `edgeTrpLib`.

### DIFF
--- a/scripts/iterative/EDGE_transport.R
+++ b/scripts/iterative/EDGE_transport.R
@@ -22,6 +22,8 @@ require(devtools)
 library(rmndt)
 library(mrremind)
 
+print("Start of the EDGE-T iterative model run")
+
 ## use cached input data for speed purpose
 setConfig(forcecache=T)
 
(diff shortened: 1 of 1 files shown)


scripts/iterative/EDGE_transport.R
```diff
@@ -111,7 +111,13 @@ pref_data$VS1_final_pref = rbind(prefdata_nonmotV, pref_data$VS1_final_pref)
 ## optional average of prices
 average_prices = TRUE
 
-ES_demand_all = readREMINDdemand(gdx, REMIND2ISO_MAPPING, EDGE2teESmap, REMINDyears, scenario)
+## calculate the ES demand (in million km)
+GDP = calcOutput("GDPppp", aggregate = T)[,, scenario, pmatch=TRUE]
+GDP <- as.data.table(GDP)
+GDP[, year := as.numeric(gsub("y", "", Year))][, Year := NULL]
+setnames(GDP, old = "ISO3", new = "region")
+ES_demand_all = readREMINDdemand(gdx = gdx, REMINDmapping = REMIND2ISO_MAPPING, EDGE2teESmap = EDGE2teESmap, years = REMINDyears, scenario = scenario, GDP = GDP)
```
Loisel, line 119: essentially there is now not much happening in `readREMINDdemand`s call to `getGDP`. Do you think it makes sense to move all this treatment of GDP inputs here and only supply the *product* to readREMINDdemand? Or have all treatment in `edgeTrpLib`, but then the "raw" output of `calcOutput` should be supplied...

Loisel (approved):

PR #325, opened 2021-03-04
Removing adjustment costs modifiers from default REMIND21 calibration runs / BugFix removing log.tx write commands if the file does not exist  

Renato-Rodrigues:
- Removing adjustment costs modifiers from default REMIND21 calibration runs
- BugFix removing log.tx write commands if the file does not exist


giannou (commented):
Thanks for the bugfix! Is adding yet another GDX and more CSV files absolutely necessary? This is exactly what causes the repository to become so big.

Renato-Rodrigues:
The csv file is not new, I just changed two lines on it. This change is definitely necessary.

On the gdx file we need to make a decision. If I don't commit it, it means that the initial point for the model is not in par with the calibration. This certainly causes the first iteration of the model to take longer on every run that anyone makes, but can also sometimes causes the model to not converge depending on how strong is the calibration change.
This needs to be further discussed, but I believe the way to handle this is not to avoid updating the initial point, but instead to move the gdx files outside of the repository.

If you want we can remove for now the gdx file from the commit until we can make a decision about that.

giannou:
This is not my call and changing our GDX and other binary files policy (sorry about the CSV) cannot happen in such short notice. I suggest you leave the GDX out for now and we start the discussion on where to put these files immediately.

PR #327, opened 2021-03-05
Add synfuels as an additional SE type, harmonize fuel price calculations in postsolve

fschreyer:
This introduces
* seliqsyn/segasyn as a seperate SE type
* internal fuel price calculations on PE, SE and FE level in postsolve files, this harmonizes different postsolve statements of this kind and will also make the price reporting more straightforward (the FE prices also include taxes now, see test [here](https://gitlab.pik-potsdam.de/REMIND/validation/-/issues/109))
* fix of potential bug of energy extraction emissions (see [issue](https://github.com/remindmodel/development_issues/issues/30))

The first two should only be a refactoring. 

Introduction of seliqsyn made changes to 45_carbonprice necessary. Adding @christophbertram to be aware. 

Note on the SE to FE implementation: The seliqfos/seliqbio/seliqsyn split we have now should at some point be complemented by introducing one overarching td technology which maps all three SE carriers to FE (now we have three seperate ones). I did not do this because this, I guess, would be the first instance of a technology, with more than one mapping and introduces some complications. So for now, the td cost of biofuels and synfuels may be a bit too high because these fuels could actually use the "old" oil pipelines and do not necessarily need new ones. But, I think we can accept that for now for more clarity on the synfuels in REMIND and its reporting and the advantage of having synfuel prices. 

PS: Currently still testing reporting and discussing the extraction issue. codeCheck will be adapted in the end because it needs adjustments in 10+ not_used.txt files in 45_carbonprice.


core/input/generisdata_tech.prn
```diff
@@ -96,7 +96,14 @@ inco0               1800         360         360          60          60
 mix0                1.00        1.00        1.00        1.00        1.00        1.00        1.00       
 eta                 0.95        0.85        0.85        1.00        1.00        1.00        1.00   
 omf                 0.03        0.02        0.02        0.03        0.03        0.02        0.06    
-lifetime              45          45          45          30          30          45          45       
+lifetime              45          45          45          30          30          45          45
```
Loisel, line 99: Again some values where it would be interesting to know where they come from... Just a general comment.

Loisel (approved):
Thanks for this important step! For all applications of synthetic liquids this is very useful. I scanned over the changes, but it's hard to tell if everything is fine given the volume. Please test thoroughly.

fschreyer:
@Renato-Rodrigues merge is done.

PR #332, opened 2021-03-12
update on tutorial restart

cchrisgong:
Thanks, Renato for the review. I have removed the sentence in question

### DIFF
--- a/modules/32_power/IntC/bounds.gms
+++ b/modules/32_power/IntC/bounds.gms
@@ -8,6 +8,9 @@
 ***                  module specific bounds
 ***------------------------------------------------------------
 
+*** Fix capacity factors to the standard value from data
+vm_capFac.fx(t,regi,te) = pm_cf(t,regi,te);
+
 *** FS: for historically limited biomass production scenario (cm_bioprod_histlim >= 0)
 *** to avoid infeasibilities with vintage biomass capacities
 *** allow bio techs to reduce capacity factor
--- a/modules/32_power/IntC/bounds.gms
+++ b/modules/32_power/IntC/bounds.gms
@@ -8,9 +8,6 @@
 ***                  module specific bounds
 ***------------------------------------------------------------
 
-*** Fix capacity factors to the standard value from data
-vm_capFac.fx(t,regi,te) = pm_cf(t,regi,te);
-
 *** FS: for historically limited biomass production scenario (cm_bioprod_histlim >= 0)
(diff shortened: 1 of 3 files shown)


modules/32_power/IntC/bounds.gms
```diff
@@ -8,9 +8,6 @@
 ***                  module specific bounds
 ***------------------------------------------------------------
 
-*** Fix capacity factors to the standard value from data
```
0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q, line 11: This is the point when you learn why it is useful to create separate branches for each pull request.  Surely your `pm_cf` tinkering is not intended to be included in the pull request for the `2_RunningREMIND.md` ;)

cchrisgong: Thanks, Michaja - good to know! Is there some way to correct my pull request title to reflect the second change? I will do 2 PRs next time

cchrisgong: ah yeah, never mind I found it at the top :D

Renato-Rodrigues: moving the vm_capFac.fx to the core potentially breaks the RLDC formulation. @cchrisgong @robertpietzcker

robertpietzcker: Getting RLDC to then work again (in case someone does that) would simply requiring setting 
vm_capFac.up = 1e6 or so for the technologies that are in the RLDC formulation, right?

so my gut feeling would be having the general fix for all "te" on core level and then relaxing it for specific "te" in submodules that need this additional flexibility for these specific "te" makes sense from a module-organization point of view?

Renato-Rodrigues: My point is still valid. Any change that occurs in a module with multiple realizations should take into consideration if they would break the other module. So I would say this change can only be accepted if the change to the RLDC module is also considered.

I also don't think that the upper bound change is enough. One would need to consider both vm_capFac.up and vm_capFac.lo to re-free a variable, and I would be very careful on using that for variables with so many dimensions to avoid adding unnecessary constraints to the model on unused dimension combinations.
 
Last year I added changes to the RLDC formulation work. I wouldn't expect that it wouldn't work today, but we never gave attention to analyse its results further so it never got back into the focus.

Renato-Rodrigues (approved):

PR #333, opened 2021-03-16
intitialize price parameters before cm_startyear

fschreyer:
intitialize price parameters before cm_startyear, fix for having prices in first years


Renato-Rodrigues:
Wouldn't this break the model for any input.gdx file that does not have pm_FEPrice defined? 
As far as I know whenever you execute a Execute_Loadpoint and the parameter is not found gams would issue an error.

fschreyer:
Well, it would if you use an incompatible reference run to your run, which normally we try to avoid anyways. It is reading only from the ``input_ref.gdx`` in case of a run that did not start 2005. Not from the not the ``input.gdx``.  So, if you run baseline first, the subsequent runs should be fine.

Renato-Rodrigues:
But you are breaking backwards compatibility for anybody that uses any older reference scenario run, even if you want to just run a single new scenario. 
I would agree with this change for example once all our input.gdx files also include pm_FEPrice values, which is not the case yet. However for now I would add it under a switch (disabled by default) to avoid breaking current papers analysis that just need a new scenario but want to take advantage of the latest fixes in the model.

Renato-Rodrigues (approved):

PR #336, opened 2021-03-16
Introducing secondary energy trade to REMIND

Renato-Rodrigues:
* secondary energy trade is currently based in exogenous defined trade capacity interconnection between countries (see pm_seTradeCapacity).
* currently only Germany hydrogen trade scenarios are parameterized (high and low hydrogen importing scenarios). To enable it: (1) choose the se_trade realization for the 24_trade module, and (2) choose the trade scenario in the cm_seTradeScenario switch.


modules/01_macro/singleSectorGr/equations.gms
```diff
@@ -41,6 +41,8 @@ qm_budget(ttot,regi)$( ttot.val ge cm_startyear ) ..
         )$( ttot.val ge max(2010, cm_startyear) )
       )
     )
+  + sum(tradeSe, pm_MPortsPrice(ttot,regi,tradeSe) * vm_Mport(ttot,regi,tradeSe))
```
fschreyer, line 44: The corresponding lines for PE trade have the trade (transport) costs here, not the price of the good. Is this necessary because we do not implement SE trade via the nash markets yet?

modules/24_trade/se_trade/bounds.gms
```diff
@@ -0,0 +1,129 @@
...
+);
+
+*** FS: constrain biomass imports in EU subregions from cm_startyear or 2020 onwards to a quarter of 2015 PE bioenergy demand
+if ( cm_biotrade_phaseout eq 1,
+	vm_Mport.up(t,regi,"pebiolc")$(t.val ge cm_startyear AND t.val gt 2015 AND regi_group("EUR_regi",regi)) = 
+      pm_demPeBio("2015",regi)$(regi_group("EUR_regi",regi))/4;
+);
+
+*** set maximum import and export secondary energy trade based on trading capacities 
+vm_Mport.up(t,regi,entySe) = sum(regi2,p24_seTradeCapacity(t,regi2,regi,entySe));
```
fschreyer, line 126: What is the purpose of introducing a trade capacity?

fschreyer (approved):
Ok, thanks for this. 

Just a general remark: Is this compatible with the (final) trade implementation via the nash markets (as PE trade now) that we want? Because I see some code that is not there for PE trade. Would be great if the final implementation was close to the PE one. 

But I totally understand that for now we go a different route.

robertpietzcker:
> 
> 
> Ok, thanks for this.
> 
> Just a general remark: Is this compatible with the (final) trade implementation via the nash markets (as PE trade now) that we want? Because I see some code that is not there for PE trade. Would be great if the final implementation was close to the PE one.
> 
> But I totally understand that for now we go a different route.

Hi Felix, 

Renato implemented it in a fixed way because I told him that we need some quick implementation of moving energy from region a to region b for ARIADNE - he wanted to do it more "correct", but I pressured him to do it this way because we have too many other things that are more urgent.
Implementing it along the nash algorithm will be the work of the PhD/PostDoc that Falko is hiring, so almost surely there will be (potentially deep) changes in the future.

on the capacities: I actually think this is something we need to add as well for PE trade, because it could solve some of the problems we have in too fast reactivity of PE trade to price changes. In the real world, you need to build large terminals, pipelines, etc - which takes years to decades. REMIND just flips from one time step to the next.

Best,
Robert

Renato-Rodrigues:
Just adding to Robert's comments, the code is 90% compatible with not enforcing the trade, however I still need some time to test it to make sure it works, so for now the current implementation is enforcing the trade to happen to the limit of the predetermined existent trading capacity between two regions. If you need to adjust the trade patterns, just adjust the capacity parameter (pm_seTradeCapacity).

Exports price is currently set by the region specific costs for the secondary energy production, in contrary to a perfect competitive global market assumption used in the PE level. As long as se trade is not large enough to cause severe unbalances, this is a preferable way to determine export prices. Once trade is large enough, we will need to add an penalization mechanism to the solve algorithm to avoid zig-zag switch of trade patterns between iterations.

modules/01_macro/singleSectorGr/equations.gms
Renato-Rodrigues: I will double check that. Thks for the comment!

modules/24_trade/se_trade/bounds.gms
Renato-Rodrigues: Trade capacity limits the total trade potential between two regions. You can think of it as the pipelines transportation limits, regasification infrastructure currently installed and transmission power lines capacities that connect two countries/regions.

modules/01_macro/singleSectorGr/equations.gms
mleimbach: Renato, the original trade cost implementation was meant as a kind of iceberg cost, i.e. costs that have to be paid by someone, but there is no beneficiary. All other revenues from the PE trade is accounted for via the intertemporal budget constraint. If you include SE trade which have a wedge in export and import prices, you are probably doing right to include this kind of producer revenue in the budget equation. But I am not fully sure about the implied consequences for the overall accounting. Is the convergence fine?

Renato-Rodrigues: Thks @mleimbach
So far the convergence is fine because we are only adding very small capacities for trade SE. As mentioned in another comment also in this pull request, this needs to be updated for a more wider secondary energy trading scenario, but this will most probably be the challenge for a newcomer PhD/PostDoc.

PR #337, opened 2021-03-16
* bug-fix: adding missing electricity and hydrogen final energy transport accounting

Renato-Rodrigues:
* bug-fix: adding missing electricity and hydrogen final energy transport accounting
* updating REMIND21 config file


robertpietzcker (approved):
quick fix - works.

PR #340, opened 2021-03-19
Fix bad choice of snapshot

johanneskoch94:


fschreyer (commented):
Ok, good to see this fixed. Yet, I have not a real clue what it is about. Maybe better involve @giannou .

giannou:
looks okay to me but has it been tested?

johanneskoch94:
It's been tested. This check can also be removed completely if that is preferred. @0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q ?

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
I'm not sure why this test is there in the first place.
To me it looks like updating a package is an atomic operation.  R installs the new version in a temporary directory (inside `00LOCK.*`) and then replaces the old directory with the new one.  When the installation fails, the old package version is still available in the library, is it not?
Judging from recent history I would guess that getting a broken package from `/p/projects/rd3mod/R/libraries/3.6.3` is higher than failing to load a package because R relinks the directory _right in that moment_.  So either users want a stable library, and defines a library to be used a priori, or they roll dice and take what `/p/projects/rd3mod/R/libraries/3.6.3` gives them.

johanneskoch94:
You're right of course Michaja. However i think the problem occurs more often that you think, at least it did crash often enough for me to want a fix for that ;) So rolling the dice isn't really a sustainable solution in my opinion. 
Coming up with a default, stable-library, setup, that doesn't require the user to manually pick an R-library would be super cool.

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
> However i think the problem occurs more often that you think, at least it did crash often enough for me to want a fix for that ;)

I still don't underatand what crashes.  
Does loading the package fail when it is being updated or when the update failed?

johanneskoch94:
When it's being updated.

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
> When it's being updated.

I don't understand why that would happen, since the update seems to be an atomic operation.  Might do some testing.

In any case, the update is fine, though I would have a slight preference for using `list.files()' on aesthetic grounds and a more restrictive regex pattern on robustness grounds.  Right now you are assuming there will be no directory named `2021_test` which would score higher (i.e. later) in the alphabetic sort.

But merge.

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
> Might do some testing.

Turns out updating packages isn't atomic after all.

But I would go with 
```diff
-    snapshot_folder <- "/p/projects/rd3mod/R/libraries/snapshots/"
-    monthly_snapshots <- snapshots[grepl("^20.._", snapshots)]
-    lattest_monthly_snapshot <- paste0(snapshot_folder, monthly_snapshots[length(monthly_snapshots)])
+    latest_monthly_snapshot <- tail(
+        sort(list.files(
+            path = '/p/projects/rd3mod/R/libraries/snapshots/', 
+            pattern = '^20[0-9]{2}_[0-9]{2}(_[0-9]{2})?$',
+            full.names = TRUE)), 
+    n = 1)
```

johanneskoch94:
Thanks for the suggestions Michaja :) I've implemented them, removing only a superfluous "/" in the path argument.

johanneskoch94:
This PR is ready to merge in my opinion, so whoever feels like reviewing and merging.. go ahead! :)

giannou:
looks like there is a need for improvement: `/home/giannou/testremind/remind2/output/SSP2-AMT-Base_2021-03-23_11.46.14/log.txt`

giannou:
will this be solved with this PR? `/home/giannou/testremind/remind2/output/SSP2-AMT-Base_2021-03-23_11.46.14/log.txt`
```
The following lock folders were found at /p/projects/rd3mod/R/libraries/3.6.3 :
         00LOCK-mrcommons
That means that the  mrcommons package(s) is(are) currently being updated.
Packages will be loaded from the library's lattest snapshot instead:
 /p/projects/rd3mod/R/libraries/snapshots/test_mask
(If the lock folder isn't deleted automatically in the next couple of minutes, that means the package failed to update/install and that the folder has to be removed manually!)
Error in library(dplyr, quietly = TRUE, warn.conflicts = FALSE) :
  there is no package called 'dplyr'
Calls: suppressMessages -> withCallingHandlers -> library
Execution halted
```

johanneskoch94:
Yes!

giannou:
> Yes!

so it will not load this weird `test_mask` snapshot anymore but a proper one?

johanneskoch94:
exactly

giannou:
> exactly

cool beans

giannou (approved):

PR #352, opened 2021-04-01
DAC FE to use any fehes,feh2s,feels,fegas and see Industry FE tax

amerfort:
Heat for DAC material recovery will now be supplied by a mix of fehes, feh2s, fegas and/or feels.
DAC FE demands are now included in Industry FE taxes.
Results can be studied here:
/p/tmp/amerfort/remind/compScen-2021-04-01_16.26.49-H12.pdf

### DIFF
--- a/modules/33_CDR/DAC/equations.gms
+++ b/modules/33_CDR/DAC/equations.gms
@@ -16,12 +16,14 @@ q33_demFeCDR(t,regi,entyFe)$(entyFe2Sector(entyFe,"cdr")) ..
 ;
 
 ***---------------------------------------------------------------------------
-*'  Calculation of (negative) CO2 emissions from direct air capture. 
+*'  Calculation of (negative) CO2 emissions from direct air capture. The first part of the equation describes emissions captured from the ambient air, 
+*'  the second part calculates the CO2 captured from the gas used for heat production assuming 90% capture rate.
 ***---------------------------------------------------------------------------
 q33_capconst_dac(t,regi)..
 	v33_emiDAC(t,regi)
 	=e=
 	- sum(teNoTransform2rlf_dyn33(te,rlf2), vm_capFac(t,regi,"dac") * vm_cap(t,regi,"dac",rlf2))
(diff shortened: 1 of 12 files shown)


Renato-Rodrigues:
Just a general comment, using vm_otherFEdemand is not necessary anymore in any CDR module realizations, as you can use vm_demFeSector directly instead, which would be the preferable method. 
It would be good if someone could do this change to avoid unnecessary additional variables in the model.

modules/33_CDR/DAC/equations.gms
```diff
@@ -33,12 +35,36 @@ q33_emicdrregi(t,regi)..
 	v33_emiDAC(t,regi);
 	
 ***---------------------------------------------------------------------------
-*'  Calculation of energy demand of direct air capture. DAC demands feels and fehes
+*'  Calculation of electricity demand for ventilation of direct air capture.
```
fschreyer, line 38: I think, the implementation around heat and electricity demand equations can be simplified and written more flexibly. I write you on Mattermost.

modules/33_CDR/all/datainput.gms
```diff
@@ -6,12 +6,13 @@
...
-!!previously used fe demands
-!!p33_dac_fedem("feels") = 7.33;    
-!!p33_dac_fedem("fegas") = 36.667;
-!!p33_dac_fedem("feh2s") = 36.667;
+!!fe demand electricity for ventilation
+p33_dac_fedem_el("feels") = 5.28;
+!!fe demand heat for material recovery
+p33_dac_fedem_heat("fehes") = 21.12;
+p33_dac_fedem_heat("fegas") = 21.12;
+p33_dac_fedem_heat("feh2s") = 21.12;
```
fschreyer, line 14: Not sure whether we need this level of detail or if this is relevant but could there be some efficiency losses when burning gas/H2 to get heat instead of taking heat directly (fehes) from the district heating network? Our pegas to sehe conversion technology ``gashp`` has an efficiency of 0.8 currently.

fschreyer (changes requested):
Looks good! I would suggest some simplification in the implementation, though, unless you need this urgently.

fschreyer (approved):

amerfort:
DAC FE to use any fehes,feh2s,feels,fegas and add Industry FE tax

PR #353, opened 2021-04-06
H2 fix

fschreyer:
This adapts the H2 implementation in simple buildings and industry by making the additional td H2 cost dependent on the stationary H2 share in total stationary FE. This should mimic that low t&d H2 cost through the use of H2 pipelines are only reached when a sufficient amount of H2 relative to total FE is demanded either by buildings or industry. Otherwise, FE H2 sees additional cost for transportation.  Test runs with three different phase-out ranges can be found here

``/p/tmp/schreyer/Modeling/REMIND_H12/H2Fix/compScen-2021-04-02_14.14.07-H12.pdf``
Runs: 
``H2_05_3`` -> phase-out between 0.5% and 3% H2 share. 
``H2_1_5`` -> phase-out between 1% and 5% H2 share. 
``H2_2_10`` -> phase-out between 2% and 10% H2 share. 

I set the medium option of 1-5% phase-out as default. Can be discussed further. 

I only did H12 default sector calibration for now. If we agree on this, I could do another setup, too, before merging.


Renato-Rodrigues:
As a suggestion, I wouldn't create an specific equation like the q_H2Share_stationary one. 
You could instead use the q_shfe equation formulation as an example and create a generic equation for calculating specific energy carriers shares in stationary sectors. Then, you would just need to enable this equation only for H2 for now. This would allow for the same equation to be used in the future to calculate alternative shares without the need for code duplication or additional equations.

Just for the record, although I am in favor of calculating the share in respect to the total stationary FE, I am against changing its base to the total final energy instead of total gases (H2 + natural gas) as I mentioned in the mattermost mensagens.
This extra cost was created to reflect the additional cost of converting existing pipelines, distribution networks and/or degasification infrastructure from natural gas to hydrogen. If it is our intention to reflect with this additional cost the conversion of pipelines from natural gas to H2, as it was designed in the first place, it is straightforward to assume that the market share that matters is H2 versus gases and not H2 versus any kind of energy carrier. For example, why hydrogen infrastructure should be cheaper if something change the relative costs between liquids and solids? If liquids and solids have difference efficiency levels (as they have), a competitiveness change between them will change the total FE, and in the way it is proposed, this would have a consequence on the H2 infrastructure cost. This just does not make any sense to me.

PR #356, opened 2021-04-13
fix 085edc7 make industry/subsectors clinker shares work for H12 and EU regions (possibly others)

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
- clinker-to-cement ratios are still hard-coded and not included in
  mrremind
- using regi_group for EUR and NEU regions works with both H12 and EU
  regions


LaviniaBaumstark (commented):
There are still hard coded region names left (e.g. "IND", "JPN", "LAM"). Please also find a solution for those to be fully flexible in the regional aggregation

PR #361, opened 2021-04-15
adapt CES parameter download to new system with versioning 

giannou:
using git commit hashes


0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q (approved):

LaviniaBaumstark (commented):
I would put the CESversion right after the input data version

PR #364, opened 2021-04-23
Improved compare scenario for EDGE-T. Started adding new plot entries.

MariannaR:
Improvements in the compare EDGE-T plotting routine.


Loisel (approved):
Thanks for adding the emissions!

PR #369, opened 2021-04-27
update industry/subsectors clinker-to-cement ratio to be region setting agnostic

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
- use file p37_clinker-to-cement-ratio.cs3r from mrremind >= 0.56.1


LaviniaBaumstark (approved):
nice :-)

PR #381, opened 2021-04-30
Added correct trucks technologies to the mapping.

MariannaR:
I am pretty sure I already changed the mapping. Probably it got lost in the merging with REMIND-EU


Loisel (approved):
Yeah I also wonder... Aren't the categories in the EDGE-T reporting already the correct ones? Is there maybe a manual mapping in the reporting script?

PR #384, opened 2021-05-04
Update ETS CO2 price trajectory and calibration files and configs to work with external gdx paths

Renato-Rodrigues:
* Updating ETS CO2 price trajectory as linear projection departing from 2010 value.
* Updating REMIND 21 config files to work with new external gdx paths
* Updating calibration files


LaviniaBaumstark (commented):
did you also add the gdx files corresponding to the CES parameters?

Renato-Rodrigues:
> did you also add the gdx files corresponding to the CES parameters?

yes, I did that last week.

LaviniaBaumstark (approved):

PR #391, opened 2021-05-18
Add the compare scenarios for Ariadne.

MariannaR:
Not the whole file is included yet - but this is a start :)


scripts/output/comparison/Ariadne_output.R
```diff
@@ -0,0 +1,45 @@
...
+
+## names of the output files
+md_template = "AriadneComparison.Rmd"
+
+## create a txt file containing the run names
+write.table(outputdirs, paste0(outdir, "/run_names.txt"), append = FALSE, sep = " ", quote = FALSE,
+            row.names = FALSE, col.names = FALSE)
+
+file.copy(file.path("./scripts/output/comparison/notebook_templates", md_template), outdir)
+rmarkdown::render(path(outdir, md_template), output_format="pdf_document")
```
fschreyer, line 44: Would be good to also be able to choose for which region (probably mostly either DEU or EUR) one would want the pdf.

fschreyer (approved):

scripts/output/comparison/Ariadne_output.R
MariannaR: I am not sure how one does that given that no options are provided in `Rscript output.R` but I will try to find out!

PR #393, opened 2021-05-19
Improvements in the reporting.

MariannaR:
- Ariadne reporting added new plots.
- EDGE-T reporting removed deprecated parts following new versions of `remind2`.


fschreyer:
Somehow I cannot comment in lines. I just noticed that [these variables](https://github.com/remindmodel/remind/blob/f937c1039da76f48fc9d86893c21f215820127fe/scripts/output/comparison/EDGEcomparison.R#L220) are deprecated. You should use ``FE|Transport|Liquids|+|Biomass/Fossil/Hydrogen``. However, the distinction in LDV and HDV does not exist yet but would be straightforward to add [here](https://github.com/pik-piam/remind2/blob/8b68716518e1f71c4a729449012084ad98b8bf61/R/reportFE.R#L191).

PR #394, opened 2021-05-21
delete deceptive MAICC files that are overwritten anyways

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
- R scripts and bau/550/450 .inc files in ./core/magicc are overwritten when
  MAGICC is copied from <magicc_template>
- changes in these files therefore have no effect, so they shouldn't be
  tracked, not to give the false impression that changes would have an effect


LaviniaBaumstark (approved):
nice, cleaning up!

PR #398, opened 2021-05-31
Merging of ariadne branch

fschreyer:
This includes:

* changes and config for latest ariadne runs
* some minor regipol changes: removed old emissions variables, made code work with H12 regionmappings

PS: many commits below are working commits when changing branches/pulling, a couple of them overwrite each other in the cleaning up process


modules/47_regipol/regiCarbonPrice/bounds.gms
```diff
@@ -42,8 +42,10 @@ $ENDIF.CCSinvestment
 
 
 ** Force historical bounds on coal
-    vm_cap.up("2020",regi,"pc","1")$((cm_startyear le 2020) and (sameas(regi,"DEU"))) = 38.028/1000;
-    vm_cap.up("2020",regi,"pc","1")$((cm_startyear le 2020) and (sameas(regi,"UKI"))) = 1.3/1000; !!2019 capacity = 7TWh, capacity factor = 0.6 ->  ~1.35GW -> Assuming no new capacity -> average 2018-2022 = ~ 1GW
```
Renato-Rodrigues, line 46: any specific reason why you removed the UKI bound? This is necessary to have better approximation to real world values.

main.gms
```diff
@@ -478,42 +482,47 @@ $setGlobal c_regi_capturescen  all !! def = all
 $setGlobal c_regi_synfuelscen  all !! def = all
 $setGlobal c_regi_sensscen  all !! def = all
 
+
+
 cm_TaxConvCheck = 0; !! def 1, which means tax convergence check is on
 																	  
-cm_biotrade_phaseout = 0; !! def 0
```
Renato-Rodrigues, line 483: this is minor, but I wouldn't commit in the main gms any change on switches initializations to numbers different from the default values. I don't mind if you want to keep those, but I think it is good practice to keep those unchanged and avoid everybody just overwriting those over and over in every commit without necessity.

modules/47_regipol/regiCarbonPrice/bounds.gms
```diff
@@ -42,8 +42,10 @@ $ENDIF.CCSinvestment
...
+vm_cap.up("2020",regi,"pc","1")$((cm_startyear le 2020) and (sameas(regi,"DEU"))) = 38.028/1000;
+*** 2019 capacity = 7TWh, capacity factor = 0.6 ->  ~1.35GW -> Assuming no new capacity -> average 2018-2022 = ~ 1GW
+*** Commented out as this line causes issues for H12 regionmappings
```
Renato-Rodrigues, line 47: I just saw this. I don't get why this would cause issues with H12 as if that was the case, the DEU constraint would also cause issues.

fschreyer: Ye, this causes issues with H12 regipol as I said. Can you please fix this? I would commit it now such that it works also for others.

main.gms
fschreyer: Ye, slips through when I add chunks with ``git add -pu``

modules/47_regipol/regiCarbonPrice/bounds.gms
Renato-Rodrigues: Again, I don't see any reason why this would cause any issues with H12. 
The region specific coding is inside a sameas conditional. If this would cause issues, we would have a lot of more issues everywhere in the code.
Sorry, but I don't see anything to be fixed here.

Renato-Rodrigues (approved):

PR #400, opened 2021-06-02
Update comparison notebook with ariadne/region-specific plots

fschreyer:
This updates and harmonizes the ariadne notebook plots. It adds if clauses to EDGE-T/industry subsectors plots which require specific variables and also adds a switch if you want to see historic data for some plots or not.

### DIFF
--- a/scripts/output/comparison/notebook_templates/AriadneComparison.Rmd
+++ b/scripts/output/comparison/notebook_templates/AriadneComparison.Rmd
@@ -6,9 +6,9 @@ output:
   html_document: default
 ---
 
-# Data analysis and plotting using quitte
+# Regional Data Analyses for REMIND ARIADNE Runs
 
-Authors: Gunnar Luderer (luderer@pik-potsdam.de), Felix Schreyer, Falk Benke, Marianna Rottoli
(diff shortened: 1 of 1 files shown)


fschreyer:
@giannou 

Some background to this: Marianna and I were adding another comparison script [output script](https://github.com/remindmodel/remind/blob/develop/scripts/output/comparison/Ariadne_output.R) for ariadne-specific analysis as a couple of people are working with the same plots. It is based on a R notebook which is rendered by the script and produces png files in a seperate comparison folder. This could be useful for the whole group, in general, as these plots are quite generic and helpful for everybody who wants to have plots for just one region which are already in a neat format to put into ppts etc. The compare scenario pdf is not always ideal for that. Moreover, it has the advantage that it is a plot collection that is shared (as compare scenario pdf), but in addition easy to handle and modify for your own use locally. 

What would be great, in addition, would be if we could add some more arguments that the user can specify when running this. At the moment, it just produces the output for the region "DEU" automatically. Do you happen to know how to implement a question like "Choose a region ..." in the beginning which the user can choose and which will be passed to the notebook as a parameter? Thanks a lot :)

MariannaR (approved):

PR #406, opened 2021-06-10
adding marginal adjustment cost (adj_cost per added cap) in postsolve

cchrisgong:
calculated from taking the first derivative: d(v_costInvTeAdj) / d(vm_deltaCap)

### DIFF
--- a/core/declarations.gms
+++ b/core/declarations.gms
@@ -182,7 +182,7 @@ p_cintraw(all_enty)                                  "carbon intensity of fossil
 p_CapFixFromRWfix(ttot,all_regi,all_te)              "parameter for fixing capacity variable to Real-World values in 2010/2015"
 p_deltaCapFromRWfix(ttot,all_regi,all_te)            "parameter with resulting deltacap values resulting from fixing capacity to real-world values in 2010/2015"
 
-p_adjCostInv(ttot,all_regi,all_te)                   "adjustment cost to technology investment"
+p_adjCostInv(ttot,all_regi,all_te)                   "marginal adjustment cost calculated in postsolve for diagnostics"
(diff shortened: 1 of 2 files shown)


core/declarations.gms
```diff
@@ -182,6 +182,7 @@ p_cintraw(all_enty)                                  "carbon intensity of fossil
 p_CapFixFromRWfix(ttot,all_regi,all_te)              "parameter for fixing capacity variable to Real-World values in 2010/2015"
 p_deltaCapFromRWfix(ttot,all_regi,all_te)            "parameter with resulting deltacap values resulting from fixing capacity to real-world values in 2010/2015"
 
+p_adjCostInv(ttot,all_regi,all_te)                   "adjustment cost to technology investment"
```
fschreyer, line 185: I think the description can be confused with the actual parameters for adjustment cost. Better say something like "marginal adjustment cost calculated in postsolve for diagnostics".

fschreyer (changes requested):

core/declarations.gms
cchrisgong: haha good point, it is changed to your suggestion. I don't think it's too jargon-y :P

fschreyer (approved):

Renato-Rodrigues:
If it is only an output "reporting" parameter you should use "o_" instead of "p_" according the code etiquette, no?
Also, if it is only for reporting reasons, moving it to the R reporting scripts are recommended to avoid cluttering the model code.

cchrisgong:
yes, I can change the p_ to o_ 
The formula would be very very complex to rewrite in R, GAMS in this case I think is clearer, cleaner and easier to check for errors for future coders

Renato-Rodrigues:
By the looks of the files commited, you forgot to rename the parameter declaration to use the preffix o_ before merging this to remind. I would guess this PR breaks remind.

PR #409, opened 2021-06-17
Updating the VRE integration modeling

robertpietzcker:
- Reduced capFac for ngt and hydrogen turbines to better represent their role of peaker units in medium-high VRE worlds
- to keep cap in line with power sector model results, increased the "flex" factor for ngt and h2turb to counter the reduction of capFac
- Reduced the VRE-connected building of electrolysis to make sure the model doesn't create oversupply
- Shifted the requirement to build elh2VRE to elh2, so that we don't have two separate techs anymore - before the learning of elh2 didn't benefit from the VRE-connected electrolysis deployment. This also meant taking the electrolysis costs out of the investment costs for the "storX" meta-technologies (because before the electrolysis costs were included in the storX costs and the dummy tech elh2VRE had costs of 1).  Bounded the cap of elh2VRE to 0 so the model doesn't see it anymore as a usable technology.
- Also reduced the efficiency of electrolysis from 0.8 to 0.73 to better represent near-term expected tech specifics
- Also increase learning part of electrolysis so that results are closer to recent analysis - ~400-500$/kW in 2050 as documented in the Ueckerdt et al 2021 NCC paper on hydrogen


core/datainput.gms
```diff
@@ -428,14 +428,25 @@ pm_cf(ttot,regi,te) =  f_cf(ttot,regi,te);
 *RP short-term fix: set capacity factors here by hand, because the input data procudure won't be updated in time
 pm_cf(ttot,regi,"apcardiefft") = 1;
 pm_cf(ttot,regi,"apcardieffH2t") = 1;
-pm_cf(ttot,regi,"h2turbVRE") = 0.15;
+***pm_cf(ttot,regi,"h2turbVRE") = 0.15;
```
Renato-Rodrigues, line 431: Could this be also added to the calcCapacityFactor function in mrremind?

This way, once new data input is created and the f_cf.cs3r file is updated this line could be simply removed from the code, and we would keep having a single place with all definitions of capacity factors assumptions.

The same would apply to the other "hard-coded" fixings that were already present in this section of the code.

Renato-Rodrigues (approved):

modules/32_power/IntC/bounds.gms
```diff
@@ -65,6 +65,8 @@ loop(regi$(p32_factorStorage(regi,"csp") < 1),
   v32_shSeEl.lo(t,regi,"csp")$(t.val > 2100) = 2;
 );
 
+*** Fix capacity to 0 for elh2VRE now that the equation q32_elh2VREcapfromTestor pushes elh2, not anymore elh2VRE, and capital costs are 1
+vm_cap.fx(t,regi,"elh2VRE",rlf) = 0;
```
Renato-Rodrigues, line 69: Is elh2VRE still useful by any reason or could it be completely removed from the model?

core/datainput.gms
robertpietzcker: Yes, exactly. 
I currently can't find the time to do it via mrremind, so when I saw that there were still old cf corrections in the code, I went the easy route, thinking "at some point of time someone will move all these changes here to mrremind"

modules/32_power/IntC/bounds.gms
robertpietzcker: in the current version it is not used anymore. 

I wasn't sure if in a few month somebody gets an idea why we might need it again, so I thought I don't take it out now only to reintroduce it in a few months.

core/datainput.gms
LaviniaBaumstark: o.k. for now, but I hope that this preliminary solution will not stay too long

PR #413, opened 2021-06-22
hot fix: switching off wind offshore in default.cfg

cchrisgong:


Renato-Rodrigues (approved):
Approved. Next time you need to be more careful with your pull request to avoid the need for these hot-fixes. The reviewer will never be able to check in detail every change made, so it is responsibility of whorever makes the code changes to test them before submitting the pull request.

PR #414, opened 2021-06-23
Add emission factors for fegat.

Loisel:
Adds the missing CO2 emission factor for gases used in transportation (`fegat`). `fegat` is only used in the `edge_esm` realization of transport. There is not much used there, but from time to time the missing emissions bump up.

For simplicity, I used the same value as the one used for gases in buildings.


MariannaR (approved):

Renato-Rodrigues (approved):
Thks for fixing this!

PR #415, opened 2021-06-23
Add variable efficiencies for Electrolyzers

Loisel:
Adjust `elh2` in a time-dependent way, according to a PSI literature review: efficiency is 0.57 for 2010 and 0.73 for 2050, with a linear adjustment in-between. I then set 0.75 for 2055 onwards.

The values will also be part of the following publication:
Electricity storage and hydrogen: Technologies, costs and environmental burden, Christian Bauer (ed.), Thomas Heck, Simon Schneider, Harshil Desai Sanjaykumar, Tom Terlouw, Karin Treyer, Xiaojin Zhang, 2021.


Loisel:
Of course, endogenous learning would be best for this important technology.

robertpietzcker:
> Of course, endogenous learning would be best for this important technology.

it IS an endogenous learning technology already.
 I think Felix implemented that some months ago.

Loisel:
> > Of course, endogenous learning would be best for this important technology.
> 
> it IS an endogenous learning technology already.
> I think Felix implemented that some months ago.

I was referring to learning which improves efficiencies, sorry for not being clear. Afaik REMIND learning only reduces costs (?).

fschreyer (approved):
Thanks for this, Alois! Of course, coupling the efficiency to the endogenous learning of capital cost would be the best long-term solution but this already helps.

PR #417, opened 2021-06-23
Remove unused realizations from 45_carbonprice and cm_iterative_target_adj

gabriel-abrahao:
- [x] Remove listed switches for cm_iterative_target_adj
  - [x] Add (deprecated) comments in default.cfg
  - [x] Delete relevant code in postolve.gms
- [x] Remove realizations of 45_carbonprice
  - [x] Add (deprecated) comments in default.cfg
  - [x] Delete subfolders from the module folder
  - [x] Delete references in module.gms

Deletes the following switches/realizations:

- cm_iterative_target_adj   
  - (1): iterative adjustment I for tax based on 2030 emissions or for 2150 budget based on 2100 budget result, see core/postsolve.gms for direct algorithms of adjustment
--> old implementation from AMPERE, not needed anymore
  - (4): iterative adjustment IV for both budget and tax runs based on CO2 FF&I emissions 2010-2100, see core/postsolve.gms for direct algorithms of adjustment
--> still needed?

- Carbon price module:

  -"NDC2constant"
  -"diffPhaseIn2Constant"
  -"diffPhaseIn2Lin"
  -"diffPhaseInLin2LinFlex"


gabriel-abrahao:
This makes ``cm_cm_peakBudgYr`` and ``cm_expoLinear_yearStart`` not used in the 45_carbonprice module, but there are still cases where they are adjusted in the core postsolve. Are they important somewhere else?

gabriel-abrahao:
@LaviniaBaumstark

LaviniaBaumstark (commented):
you can directly delete deprecated realizations in default.cfg

LaviniaBaumstark (commented):
Did you try a test run to ensure that the code compiles?

giannou:
As a first-time contributor opened the PR an admin had to activate the checks. They are failing, by the way, please correct the errors

gabriel-abrahao:
``cm_peakBudgYr`` and ``cm_taxCO2inc_after_peakBudgYr`` were only used in the deleted realizations and were still in the not_used files of the remaining ones. However, even after removing them from the not_used files codeCheck raises this error:

```
Error in gms::codeCheck(strict = T, interactive = T) : 
  codeCheck returned warnings. Fix warnings to proceed!
In addition: Warning messages:
1: cm_peakBudgYr appears only in "core" even though it is supposed to be an interface (perhaps only in not_used.txt?)! 
2: cm_taxCO2inc_after_peakBudgYr appears only in "core" even though it is supposed to be an interface (perhaps only in not_used.txt?)! 
```

These variables are adjusted in ``core/postsolve.gms`` and ``core/preloop.gms`` in all cases with ``cm_iterative_target_adj == 9``, apparently in all realizations of 45_carbonprice. Does this error means they are supposed to be declared inside a realization or something?

LaviniaBaumstark:
did you run codeCheck in the interactive mode? It should "automatically" adjust the not_used.txt files in the right way.

LaviniaBaumstark:
if there are no left-overs, the switches cm_peakBudgYr and cm_taxCO2inc_after_peakBudgYr only were interfaces for the deleted realizations. In this case, check where they are used and rename accordingly (e.g. "cm_" to "c_" if they are only used in the core)

gabriel-abrahao:
> 
> 
> if there are no left-overs, the switches cm_peakBudgYr and cm_taxCO2inc_after_peakBudgYr only were interfaces for the deleted realizations. In this case, check where they are used and rename accordingly (e.g. "cm_" to "c_" if they are only used in the core)

I did use the interactive codeCheck, and there are no mentions left in not_used.txt or in the other realizations, so now it is only used in the core.

However, it seems like these switches are used in ``cm_iterative_target_adj` == 9` which is the ``REMIND 2.1 default for validation peakBudget runs, in combination with carbonprice = diffCurvPhaseIn2Lin``. I'm pretty new to this but this sounds like a common setup. Is there a risk that changing the name of the switches will break someone's production runs that are configured with the "old" name? If you don't think so I'll go ahead and rename them.

LaviniaBaumstark:
renaming should be o.k. if you rename at all places in the code (also do not forget the .R and .csv files). In this case as it is no parameter but only a switch, I assume that it is not often used. In addition, you should check whether this switches are used in the reporting (remind R-package) ensuring consistency of the reporting and the model code. If you need to make adjustments in the remind R-package they should be backward compatible.

gabriel-abrahao:
> 
> 
> renaming should be o.k. if you rename at all places in the code (also do not forget the .R and .csv files). In this case as it is no parameter but only a switch, I assume that it is not often used. In addition, you should check whether this switches are used in the reporting (remind R-package) ensuring consistency of the reporting and the model code. If you need to make adjustments in the remind R-package they should be backward compatible.

Would that be the [remind ](https://github.com/pik-piam/remind) or the [remind2](https://github.com/pik-piam/remind2) package?

LaviniaBaumstark:
ah, right, remind2

gabriel-abrahao:
Renamed the `cm_peakBudgYr` and `cm_taxCO2inc_after_peakBudgYr` to `c_peakBudgYr` and `c_taxCO2inc_after_peakBudgYr` since they are now used only in the core. They were also renamed in all `scenario_config_*.csv` files that were in the `config` folder so they still work. Maybe that will lead to merge conflicts?

Test runs with `cm_iterative_target_adj` 7 and 9, which use these switches, yielded results very close to the same runs made before the change.

I didn't find any reference to them in both `remind` R packages codebases. Is there a test that can be done to make sure this renaming will be compatible with them?

LaviniaBaumstark:
we have no process for directly testing the reporting but as your test-runs lead to same results the reporting should have worked for them. The whole workflow also will be tested by your weekly automated model test scenarios. 

I would not expect any merge conflicts if you rename switches in the scenario_config* files

LaviniaBaumstark:
there are still errors in codeCheck, please take care

gabriel-abrahao:
> 
> 
> there are still errors in codeCheck, please take care

Strange, those are not showiing up when I run codeCheck on the cluster.

gabriel-abrahao:
Apparently the issue was that I didn't commit my `main.gms`, should I always commit changes inside the "warning zone"? That led to a merge conflict but I think it's resolved now, sorry about that!

LaviniaBaumstark:
the warning zone is there indicating, that you should not make there changes ( to the default settings) by hand because this file is updated automatically when you start a run based on the default.cfg. But if you rename a switch or add a new one, you have to adjust the main.gms.

LaviniaBaumstark (approved):

PR #419, opened 2021-06-25
Merge ariadne develop branch into trunk

fschreyer:
This PR merges the changes made in the ariadne branch back to the trunk. 

The changes are

* adding switches ``cm_FEtax_trajectory_abs`` and ``cm_FEtax_trajectory_rel`` to flexibly rescale FE tax trajectories to different levels 
* slightly reformulate calculation of co2 price rescaling factor beween iterations and convergence criteria of emissions targets in regipol module to be able to improve target convergence for
  - year targets for close to zero emissions
  - year target for below zero emissions
* add switch ``cm_regipol_slope_beforeTarget`` in regipol module: allows adjustment of the shape of the co2 price trajetory to some extent to have a "kink". You can define a year after which the co2 price slope should be higher than before. This helps in reaching given budget and year targets at the same time. 
* make markup cost in buildings to be scalable by switches (``cm_heatPumpMkup_build`` etc.)
* modify convergence criterion for nash markets to only take into account surpluses from cm_startyear onwards
* some minor adjustments for Germany (coal capacity factors, near-term VRE expansion, ...)


LaviniaBaumstark (commented):
please do not use card coded names of regions like "DEU" in the code, especially in not in the core

fschreyer:
> 
> 
> please do not use card coded names of regions like "DEU" in the code, especially in not in the core

Well, how else would I do it?

I see, these formulations are not nice but they do not break H12 because of the conditional formulation. 

I agree that at some point this needs to be done for all regions and partially even in mrremind but this always takes much time and does not serve our purpose. I also think, it needs an overall strategy how to extent particular regional changes to global ones.

Renato-Rodrigues:
Maybe I am too late for the conversation, but just to give my opinion, if you must use hard coded regions, move it preferably to inside the regipol module.
For the capacity factor specific case, it is quite easy to adapt the input data to use a region specific capacity factor. If your assumption is a general improve to represent capacity factors for a specific technology in a specific region, juts add it directly to the calcCapacityFactor function instead with comments why this regions has a different values from others.

LaviniaBaumstark:
I agree with Renato. Best would be to either add regional specific adjustments either to the input data generation or find a general formulation which would help for all regional resolutions. If this is not possible, please move regional specific pat into the regipol module

MariannaR (approved):

fschreyer:
Well, I think that moving it to the regipol module also has downsides as the module has two realizations which are both needed such that I would need to add the same code twice. 

I guess, I see your point, though, that you want to have all hard-coded regions in one place to not forget about them. 


In general I see it like this: In the medium term, some of these changes we will want to extent to other EU regions, some to the whole world. Others, we will probably just want to leave them as they are for one region. Once the time has come and we decide to clean up all hard-coded regions, I am willing to help translating these statements into mrremind. In the meantime, I would be in favor of not being too strict about hard-coded regions as it is quite impractical to do it via mrremind for every small adjustment.

fschreyer:
@Renato-Rodrigues 

I moved all hard-coded regions from the core to regipol. Could briefly you check if you see any issues with this [commit]https://github.com/remindmodel/remind/pull/419/commits/47a48f51afe2f2d00ab33543192c2c316f25e079)?

@MariannaR 

I updated the ariadne plots with [this commit](https://github.com/remindmodel/remind/pull/419/commits/82c16f925fdd517f185b5eebef7105dab0974a2d). I removed again your two plots below because they actually were already there in the parts above (or I missed something?).I hope I have everything now because I was actually developing this locally unmanaged for some time, sry for that. However, you can check the plots here and both plots you recently added are still there: ``/p/tmp/schreyer/Modeling/REMIND_H12/ariadne/output/compareRunAriadne_2021-06-28_13.26.12``. Maybe just take a glance and let me know if stuff is missing.

Renato-Rodrigues:
> @Renato-Rodrigues
> 
> I moved all hard-coded regions from the core to regipol. Could briefly you check if you see any issues with this [commit](https://github.com/remindmodel/remind/pull/419/commits/47a48f51afe2f2d00ab33543192c2c316f25e079)?

It looks fine as far as I can see. Did you make a test run and compared the results to the previous formulation see if tthe results remain unchanged?

LaviniaBaumstark (approved):

fschreyer:
> It looks fine as far as I can see. Did you make a test run and compared the results to the previous formulation see if tthe results remain unchanged?

Did a quick test. Looks good. It is only this part of ``cm_altFeEmiFac`` which is affected in addition. This is the version before (``/p/tmp/schreyer/Modeling/REMIND_H12/ariadne/output/xx_Calib_FE_med_2021-06-29_08.54.48``), this is the version after moving (``/p/tmp/schreyer/Modeling/REMIND_H12/ariadne/output/xx_Calib_FE_med_2021-06-29_07.52.24``). ``pm_emifac`` is the same in both so it should not have changed. Although, I am not entirely sure this part does what you wanted in the first place because ``pm_emifac`` has different values than specified in those [datainput lines](https://github.com/remindmodel/remind/blob/9db359d7f2c4d6508198802fe53edc5997156b58/core/datainput.gms#L1353) but maybe that's due to some required conversions afterwards.

Renato-Rodrigues:
The model used emifac should be different from these input pm_emifac values because in the model level they are discounted by the end-use (final energy) emission factors. The original input values correspond to the "total" emissions, without the split between pe2se and fe emissions. 
This split is only done in the preloop, at the 05_initialCap module. 

If you just compare the results between the run with and without your change although, they should be exactly the same, and this is what matters.

PR #424, opened 2021-06-30
 Add switch to lower German agricultural ch4 and no2 baseline emissions, add CDR variable and bug fixes in CDR module

amerfort:
Future base line ch4 and no2 emissions shall be reduced to align remind projections with current literature that includes behavioral change. Add a variable summing all negative emissions (i.e. from BECCS and Afforestation as well as from the CDR module). Minor bug fixes in the CDR Module.


core/equations.gms
```diff
@@ -623,6 +623,26 @@ q_emiMac(t,regi,emiMac) ..
...
 
+***--------------------------------------------------
+*' All CDR emissions summed up
+***--------------------------------------------------
+q_emiCdrAll(t,regi)..
+  vm_emiCdrAll(t,regi)
+       =e= !! BECC + DACC
+  (sum(emiBECCS2te(enty,enty2,te,enty3),vm_emiTeDetail(t,regi,enty,enty2,te,enty3))
+  + sum(teCCS2rlf(te,rlf), vm_ccs_cdr(t,regi,"cco2","ico2","ccsinje",rlf)))
+  !! scaled by the fraction that gets stored geologically
```
fschreyer, line 634: I defined a parameter for that share, see [here](https://github.com/remindmodel/remind/blob/92dbf12c183cb2e8f337c9568a3ce1b31c6b81a9/core/postsolve.gms#L674) which would make it shorter.

core/equations.gms
```diff
@@ -623,6 +623,26 @@ q_emiMac(t,regi,emiMac) ..
...
+  * (sum(teCCS2rlf(te,rlf),
+        vm_co2CCS(t,regi,"cco2","ico2",te,rlf)) /
+  (sum(teCCS2rlf(te,rlf),
+        vm_co2capture(t,regi,"cco2","ico2","ccsinje",rlf))+sm_eps))
+  !! net negative emissions from co2luc
+  -  p_macBaseMagpieNegCo2(t,regi)
+       !! negative emissions from the cdr module that are not stored geologically
```
fschreyer, line 641: I am not sure whether that is correct as ``vm_ccs_cdr`` refers to the total carbon *captured* by DAC, not the stored carbon, right?

I would think ``pm_share_CCS_CCO2 * vm_ccs_cdr`` would give you the DACCS.

modules/47_regipol/regiCarbonPrice/bounds.gms
```diff
@@ -79,9 +79,10 @@ loop(regi$(sameAs(regi,"DEU")),
 
 *** only small amount of co2 injection ccs until 2030 in Germany
 vm_co2CCS.up(t,regi,"cco2","ico2",te,rlf)$((t.val le 2030) AND (sameas(regi,"DEU"))) = 1e-3;
-*** no Pe2Se fossil CCS in Germany, if cm_noPeFosCCDeu = 1 chosen 
-vm_emiTeDetail.up(t,regi,peFos,enty,te,"cco2")$((sameas(regi,"DEU")) AND (cm_noPeFosCCDeu = 1)) = 1e-4;
-
+*** no Pe2Se fossil CCS in Germany, if c_noPeFosCCDeu = 1 chosen
```
fschreyer, line 82: Cool, so you put all your hard-coded Germany references in regipol? That will please Lavinia! Do you only need these bounds in case you have regional emissions target and regipol = ``regiCarbonPrice``? If you also need it in BAU runs you would also need these bounds in the ``none`` realization of regipol.

fschreyer (approved):
Looks good. Did you test whether a default run starts with ``Rscript start.R``? Can be merged then. The other issues I mentioned do not need to be adjusted now, I guess.

PR #437, opened 2021-07-07
adding convergence checks for ETS and ESR targets

Renato-Rodrigues:
* adding convergence checks for ETS and ESR targets
* splitting regional from global targets in solution report and improving description.


robertpietzcker (approved):
looks ok. 
I would propose to in the future shift here as well to using the 2015 emission levels as the denominator for the deviation to get a more robust solution. Or maybe introduce an offset, as I agree with the basic idea that close to a target of 0 emissions, price changes likely need to be larger than close to a very high emission target.

fschreyer:
Dear all,
    I am on vacation and will be back in office by July 26. 
    
    Best regards,
    Felix Schreyer
    
    -- 
Felix Schreyer
Doctoral Researcher
RD3 Transformation Pathways

Potsdam Institute for Climate Impact Research
Telegraphenberg A 56
14412 Potsdam  (Germany)
Phone:  +49 (0) 331 288 2419

PR #439, opened 2021-07-09
Enforcing biomass share in FE carriers for buildings and industry based on IEA data

Renato-Rodrigues:
* Enforces biomass share in FE carriers for buildings and industry based on IEA data

### DIFF
--- a/core/declarations.gms
+++ b/core/declarations.gms
@@ -113,7 +113,7 @@ p_ef_dem(all_regi,all_enty)                                 "Demand side emissio
 p_bioshare(tall,all_regi,all_enty)                          "bioshare for each FE carrier"
 pm_demPeBio(tall,all_regi)                                  "Primary energy bioenergy demand from last iteration [TWyr]"
 
-pm_secBioShare(tall,all_regi,all_enty,emi_sectors)           "share of biomass per carrier for each sector"
+pm_secBioShare(ttot,all_regi,all_enty,emi_sectors)           "share of biomass per carrier for each sector"
 
 p_avCapFac2015(all_regi,all_te)                             "average capacity factor of non-bio renewables in 2015 in REMIND"
 p_aux_capToDistr(all_regi,all_te)                           "aux. param. to calculate p_avCapFac2015; The historic capacity in 2015"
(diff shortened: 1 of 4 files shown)


fschreyer:
Just a comment on the calculation of the target deviation in your ETS/ESR implementation (I know we disagreed here earlier ;-)) [here](https://github.com/Renato-Rodrigues/remind/blob/f49e623857c4cdf405c9dd8174522327e395ed52/modules/47_regipol/regiCarbonPrice/postsolve.gms#L50). For small (like 1-10 MtCO2) and negative p47_regiCO2ETStarget you might get convergence problems with this. My suggestion would be to divide by 2015 or 1990 historic emissions to avoid ``pm_ETSTarget_dev`` to become super large (even though you are close to the target) or negative.  Not sure whether I am missing any problems with that. We can discuss at some point again once I am back.

Renato-Rodrigues:
@fschreyer Thks! I actually agree more with your implementation than the current one, but I decided to leave the ETS and ESR implementations unchanged for now because I think we can do even better to assure a faster convergence now that we have a better knowledge of the mitigation response of the model against different carbon price levels.
Let's discuss this as soon as we both have some more available time.

robertpietzcker:
I agree - using 1990, 2005, or even better, 2015,  emission values as comparison point will be more robust.

robertpietzcker (commented):
a few points that I think would be good to improve: 

a) I don't get the time dimension - why does the equation work also for future t? I mean from the calculation in module 04 I would guess it is empty?
Ahhh. 
now I get it.
To make it easier to understand when the equation counts, please add 
"*** (or whatever the current commenting sign is) This equation is only active for time steps for which historic values from the IEA exist. Otherwise, pm_secBioShare will be calculated to 0, thus $(pm_secBioShare) will turn off the equation. "
right after the header 
*' Enforce historical data biomass use per carrier in sector final energy



b) I think the convention is to define the equation and parameter only over ttot, not tall, whenever possible,  to prevent high resource use when filling in zeros

robertpietzcker (approved):

robertpietzcker (approved):
I think when we add such an exception to the model, it helps to add a comment to make sure a reader 2 years later will know this should be removed...

so just add to this line
!! this is just a short-term workaround until the input data routines can be fixed so that pm_fedemand is again the same as f04_IO_output

robertpietzcker (approved):

PR #442, opened 2021-07-12
Rename land-use CO2 emissions variable in start script according to c…

merfort:
…hanges in MAgPIE

In Magpie there is now an explicit distinction between direct and indirect land-use based CO2 emissions. `Emissions|CO2|Land|+|Land-use Change` from MAgPIE reporting only contains the direct part. The indirect effects (`Emissions|CO2|Land|+|Indirect`, such as carbon uptake in forests due to higher CO2 concentration) are already reflected in the carbon budgets. This is why they must not be added again to the budget, so in the coupling REMIND only reads in the direct part of the CO2 emissions from MAgPIE.


nicobauer:
Simple renaming. No problem.

PR #444, opened 2021-07-13
update input cache: improved historical capfac

cchrisgong:
- make sure historical capfac is calculated using gen(t)*2/(cap(t) + cap(t-1)) for the "added capacity effect"

- capfac is averaged over 5 years correctly

- remove capfac=0 from average

- remove too small standing capacity from data point (<200MW) (so artefact won't skew future projection)


Renato-Rodrigues:
Could you provide a short description of the changes made in the historical capacity factors in this new input data?

cchrisgong:
there is a list of description here: https://github.com/pik-piam/mrremind/pull/96

Renato-Rodrigues (approved):

PR #445, opened 2021-07-13
fix sectoral mapping not found

pfuehrlich-pik:


merfort (approved):
Looks good. Test run was successful.

PR #447, opened 2021-07-13
Limit rows and columns print-out in preloop models

cchrisgong:


Renato-Rodrigues:
Wouldn't it be better to avoid the code repetition and instead add it a single time to the core/preloop.gms file instead?

PR #450, opened 2021-07-15
Bugfix scenario config

flohump:
The column `path_gdx_ref ` was wrong for all scenarios except for SSP2 in `scenario_config_SSPSDP`


merfort (approved):
Good catch!

PR #451, opened 2021-07-15
Adding adjustment cost to gdx output

cchrisgong:
for model diagnostics: adding average adjustment cost, ratio of adjustment cost and direct investment cost to the output


robertpietzcker (commented):
great!

I would change 
o_avgAdjCost_2_InvCost_ratio
to
o_avgAdjCost_2_InvCost_ratioPc   (for Percent)

and multiply by 100 in the equation. 

Reason: it fits the description ("in %") and most other values are >1 in the reporting. Keeping ratios in % makes formatting faster. 
I know this isn't the standard yet, but you have both in the reporting, so I would propose to use it from now on

robertpietzcker (approved):

robertpietzcker (approved):

cchrisgong:
Thanks for checking: I forgot to add the x 100%. Now it is added

PR #457, opened 2021-07-26
udpate input data version to fix SSP2-NDC infeas

giannou:


amnmalik:
Hi @giannou , I only tested my bugfix but can't vouch for other changes that might have happened from 6.222 to 6.223.

PR #462, opened 2021-07-27
Add elh2VRE efficiencies in line with elh2

Loisel:


Loisel:
What's really strange: there are efficiencies in the XLS file for `coalhp` and `gashp`. According to history they were never part of the  `.prn` file. Any ideas @LaviniaBaumstark maybe?

fschreyer:
The ``elh2VRE`` technology is deprecated since this [commit](https://github.com/remindmodel/remind/commit/2cf7e1c12768c08eb745e7a2a4c6b7d98c6e9099). So changing it is nice but will only have an effect if we should ever decide to use ``elh2VRE`` again.

Loisel:
ah good to know! So I can close this PR? Any news on the *lost* columns coalhp and gashp?

PR #469, opened 2021-08-05
Remove stationary module

pfuehrlich-pik:
This PR removes the 38_stationary module. There's a successful default REMIND run on the cluster at `/p/tmp/pascalfu/remindmodel-fork/output/default_2021-08-18_10.04.26/`. I'm a GAMS newbie, so please be extra careful when reviewing this PR. Also, there's a related PR for remind2 here: https://github.com/pik-piam/remind2/pull/87


LaviniaBaumstark (approved):

LaviniaBaumstark:
I think, we now have the standalone-option for modules and would not use the `off` realizations for this case any more. If we will see the need for this realization in future, we can easily re-implement them. So, I agree that they could be deleted.

PR #473, opened 2021-08-11
Fit For 55 scenarios and several bug fixes.

Renato-Rodrigues:
* adding Fit for 55 scenario config.
* adding EU27 region support in code and in comparison scenario pdfs.
* adding support for GHG and CO2 emission targets without LULUCF and without Bunkers in module 47. 
* improving ESR target convergence algorithm.
* updating input data to include changes promoted by the Fit for 55 package.
* updating calibration files for ETS/ESR runs. 
* removing unnecessary transportation bound.
* fix bug with ESR 2020 target being considered for convergence check when co2 price is fixed in the scenario config.
* fix bug for missing reference year data in regional emission targets when cm_startyear is higher than 2015.
* fix bug on convergence check for regional targets that caused displaying debugging information about global targets in place of regional targets info.


LaviniaBaumstark (commented):
what is the reason for deleting the bound in transport?

Renato-Rodrigues:
> what is the reason for deleting the bound in transport?

Marianna was having issues with it and after a quick check I couldn't find any reason why the vm_prodUe would need to be fixed to the CES input value.
The model "initialization/calibration" to 2005 values is already capable of assigning a value to apCarPeT that should be compatible with the fe_demand calibration data by definition.

My guess is that this variable fixing (more than two years old) was introduced to help calibrating new formulations of the model and it was simply forgotten there because it was not hurting anything until now.

LaviniaBaumstark (approved):

PR #482, opened 2021-08-23
fossil module repo size mgmt & bugfix

stephenbi:


LaviniaBaumstark (changes requested):
* why did you change scenario_config_SSPSDP?
* Please, run codeCeck and fix the errors to pass all checks.
* What did you change in realization "grades2poly"? 
* Realization "timeDepGrades still contains hard coded region names. Could this be substituted?

stephenbi:
- Oops, I didn't mean to change scenario_config_SSPSDP. These are mostly changes I made to my fork long ago but never pushed since the timeDepGrades update is still in progress. I must have merged that file incorrectly.
- Yes, I'll double check everything.
- We fixed a fairly minor, unrelated bug in grades2poly. A rent discount factor term in the fossil extraction costs was being calculated in terms of cm_startyear when it should have been t0. 
- The lines with hard coded region names are all commented out. Since we weren't able to test the new implementation yet, I think we left those in for reference. I can keep those in my fork, though.

PR #492, opened 2021-09-06
delete use of prices for CES parameter calibration based on input fil…

LaviniaBaumstark:
…es. Use default prices in that case instead.


0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q (changes requested):
minor fixes to flavour text https://github.com/LaviniaBaumstark/remind/pull/1

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q (approved):

PR #504, opened 2021-09-20
Update ariadne scenario config

fschreyer:
Update ariadne scenario config to be in line with the runs of submission from September 17.


MariannaR (approved):
I guess these are the latest changes we already used in the runs, right?

fschreyer:
> 
> 
> I guess these are the latest changes we already used in the runs, right?

Exactly.

PR #505, opened 2021-09-21
Refactor CES markup cost implementation and extent it to industry subsectors 

fschreyer:
Changes:
* refactor CES markup cost implementation to work with different CES structures in buildings and industry
* extent implementation to industry subsectors realization

Background: 
The CES markup cost serves to have some representation of the cost of end-use technologies in buildings and industry (for example cost of heat pumps, electric arc furnaces etc.). The cost are placed on the bottom-level in ``vm_cesIO`` and can be controled via the switches ``cm_CESMkup_ind`` and ``cm_CESMkup_build``. 

By default, there is currently a markup only on heat pump electricity in buildings (``feelhpb``), district heat (``feheb``) and electricity in fixed_shares industry (``feeli``). In the future, we could extent this to ppfen in industry subsectors like ``feelwlth_otherInd`` or others to represent some end-use transformation cost. In specific policy scenarios, these markup cost might also be lowered to represent policy subsidies in certain sectors/for certain technologies. So currently, nothing changes but I might get back to you on this topic about where/how markups make sense.

### DIFF
--- a/config/default.cfg
+++ b/config/default.cfg
@@ -411,12 +411,12 @@ cfg$gms$cm_H2targets <- 0 # def 0, off
 cfg$gms$cm_PriceDurSlope_elh2 <- 20 # def 20
 cfg$gms$cm_FlexTaxFeedback <- 0 # def 0, off
 
-cfg$gms$cm_build_H2costAddH2Inv <- 0.2  # Buildings def 6.5$/kg = 0.2 $/Kwh
+cfg$gms$cm_build_H2costAddH2Inv <- 0.2  # Buildings def 0.2 $/Kwh (6.5$/kg)
 cfg$gms$cm_build_costDecayStart <- 0.05 # Buildings def 5%
 cfg$gms$cm_build_H2costDecayEnd <- 0.1  # Buildings def 10%
 cfg$gms$cm_build_AdjCostActive  <- 0    # def 0 (off)
 
-cfg$gms$cm_indst_H2costAddH2Inv <- 0.1  # Industry def 6.5$/kg = 0.2 $/Kwh
+cfg$gms$cm_indst_H2costAddH2Inv <- 0.1  # Industry def 0.1 $/Kwh (3.2$/kg)
 cfg$gms$cm_indst_costDecayStart <- 0.05 # Industry def 5%
 cfg$gms$cm_indst_H2costDecayEnd <- 0.1  # Industry def 10%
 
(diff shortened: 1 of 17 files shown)


rkrekeler (approved):

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q (changes requested):
```
s37_costAddH2Inv "additional h2 distribution costs for low diffusion levels. [3.25$/kg = 0.1$/kWh]"
```

Square brackets should contain the unit of the scalar.  From this it is not clear whether the number is in \$/kg or \$/kWh.

```
q37_costAddH2PhaseIn(t,regi)..
  v37_costAddTeInvH2(t,regi,"tdh2s")
  =e=
    ( 1 /(
      1 + (3**v37_costExponent(t,regi))
      )
    ) * (
      s37_costAddH2Inv * 8.76
      * ( sum(emiMkt, vm_demFeSector(t,regi,"seh2","feh2s","indst",emiMkt)))
    )
+ (v37_expSlack(t,regi)*1e-8)
```

Reformulate to

```
q37_costAddH2PhaseIn(t,regi)..
  v37_costAddTeInvH2(t,regi,"tdh2s")
  =e=
    (1 / (1 + (3 ** v37_costExponent(t,regi)))) 
  * ( s37_costAddH2Inv 
    * 8.76
    * sum(emiMkt, vm_demFeSector(t,regi,"seh2","feh2s","indst",emiMkt))
    )
  + (v37_expSlack(t,regi) * 1e-8)
;
```

for legibility

- Where does the magic number `8.76` come from?  Couldn't `s37_costAddH2Inv` be in correct units already, since it isn't used anywhere else?
- Same for `1e-8`.  Wouldn't `sm_eps` (1e-9) do, too?

fschreyer:
Ok, I clarified the use ``s37_costAddH2Inv`` and its unit conversion. This ``q37_costAddH2PhaseIn`` implementation was done by @Renato-Rodrigues. He should intervene in case I misunderstood something.

Renato-Rodrigues:
If you care about the unit conversion in the equation you can change the s37_costAddH2Inv unit. Just remember to also change the s36_costAddH2Inv scalar the same way.

1e-8 could also be replaced by sm_eps. If you do that, I would recommend doing a test run after the change just for sanity check.

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q (approved):

PR #508, opened 2021-09-23
Add output parameters for carbon captured in industry to use in reporting and correctly calculate gross emissions targets

fschreyer:
This contains:

* addition of o37_CO2captured as an output parameter in industry to contain the carbon captured in industry by SE, FE, subsector etc.
* correction of calculation of ``o37_demFeIndSub`` output parameter of industry FE per subsector for correct emissions markets mapping in industry subsectors realization
* correction of gross emissions calculation ``v47_emiTarget`` in regipol module for industry CCS from either biomass or synfuel sources


tutorials/7_Advanced_ChangeCode.md
```diff
@@ -67,6 +67,10 @@ The units (e.g., TWa, EJ, GtC, GtCO2, ...) of variables and parameters are docum
 * Don't use $+number combinations, e.g., $20 (this interferes with GAMS error codes).
 * Indicate the end of a file by inserting \*** EOF filename.inc \*** 
 
+#### Sets
```
0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q, line 70: Sure you don't want to put this gem in its own commit?

fschreyer: Ah right. That got squashed into another commit when rebasing. I could single it out again.

fschreyer:
Ok, well I appreciate your code cleaning. So the general rule would be to do line breaks after, I dont know, 50+ characters for legibility in vi?

fschreyer:
And another question: ``o37_CO2Captured(ttot,regi,entySe,entyFe,secInd37,emiMkt)`` is also used in ``regipol/regiCarbonPrice/equations.gms``. Should I rename to start with ``p37_``?

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
> Ok, well I appreciate your code cleaning. So the general rule would be to do line breaks after, I dont know, 50+ characters for legibility in vi?

https://en.wikipedia.org/wiki/Characters_per_line#In_programming

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
> And another question: `o47_CO2Captured(ttot,regi,entySe,entyFe,secInd37,emiMkt)` is also used in `regipol/regiCarbonPrice/equations.gms`. Should I rename to start with `p47_`?

I didn't really look into that module, since I don't know about it.  
I think you should rename it to `pm_CO2Captured` since it shows up in more than one module.  And you should move 
```
o37_CO2Captured(t,regi,entySe,entyFe,secInd37,emiMkt) = 0;
```
to `preloop.gms` instead of `presolve.gms`, else it will always be 0 during solve.
Are you OK with `regipol` lagging behind the rest of REMIND in this regard?  Because it will always hold last iterations value, not the current one.

fschreyer:
Ah great, thx!! That's why ``v47_emiTarget`` didn't fully add up with ``Emi|CO2|Gross|Energy`` in the reporting. 

And I guess due to this

> Are you OK with regipol lagging behind the rest of REMIND in this regard? Because it will always hold last iterations value, not the current one.

there will still be a bit of discrepancy left. That regipol is lagging behind in that regard is actually not ideal, but I think the error is negligible within the range of target convergence that REMIND accepts anyways. I will do another test run with those changes. 

Adding @Renato-Rodrigues who might have an opinion on regipol emissions as well.

Renato-Rodrigues:
What do you mean by lagging behind? I am not sure if I follow. Is this specific to the `grossEnCO2_noBunkers`. If it is, I never used it so far.

Most of the regipol code I added is either located at the preloop, to parameters initialization, or to the postsolve. I wouldn't anticipate any problems with results lagging in this case. 
I am all ears although to understand better what you two mean.

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
> I am all ears although to understand better what you two mean.

`o47_CO2Captured` is calculated from level values of variables in the `industry` module after the model is solved.  Then in the next iteration, its value is used in an equation in the `regipol` module, but there will be a discrepancy between the value used in `regipol` and the _new_ value that will be calculated in `industry` after the model is solved again.  So the modules will not be in equilibrium.
This might be acceptable behaviour, and we use formulations like that when necessary (e.g. Negishi and Nash do the same thing, but also the MAC curves for non-CO<sub>2</sub> emissions and industry CCS), as this usually converges towards equilibrium.  
But it might not converge, introducing oscillations.

Renato-Rodrigues:
Thks Michaja. 
If it is only regarding the `o37_CO2Captured`, I guess this only affects @fschreyer newly added code.

In that case Felix, I would either accept the compromise if the error is small enough, or add its calculation endogenously to the q47_emiTarget_grossEnCO2(t,regi) equation in the contrary case, and if it is not a crazily nonlinear calculation. 

My guess is that this does not change in values substantially between the last iterations, so keeping the error would not be such a big deal. In this case, I would add a "debug" parameter to display the lag error between each iteration (you can calculate it in the postsolve of the regipol), and after the last iteration I would "fix" both the `o37_CO2Captured` and the `v47_emiTarget(t,regi,"grossEnCO2_noBunkers")` values to provide the reporting with the corrected model results without the lag and a consistent sum to the emission target.

fschreyer:
Ok thx you two.  I corrected the gross emissions and added a parameter to save them over iterations to check convergence. They converge fine. And you are right the other type of emissions are not affected because they don't feature parameters.

tutorials/7_Advanced_ChangeCode.md
```diff
@@ -67,6 +67,10 @@ The units (e.g., TWa, EJ, GtC, GtCO2, ...) of variables and parameters are docum
...
+
+* don't use set names with three capital letters (like "ETS" or "ESR"), otherwise the maglcass R library might interpret this as a region name when reading in GDX data
```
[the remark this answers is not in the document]
0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q: Still wrong.

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q (changes requested):

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q (approved):

PR #509, opened 2021-09-28
disabling EU regions if ETS is not active

Renato-Rodrigues:


Renato-Rodrigues:
@cchrisgong could you check if the regipol change, fix the issue you had in your runs with the ETs set?
@Loisel I added conditionally the edge_esm item to the entyFe set. Could you review if you approve the change?

Renato-Rodrigues:
The fegat set change was removed from this pull request as it is superseded by https://github.com/remindmodel/remind/pull/510

Renato-Rodrigues:
I am not sure Chris. 
I will set a test with H12 and if it works fine I will merge it, so you can let me know if it works in your runs also.

cchrisgong (approved):

PR #516, opened 2021-10-01
Removed depreacted file and call to edgeTrpLib function.

MariannaR:
Rebates are not applied 
**note**: this requires an update to the `edgeTrpLib` library!


Loisel (approved):
Looks good, thanks for cleaning up!

PR #517, opened 2021-10-01
update industry/subsectors for dynamic minimum primary steel share

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
- minimum share of primary steel in total steel production converges
  from historic (2015) values to 10 % until 2040
- share and temporal behaviour are controlled by
  - `s37_min_primary_steel_share_target` (0.1)
  - `s37_min_primary_steel_share_from` (2015)
  - `s37_min_primary_steel_share_by` (2040)

### DIFF
--- a/modules/37_industry/subsectors/equations.gms
+++ b/modules/37_industry/subsectors/equations.gms
@@ -32,9 +32,12 @@ q37_energy_limits(ttot,regi,industry_ue_calibration_target_dyn37(out))$(
 
 *** No more than 90% of steel from secondary production
 q37_limit_secondary_steel_share(ttot,regi)$( ttot.val ge cm_startyear ) .. 
-  9 * vm_cesIO(ttot,regi,"ue_steel_primary")
+    vm_cesIO(ttot,regi,"ue_steel_primary")
+  * p37_min_primary_steel_share(ttot,regi)
   =g=
-  vm_cesIO(ttot,regi,"ue_steel_secondary")
+  sum(cesOut2cesIn(out,in)$( cesOut2cesIn2(out,"ue_steel_primary") ),
+    vm_cesIO(ttot,regi,in)
+  )
 ;
 
 *' Compute gross industry emissions before CCS by multiplying sub-sector energy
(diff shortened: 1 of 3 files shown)


0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
Adding @silviamade @mellamoSimon to keep you in the loop.

modules/37_industry/subsectors/equations.gms
```diff
@@ -32,9 +32,12 @@ q37_energy_limits(ttot,regi,industry_ue_calibration_target_dyn37(out))$(
...
 q37_limit_secondary_steel_share(ttot,regi)$( ttot.val ge cm_startyear ) .. 
-  9 * vm_cesIO(ttot,regi,"ue_steel_primary")
+    vm_cesIO(ttot,regi,"ue_steel_primary")
+  * p37_min_primary_steel_share(ttot,regi)
   =g=
-  vm_cesIO(ttot,regi,"ue_steel_secondary")
+  sum(cesOut2cesIn(out,in)$( cesOut2cesIn2(out,"ue_steel_primary") ),
+    vm_cesIO(ttot,regi,in)
+  )
 ;
```
fschreyer, line 41: Sure, this sums up primary and secondary steel? So, it sums all inputs of the output of ``ue_steel_primary``, right? Maybe just the sum of ``ue_steel_primary`` and ``ue_steel_secondary`` would be more intuitive to read.

fschreyer (changes requested):

modules/37_industry/subsectors/equations.gms
0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q: It sums up all inputs to `out`, where `out` is the output that has `ue_steel_primary` as an input.

fschreyer: Ye, I see. If there is no specific reason why to formulate it in that general way, maybe just write ``vm_cesIO("ue_steel_primary") + vm_cesIO("ue_steel_secondary")`` which makes it more intuitive to read and people will not wonder in a couple of years whether that's how it is supposed to be.

fschreyer (approved):

PR #564, opened 2021-11-04
fix of industry emissions factor

fschreyer:
This aligns the emissions factor used for industry carbon capture with the fe emissions factor used in the new REMIND for calculating industry energy emissions. 

Also remove outdated industry emissions reporting from postsolve.


Renato-Rodrigues:
There is really the need for the parameter p37_fctEmi, or could we simply use the pm_emifac parameter instead directly and remove the other one from the code?

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
We can use 'pm_emifac'.  Just needs some set for linking `fety` to `(sety,fety,te)`.

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q (approved):

Renato-Rodrigues:
> We can use 'pm_emifac'. Just needs some set for linking `fety` to `(sety,fety,te)`.

I guess you could define an additional set, or simply use something like this:
`sum((entySe,te)$(se2fe(entySe,entyFe,te) and entySeFos(entySe)), pm_emifac(t,regi,entySe,entyFe,te,"co2")) 
`
to control the Se and te sets in the equation.

PR #570, opened 2021-11-08
turn on wind offshore in default

cchrisgong:


fschreyer:
Hey Chris, sry I forgot to mention: Could you also remove the column of this switch in the rev3p0p0 config file? If not, we will probably forget to switch it on in the ongoing validation.  
Could you switch on "on" by default in main.gms as well so that it does need not overwrite that each time?

giannou:
Please make sure to perform the change also in the `config/default.cfg` file and in all other `config/scenario_config_XXX.csv` files that contain this setting

fschreyer (approved):

Issue #574, opened 2021-11-15
this change breaks several scenarios (SSP1, SSP5, SDP)


giannou:
bf80e45b4207ce4d7d26c71a992d6aa77f579073

Can you take a look?

PR #576, opened 2021-11-15
adding data for wind offshore for SSP1, SSP5, SDP

cchrisgong:
bug fix for issue #574


robertpietzcker (changes requested):
In general it looks ok, but in SSP1 the values from SSP2 were used. 
incolearn for wind and PV should be higher, as it was before the changes (2200 instead of 2000, 5110 instead of 5060).
I would similarly propose to increase incolearn for windoff to 4400 in SSP1, just as you reduced it to 3600 in SSP5

cchrisgong:
> In general it looks ok, but in SSP1 the values from SSP2 were used. incolearn for wind and PV should be higher, as it was before the changes (2200 instead of 2000, 5110 instead of 5060). I would similarly propose to increase incolearn for windoff to 4400 in SSP1, just as you reduced it to 3600 in SSP5

Thanks! This has been fixed

robertpietzcker (approved):

giannou (approved):
looks good, please don't merge until the NDC test runs have also shown the same behavior as before #570

PR #577, opened 2021-11-16
set SSP2EU as default GDP and population scenario

LaviniaBaumstark:


giannou:
Do we need to adjust also the `scenario_config_XXX.csv` files?

LaviniaBaumstark:
I will adjust the scenario_config for the validation. All others need to take this decision project based. Anyway, we need to clean up the number of scenario_configs in the config-folder

giannou:
what about the scenario_config_AMT.csv?

LaviniaBaumstark:
We could change them now, but also later

giannou (approved):

PR #579, opened 2021-11-16
Adding emissions target to regipol module which takes into account land-use change emissions in line with national accounting data

fschreyer:
This adds ``netGHG_LULUCFGrassi_noBunkers`` as an emissions type to be selected for a regipol emissions target. These emissions are shifted by the difference between Magpie land-use change emissions in 2015 and national accounting data due to the Grassi discrepancy. It means that the given emission level are reached assuming that land-use change emissions are corrected for the 2015 discrepancy. REMIND emissions variables are not affected by this. 

Test runs are here: 

``/p/tmp/schreyer/Modeling/REMIND_H12/Current/output/PkBudg900_EUR10Mt_noGrassi_2021-11-12_13.34.15``,

 ``/p/tmp/schreyer/Modeling/REMIND_H12/Current/output/PkBudg900_EUR10Mt_Grassi_2021-11-12_13.34.13`` 
 
 (note that the input data version for these runs still uses PRIMAP data, however an update to UNFCCC was already done in mrremind and should be available in future input data versions). 

An addition in the reporting will follow which reports emissions variables w/ Grassi Shift.


amerfort:
Looks good to me! Let's keep an eye out whether things are still within our expectations with UNFCCC input data.

amerfort (approved):
Looks good to me! Let's keep an eye out whether things are still within our expectations with UNFCCC input data.

giannou:
The current REMIND trunk crashes for all scenarios, with the following error:
```
ExoGAINS - log for first iteration...

Error in p_macBase[, y, ] : incorrect number of dimensions
Calls: mbind -> reportPE
In addition: Warning message:
In readGDX(gdx, c("p_macBaseMagpie", "p_macBase"), format = "first_found") :
  No corresponding object found in the GDX. NULL is returned!
Execution halted
```
Looks like the error is related to this commit, can you take a look @fschreyer?

fschreyer:
Sry, thanks for pointing out. I actually did a test run but killed it before the first iteration completed.... :(

I fixed the reporting to work again with the latest trunk with [this PR](https://github.com/pik-piam/remind2/pull/128). See a test run here ``/p/tmp/schreyer/Modeling/REMIND_H12/Current/output/default_2021-11-18_11.03.55``.

PR #600, opened 2021-12-07
Regionalize early retirement rate cap

cchrisgong:
In this pull request, we make annual early retirement allowances dependent on region, time, technology, and scenario. This improves upon the early retirement feature designed (ad-hoc) to prevent sudden tipping (e.g. within a single model period) of unprofitable technology once the run starts. In a sense this imitates real-world inertia against sudden infrastructure change to some degree. 

We have deprecated the old "cm_earlyreti_rate" switch, which limited the maximum rate of early retirement of economically uncompetitive capacities to 9% per annum with a handful of exceptions that were scattered throughout input data routines and the REMIND code. 

We replace this switch with two new ones:

1. "c_regi_earlyreti_rate" = sets the general early retirement limit in each region for all technologies and periods. For example, the current default values for policy runs are: "GLO 0.09, EUR_regi 0.15". (Note: "_regi" is needed to make the region definition recognizable by GAMS under all possible mapping). This preserves the old H12 default rate (cm_earlyreti_rate = 0.09) for all regions except EUR, where 15% can be retired per year to match the default behavior in REMIND-EU runs.

2. "c_tech_earlyreti_rate" = allows definition of technology-specific early retirement rates in each region, which only apply through 2030. From 2035 onward, all technologies (except in EUR_regi) revert to the global technology-specific rate.
Current default values (for H12):
"GLO.(biodiesel 0.14, bioeths 0.14), EUR_regi.(biodiesel 0.15, bioeths 0.15), USA_regi.pc 0.13, REF_regi.pc 0.13, CHA_regi.pc 0.13"
Above indicates a fast phase-out of conventional biomass, a high retirement rate of coal plants in US and REF, CHA.

These switches are fed into a new parameter, pm_regiEarlyRetiRate, which enters q_smoothphaseoutCapEarlyReti.

Additionally, we have introduced special default values for certain scenarios: 
1. Basline: c_tech_earlyreti_rate = "off", but from 2030 onwards all techs in all regions are set to 0.01 to prevent rapid early phase-outs of coal and gas plants that make the baseline look unnatural/over-optimistic. 
2.  c_regi_earlyreti_rate in SDP scenarios: "GLO 0.12, EUR_regi 0.15"

The time dependencies are currently hard-coded in core/datainput.gms and thus not user-configurable through scenario_config. Current default value for REMIND-EU is the same as above, except also adding DEU.tnrs 0.2 (Germany's nuclear phase-out). Note that EU regions do not have the "_regi" extension

A summary of the current values and questions for group discussion is here: 
image

Note: only scenario_config_AMT.csv, scenario_config_rev3p0p0.csv and scenario_config_21_EU11_ARIADNE.csv have added columns of the new parameters, other scenario configs don't have this and will take the above mentioned default values. We recommend that these switches be added to all scenario_configs, especially those from which Baselines are run.


config/21_regions_EU11/scenario_config_21_EU11_ARIADNE.csv
```diff
@@ -1,8 +1,8 @@
...
 _____Calibration_____;0;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
-Calib_EU;0;calibrate;0;14;off;nash;DEU;parallel;off;GLO 0.09, EUR_regi 0.15;GLO.pc 0.05, EUR_regi.pc 0.15;se_trade;off;1;-1;0;0;0;0;EUR_regi, NEU_regi;on;1;1;0;20;0.2;standard;standard;off;off;off;off;off;1;./config/21_regions_EU11/regionmapping_21_EU11.csv;REG;none;0;0;none;-1;2100;3;2050;regiCarbonPrice;off;off;off;off;on;on;debt_limit;subsectors;simple;edge_esm;Mix;1.05;1;none;2;on;pop_SSP2EU;gdp_SSP2EU;gdp_SSP2EU;2005;;
+Calib_EU;0;calibrate;0;14;off;nash;DEU;parallel;off;GLO 0.09, EUR_regi 0.15;;EUR_regi.pc 0.15;se_trade;off;1;-1;0;0;0;0;EUR_regi, NEU_regi;on;1;1;0;20;0.2;standard;standard;off;off;off;off;off;1;./config/21_regions_EU11/regionmapping_21_EU11.csv;REG;none;0;0;none;-1;2100;3;2050;regiCarbonPrice;off;off;off;off;on;on;debt_limit;subsectors;simple;edge_esm;Mix;1.05;1;none;2;on;pop_SSP2EU;gdp_SSP2EU;gdp_SSP2EU;2005;;
```
Renato-Rodrigues, line 3: You don't need a tech specific value (EUR_regi.pc 0.15) in the calib and base scenarios after your changes, as all techs will use 0.15 anyway in this case, no?

stephenbi:
True, good catch. That's a remnant of our previous iteration which also set
GLO.pc to 0.05.

Maybe it's good to leave it to be safe, in case we later add other tech
specific GLO values? I don't have a strong opinion on it though.

config/21_regions_EU11/scenario_config_21_EU11_ARIADNE.csv
stephenbi: Ah, but the double semicolon there is a bug.

Renato-Rodrigues: I would definitely prefer to remove it if it is unnecessary

stephenbi: OK, do you have a convention for a null value for these types of switches? Will it accept a simple string, e.g. "none"?

Renato-Rodrigues: I usually use "off" as a string to disable using the switch. You can check the regipol module to see examples of that.

stephenbi (approved):

config/21_regions_EU11/scenario_config_21_EU11_ARIADNE.csv
cchrisgong: Thanks, @Renato-Rodrigues! This has been fixed

PR #602, opened 2021-12-09
Rescale FE input data for 2005 capacity calibration and enable historic biomass share constraint for buildings and industry again

fschreyer:
This PR addresses the [historic biomass vs. fossil split issue](https://github.com/remindmodel/development_issues/issues/37) and provides a quick fix for the input data inconsistency between the 2005 calibration and the CES calibration FE trajectories. 
 
* rescaling of ``f04_IO_output`` 2005 FE data used for the 2005 se2fe capacity calibration of REMIND to be in line with the FE data provided by ``pm_fedemand`` for the CES calibration, this is a quick fix; the long-term solution should be to make both input data in line in mrremind. It mainly affects solids/liquids/gases in buildings and industry. 

* activate the ``q_shbiofe_lo`` and ``q_shbiofe_up`` constraints again under industry subsectors to constrain biomass shares to historic levels

You find a comparison PDF here: ``/p/tmp/schreyer/Modeling/REMIND_H12/NewDev/compScen-2021-12-08_17.25.03-H12`` (default_2 is the "fixed" version).

### DIFF
--- a/core/datainput.gms
+++ b/core/datainput.gms
@@ -1512,7 +1512,7 @@ sm_globalBudget_dev = 1;
 s_histBioShareTolerance = 0.02;
 *** temporary: until subsectors historical FE mix checked -> high tolerance for industry subsectors to make it run 
 $ifthen.subsectors "%industry%" == "subsectors"   !! industry
-s_histBioShareTolerance = 0.3;
+s_histBioShareTolerance = 0.02;
 $endif.subsectors
 
 
--- a/core/datainput.gms
+++ b/core/datainput.gms
@@ -1507,14 +1507,4 @@ $endif.subsectors
 *** initialize global target deviation scalar
 sm_globalBudget_dev = 1;
 
-*** define tolerance level by how much biomass share needs to comply with 2005 historical values
-*** low tolerance for fixed_shares, there is works
-s_histBioShareTolerance = 0.02;
-*** temporary: until subsectors historical FE mix checked -> high tolerance for industry subsectors to make it run 
-$ifthen.subsectors "%industry%" == "subsectors"   !! industry
-s_histBioShareTolerance = 0.02;
-$endif.subsectors
-
-
-
 *** EOF ./core/datainput.gms
--- a/modules/04_PE_FE_parameters/iea2014/datainput.gms
+++ b/modules/04_PE_FE_parameters/iea2014/datainput.gms
@@ -24,11 +24,110 @@ $offdelim
 /
 ;
 
+
+
+
 *** making sure f04_IO_output is compatible with pm_fedemand values in 2005  
 *** this will become irrelevant to the model once the input data routines can be fixed so that pm_fedemand is again the same as f04_IO_output
 *** these lines should be removed once this is fixed at mrremind side.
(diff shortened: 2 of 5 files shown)


core/datainput.gms
```diff
@@ -1507,14 +1507,4 @@ $endif.subsectors
 *** initialize global target deviation scalar
 sm_globalBudget_dev = 1;
 
-*** define tolerance level by how much biomass share needs to comply with 2005 historical values
-*** low tolerance for fixed_shares, there is works
```
Renato-Rodrigues, line 1511: "it" instead of "is"?

core/datainput.gms
```diff
@@ -1507,14 +1507,4 @@ $endif.subsectors
...
-s_histBioShareTolerance = 0.02;
-*** temporary: until subsectors historical FE mix checked -> high tolerance for industry subsectors to make it run 
-$ifthen.subsectors "%industry%" == "subsectors"   !! industry
-s_histBioShareTolerance = 0.3;
```
Renato-Rodrigues, line 1515: 30% of tolerance is quite crazy. I understand why you are doing this, but I am afraid that this is an easy way out, and it will give the wrong impression that things are working meanwhile the issue is not really fixed in the industry subsectors side.

Renato-Rodrigues (commented):
I don't see any problem with the formulation at a first look, however I am not sure if I agree of pushing this.
I implemented the fix in datainput in buildings, just because the buildings numbers at the time had very small deviations (less than 0.001). It was always intended to be a very temporary fix.
Generalizing this can be counterproductive.
I am afraid that this is an easy way out, and it will give the wrong impression that things are working meanwhile the issue is not really fixed in the industry subsectors side.
It will just hide the problem for now, and not push for a solution.

Renato-Rodrigues (approved):

modules/04_PE_FE_parameters/iea2014/datainput.gms
```diff
@@ -24,11 +24,109 @@ $offdelim
...
+                                                               + pm_fedemand("2005",regi,"feli_steel")
+                                                               + pm_fedemand("2005",regi,"feli_chemicals"))
+                                                            /  p04_IO_output_beforeFix_Total("2005",regi,"fehoi");
+
+*** industry gases
+p04_IO_output_beforeFix_Total(t,regi,"fegai") = p04_IO_output_beforeFix(t,regi,"segabio","fegai","tdbiogai")
+                                                  + p04_IO_output_beforeFix(t,regi,"segafos","fegai","tdfosgai");
+
+
+f04_IO_output("2005",regi,"segabio","fegai","tdbiogai")$(p04_IO_output_beforeFix_Total("2005",regi,"fegai")) = p04_IO_output_beforeFix("2005",regi,"segabio","fegai","tdbiogai")
```
LaviniaBaumstark, line 101: I do not like that this is a change of pure input data (f*). Could we use a parameter for this?

LaviniaBaumstark (commented):
I think, we should push this for now but agree with Renato that we really need to fix it in in mrremind soon.

modules/04_PE_FE_parameters/iea2014/datainput.gms
fschreyer: The original data are still in ``p04_IO_output_beforeFix``. Can be aligned once temporary solution not needed anymore.

PR #612, opened 2021-12-21
Update input data to rev 6.275

rkrekeler:
I was irritated that there are only 2 instead of the usual 4 files (regionmappings). Is it okay to only have `2b1450bc` and `62eff8f7`? Otherwise we should not merge this.


LaviniaBaumstark (approved):

LaviniaBaumstark:
yes, this are the currently needed two regionmappings. We maybe should delete the other mappings from our input data generation script

PR #620, opened 2022-01-05
Moving cm_regiExoPrice to the regipol module

Renato-Rodrigues:
Allow setting exogenous carbon price trajectories only for specific regions


robertpietzcker (approved):
I would like 2 changes: 
1. I think for all these "new" parameter fillings via config, additional information on the parameter filling in the parameter definition is needed. so instead of

p47_exoCo2tax(ext_regi,ttot) "Exogenous CO2 tax level [$/tCO2]" / %cm_regiExoPrice% /

explicitly write

p47_exoCo2tax(ext_regi,ttot) "Exogenous CO2 tax level [$/tCO2]. The input has to have the form <region {RP: are both regions and region groups allowed?}>.year price, year price {?no idea what has to be entered between two different regions?} <region group>.year price" / %cm_regiExoPrice% /
*' This parameter overrides {RP: or is added to?!?} the carbon prices in pm_taxCO2eq, which are otherwise resulting from the normal climate target settings for the regions for which it is given  for all time steps? does it only override the full-system prices or also the ETS/ESR prices?}. If not all time steps are explicitly given, the model inserts 0 for those time steps. 

2. also, it is important to add a display of pm_taxCO2eq if this is the last time the parameter  is changed before the next iteration - otherwise it will impossible to understand from the full.lst what the tax level was when the iteration started.

Renato-Rodrigues:
> I would like 2 changes:
> 
> 1. I think for all these "new" parameter fillings via config, additional information on the parameter filling in the parameter definition is needed. so instead of
> 
> p47_exoCo2tax(ext_regi,ttot) "Exogenous CO2 tax level [$/tCO2]" / %cm_regiExoPrice% /
> 
> explicitly write
> 
> p47_exoCo2tax(ext_regi,ttot) "Exogenous CO2 tax level [$/tCO2]. The input has to have the form <region {RP: are both regions and region groups allowed?}>.year price, year price {?no idea what has to be entered between two different regions?} .year price" / %cm_regiExoPrice% / *' This parameter overrides {RP: or is added to?!?} the carbon prices in pm_taxCO2eq, which are otherwise resulting from the normal climate target settings for the regions for which it is given for all time steps? does it only override the full-system prices or also the ETS/ESR prices?}. If not all time steps are explicitly given, the model inserts 0 for those time steps.
> 

GAMS has as a general limitation to not allow anything in the code superseed 255 characters. I included a more concise description to the parameter that hopefully address the points you raised above.
https://github.com/remindmodel/remind/pull/620/commits/67475f08311e8fc144d2ca65cdc9bb6d02f5d029

> 2. also, it is important to add a display of pm_taxCO2eq if this is the last time the parameter  is changed before the next iteration - otherwise it will impossible to understand from the full.lst what the tax level was when the iteration started.

This is not the right place for that in my opinion.
I already included to the solution report the display of the most relevant variables and parameters for convergence (check the optimization, nash realisation, postsolve code). pm_taxCO2eq  is most probably already there.
If it is not there, I would strongly recommend to include ir or any additional displays to the optimization module instead of other parts of the code. This is the only to be certain that you are actually displaying something that is not being modified by some other module realisation in the way.

robertpietzcker:
> GAMS has as a general limitation to not allow anything in the code superseed 255 characters. I included a more concise description to the parameter that hopefully address the points you raised above.

I would add the additional information that doesn't fit into the 255 characters in a comment below the definition. The current description is better, but I think it would help if you add a) how to do several regions, b) say explicitly that one has to put in data for ALL timesteps, otherwise you might get really weird effects (at least that is my interpretation - if you eg only overwrite the 2030 time step, then the next update of CO2 prices in some iterative-target-adjust might take that 2030 value as the basis for the change for all the later time steps, and as they are not overwritten you could get crazy results)

robertpietzcker:
> > 3\. also, it is important to add a display of pm_taxCO2eq if this is the last time the parameter  is changed before the next iteration - otherwise it will impossible to understand from the full.lst what the tax level was when the iteration started.
> 
> This is not the right place for that in my opinion.



> > 3\. also, it is important to add a display of pm_taxCO2eq if this is the last time the parameter  is changed before the next iteration - otherwise it will impossible to understand from the full.lst what the tax level was when the iteration started.
> 
> This is not the right place for that in my opinion. I already included to the solution report the display of the most relevant variables and parameters for convergence (check the optimization, nash realisation, postsolve code). pm_taxCO2eq is most probably already there. If it is not there, I would strongly recommend to include ir or any additional displays to the optimization module instead of other parts of the code. This is the only to be certain that you are actually displaying something that is not being modified by some other module realisation in the way.

Your reply made me think about it a bit.
I see two aspects: 
(1) in my view, it still makes sense to have a display WHENEVER such key solution-relevant iterative-adjusted parameters as the CO2 price are changed in between iterations. Otherwise it can be really painful / difficult to understand why the CO2 price changes in a certain way (or does not change in a certain expected way) between iterations.
so I would ask you to add a display here (maybe with an explicit:"display 'update of CO2 prices due to exogenously given CO2 prices in p47_exoCo2tax';") inside the code part that is only included when exogenous co2 taxes are given.

(2) I agree that one cannot be sure that the resulting values are actually the ones the code sees in the solve, so adding that in a central place might make sense. However, while I think postsolve in module 80 is theoretically the right point, practically I would say it is not  a good place to do this, given the (problematic) approach that iterative updates mostly happen in POSTsolve and not in PREsolve (as I now think they should)
this means that any updates of co2 prices (and other things) in all the different modules will take place AFTER the solve statement and BEFORE postsolve of module 80 is called, so a display there will not show what the model saw during solve.

Renato-Rodrigues:
>  a) how to do several regions

I think we are overcomplicating this. The possible ways to fill these parameters are exactly the possible ways documented in the GAMS docs on how to define parameter values. This is already extensively documented in the GAMS help files, and I see no point of repeating the GAMS documentation for parameters definition in the REMIND code. Anyone can simply read the parameters section of the GAMS documentation to know how to define values of parameters with multiple dimensions. 

https://www.gams.com/34/docs/UG_DataEntry.html#UG_DataEntry_ParametersAnIllustrativeExamples
https://www.gams.com/34/docs/UG_DataEntry.html#UG_DataEntry_ParameterDataForHigherDimensions

If you think the GAMS docs are not clear enough, I can do a seminar explaining these alternatives to define parameter values to the REMIND group. 

>  b) say explicitly that one has to put in data for ALL timesteps, otherwise you might get really weird effects (at least that is my interpretation - if you eg only overwrite the 2030 time step, then the next update of CO2 prices in some iterative-target-adjust might take that 2030 value as the basis for the change for all the later time steps, and as they are not overwritten you could get crazy results)

You don't need to explicitly define all time steps. You could use this switch for example to only fix historical year values. As the regipol module is the last module that deals with carbon price determination, these are the values that will be used in your run. If you define exogenous prices just for specific years, it would be natural that you expect the model to use those as reference values for any iteration, so I don't see any inconsistency about that. 
Exogenous prices are not the default behavior, so anybody using those will do it because he or she wants to force them, either for all time steps or for only a specific one.

>  (1) in my view, it still makes sense to have a display WHENEVER such key solution-relevant iterative-adjusted parameters as the CO2 price are changed in between iterations. Otherwise it can be really painful / difficult to understand why the CO2 price changes in a certain way (or does not change in a certain expected way) between iterations.
so I would ask you to add a display here (maybe with an explicit:"display 'update of CO2 prices due to exogenously given CO2 prices in p47_exoCo2tax';") inside the code part that is only included when exogenous co2 taxes are given.

I am commiting the requested additional display.

>  (2) I agree that one cannot be sure that the resulting values are actually the ones the code sees in the solve, so adding that in a central place might make sense. However, while I think postsolve in module 80 is theoretically the right point, practically I would say it is not a good place to do this, given the (problematic) approach that iterative updates mostly happen in POSTsolve and not in PREsolve (as I now think they should)
this means that any updates of co2 prices (and other things) in all the different modules will take place AFTER the solve statement and BEFORE postsolve of module 80 is called, so a display there will not show what the model saw during solve.

Calculating the iterative target adjustments in the post solve is not an issue from my point of view, as they are exclusively dependent of the previous run results. However, it is true that the code in the presolve should never change any iterative target adjustments so the next iteration sees the calculated and displayed results as they were calculated in the postsolve.
Checking this would require some dedicated time that I currently do not have at hand.

Issue #315, opened 2021-02-16
add job type selection for scenario comparisons and other improvements


Renato-Rodrigues:
After you select:
Rscript output.R > 2: Comparison across runs > 9: compareScenarios > and choose the scenario to be compared

It would be important to have the two following questions asking for additional user input:

(1) Please provide a file name prefix (press enter to none):

If the user type any string, this should be added to the beginning of the comparison scenario filenames. This is highly important when creating multiple comparison scenarios simultaneously.

(2) Please choose the SLURM configuration for your submission (default: priority):

options: 
1: SLURM priority
2: SLURM standby 
 
This is highly important because the cluster is taking too long in the standby queue. If you are creating reportings by hand, there is a higher chance that this is one of your priorities.

amnmalik:
I have a few suggestions:
1. Currently, _short_ until 2050 and _long/normal_ until 2100 are automatically produced, could we have other arguments that are part of the function remind2:compareScenarios(), allowing to have specific years, regions etc.?
2. Some comments/description on the various options after executing `Rscript output.R` would be very useful.

Renato-Rodrigues:
About 1, this is already support in the parameters: 

- y = c(seq(2005, 2060, 5), seq(2070, 2100, 10))
- y_hist = c(seq(1960, 2020, 1), seq(2025, 2100, 5))
- y_bar = c(2010, 2030, 2050, 2100)
- reg = NULL
- mainReg = "GLO"

amnmalik:
True, but they are not available in the routine you mentioned : Rscript output.R > 2: Comparison across runs > 9: compareScenarios > and choose the scenario to be compared, or where can they be entered?

Renato-Rodrigues:
yeah, but the Rscript output.R is just including the most common uses for the reporting. 
If you want a specific configuration you should run the function instead.

I am not saying that more flexibility couldn't be added to it if there is enough use cases and user requests to justify them, but I would put this request in a different group from the ones I mentioned in the first message.

giannou:
You can change the configuration of the R function in `scripts/output/comparison/compareScenarios.R`, no?

PR #621, opened 2022-01-05
add filename_prefix and slurmConfig to output.R with compareScenario

orichters:
this adds the possibility to specify a filename_prefix and the slurmConfig (priority vs. standby) for the compareScenario run of output.R.

If compareScenario is selected, the script asks to enter a filename_prefix and checks whether it contains only letters, numbers, `-` or `_` to be safe. In any case, the filename of the final pdf then reads something like `compScen-whatever-2022-01-05_17.36.03-H12.pdf`. If left empty, everything should remain as it is. As I know that these scripts are called from many other scripts, my extension carefully checks whether `filename_prefix` exists to be sure not to break anything.

If compareScenario is selected, the scripts asks you to run it under slurmCondig priority or standby.

Alternatively, these information can be specified by running `Rscript output.R filename_prefix=whatever slurmConfig=priority`. If you require more specific slurm configs that contains blanks, write it as `slurmConfig="--qos=priority --further=data"`.

If further scripts in `scripts/output/comparison` want to use this, simply add it their names to `modules_using_filename_prefix` resp. `modules_using_slurmConfig` and use it. This is done to avoid annoying people that use other output scripts, as they aren't asked now to specify something that is not used later.

### DIFF
--- a/output.R
+++ b/output.R
@@ -138,7 +138,7 @@ choose_mode <- function(title = "Please choose the output mode") {
 }
 
 choose_slurmConfig_priority_standby <- function(title = "Please enter the slurm mode, uses priority if empty") {
-  slurm_options <- c("--qos=priority", "--qos=short", "--qos=standby")
+  slurm_options <- c("priority", "short", "standby")
   cat("\n\n", title, ":\n\n")
   cat(paste(seq_along(slurm_options), slurm_options, sep = ": "), sep = "\n")
   cat("\nNumber: ")
@@ -213,8 +213,10 @@ if (comp == TRUE) {
   if (!exists("slurmConfig") && any(modules_using_slurmConfig %in% output)) {
     slurmConfig <- choose_slurmConfig_priority_standby()
   }
-  if (slurmConfig %in% c("priority", "short", "standby")) {
-    slurmConfig <- paste0("--qos=", slurmConfig)
+  if (exists("slurmConfig")) {
+    if (slurmConfig %in% c("priority", "short", "standby")) {
+      slurmConfig <- paste0("--qos=", slurmConfig)
+    }
   }
 
   # Set value source_include so that loaded scripts know, that they are
--- a/output.R
+++ b/output.R
@@ -29,7 +29,7 @@ library(gms)
 if (!exists("source_include")) {
   # if this script is not being sourced by another script but called from the command line via Rscript read the command
   # line arguments and let the user choose the slurm options
-  readArgs("outputdir", "output", "comp", "remind_dir")
+  readArgs("outputdir", "output", "comp", "remind_dir", "filename_identifier")
 }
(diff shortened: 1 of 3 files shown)


Renato-Rodrigues (approved):
From a first look I don't see anything that I would change. If you tested it and it is working fine I would say go ahead and merge it.

dklein-pik:
If you find it helpful, you could also provide a sample filename when the user is prompted for the filename, so the users knows how their identifiers will show up in the final file name. Like you did in the explanation of this PR above "the filename of the final pdf then reads something like compScen-whatever-2022-01-05_17.36.03-H12.pdf". Only if you think its helpful.

chroetz (approved):
Looks good

orichters:
@Renato-Rodrigues: I also added slurmConfig choice. This issue should now be fixed: https://github.com/remindmodel/remind/issues/315

@dklein-pik: great idea, I added this explanation.

I also added a bit of explanation to the tutorial.

output.R
```diff
@@ -173,6 +198,30 @@ if (comp == TRUE) {
...
+  modules_using_filename_prefix <- c("compareScenarios")
+  if (!exists("filename_prefix")) {
+    if (any(modules_using_filename_prefix %in% output)) {
+      filename_prefix <- choose_filename_prefix(modules = modules_using_filename_prefix[modules_using_filename_prefix %in% output])
+    } else {
+      filename_prefix <- ""
+    }
+  } else {
+    # because readArgs cannot treat empty strings, `no prefix` is specified by `-`
+    if (filename_prefix == "-") {
```
LaviniaBaumstark, line 211: does it mean, the users always need to type at least "-"?

orichters: I added this because you could not use the command line option `Rstart output.R filename_prefix=` to get the old result without prefix. But [my now-merged PR](https://github.com/pik-piam/lucode2/pull/62) in lucode2 fixed this, so we can delete it.

output.R
```diff
@@ -132,11 +132,36 @@ choose_mode <- function(title = "Please choose the output mode") {
...
     comp <- TRUE
   } else {
-    stop("This mode is invalid. Please choose a valid mode")
+    stop("This mode is invalid. Please choose a valid mode.")
   }
   return(comp)
 }
 
+choose_slurmConfig_priority_standby <- function(title = "Please enter the slurm mode, uses priority if empty") {
+  slurm_options <- c("--qos=priority", "--qos=standby")
```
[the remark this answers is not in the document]
orichters: I did.

output.R
LaviniaBaumstark: perfect

LaviniaBaumstark (approved):

LaviniaBaumstark (approved):

PR #625, opened 2022-01-11
Bugfix: tolerate missing "path_gdx_bau" column in scenario_config.csv…

dklein-pik:
… files


orichters (approved):
I wrote and tested the part in start.R and it should work fine. I have not much experience with prepare_and_run.R, but I had a look at it earlier and it looks convincing.

Renato-Rodrigues (approved):

PR #633, opened 2022-01-18
add compareScenarios2 to output.R

chroetz:
compareScenarios2 can now be called via output.R. The necessary additional scripts are copies of the respective scripts for the old compareScenarios with tiny changes to adapt to the new compareScenarios2


scripts/output/comparison/compareScenarios2.R
```diff
@@ -0,0 +1,140 @@
...
+
+start_comp <- function(outputdirs,
+                       shortTerm,
+                       outfilename,
+                       regionList,
+                       mainReg) {
+  if (!exists("slurmConfig")) {
+    slurmConfig <- "--qos=standby"
+  }
+  jobname <- paste0(
```
orichters, line 77: Is there a good reason not to call it `compScen` as well? I mean, as you said, for the user, nothing needs to change… If you stick to `cs2`, please add it in the [.gitignore](https://github.com/remindmodel/remind/blob/develop/.gitignore#L27) in the same way.

scripts/output/comparison/compareScenarios2.R
```diff
@@ -0,0 +1,140 @@
...
+      "cs2",
+      ifelse(outfilename == "", "", "-"),
+      outfilename,
+      ifelse(shortTerm, "-shortTerm", "")
+    )
+  cat("Starting ", jobname, "\n")
+  on_cluster <- file.exists("/p/projects/")
+  clcom <- paste0(
+    "sbatch --qos=standby",
```
orichters, line 86: should be `paste0("sbatch", slurmConfig, …`

scripts/output/comparison/compareScenarios2.R
```diff
@@ -0,0 +1,140 @@
...
+start_comp <- function(outputdirs,
+                       shortTerm,
+                       outfilename,
+                       regionList,
+                       mainReg) {
+  if (!exists("slurmConfig")) {
+    slurmConfig <- "--qos=standby"
+  }
+  jobname <- paste0(
+      "cs2",
```
orichters, line 78: Is there a good reason not to call it `compScen` as well? I mean, as you said, for the user, nothing needs to change… If you stick to `cs2`, please add it in the [.gitignore](https://github.com/remindmodel/remind/blob/develop/.gitignore#L27) in the same way.

scripts/output/comparison/compareScenarios2.R
chroetz: Thanks for spotting that! Should this also be changed in compareScenarios1: https://github.com/remindmodel/remind/blob/bc8acd242c9958d779ef1b0e88520d00384d6f6b/scripts/output/comparison/compareScenarios.R#L56

scripts/output/comparison/compareScenarios2.R
chroetz: Well, it's not necessarily a "good" reason, but I usually use shorthand cs2 for compareSceanrios2 and it might be better to have a different name for the document to indicate that it is not coming from the old compareScenarios. But, yes, if `cs2` is used it has to go into gitignore

scripts/output/comparison/compareScenarios2.R
orichters: yes, please! I like the way you formulated it, avoiding redundancies and these errors in the future!

scripts/output/comparison/compareScenarios2.R
orichters: I am certainly not the one who should decide that :)

PR #636, opened 2022-01-19
fix regex in output.R

chroetz:
The first argument of [`sub()`](https://github.com/remindmodel/remind/blob/49f1db07fd3a2b3167be7d229975212f3f5be4fe/output.R#L63) is a regex-pattern, in which the `.` is a wildcard. If I understand the intention of the modified line of code correctly, only the symbol `.` should be matched and the matches should occur at the beginning and end of the strings, respectively. The regex-patterns are modified to achieve this.

### DIFF
--- a/output.R
+++ b/output.R
@@ -60,7 +60,7 @@ choose_folder <- function(folder, title = "Please choose a folder") {
   # Detect all output folders containing fulldata.gdx
   # For coupled runs please use the outcommented text block below
 
-  dirs <- sub("/fulldata\\.gdx$", "", sub("^\\./output/", "", Sys.glob(file.path(folder, "*", "fulldata.gdx"))))
+  dirs <- basename(dirname(Sys.glob(file.path(folder, "*", "fulldata.gdx"))))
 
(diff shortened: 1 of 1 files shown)


output.R
```diff
@@ -60,7 +60,7 @@ choose_folder <- function(folder, title = "Please choose a folder") {
   # Detect all output folders containing fulldata.gdx
   # For coupled runs please use the outcommented text block below
 
-  dirs <- sub("/fulldata.gdx", "", sub("./output/", "", Sys.glob(file.path(folder, "*", "fulldata.gdx"))))
+  dirs <- sub("/fulldata\\.gdx$", "", sub("^\\./output/", "", Sys.glob(file.path(folder, "*", "fulldata.gdx"))))
```
pfuehrlich-pik, line 63: This looks correct, although I find regex in general hard to read, and regex tends to produce fragile code. What about this:
```suggestion
  dirs <- basename(dirname(Sys.glob(file.path(folder, "*", "fulldata.gdx"))))
```

chroetz: Yes, I like that much better

pfuehrlich-pik (approved):

PR #640, opened 2022-01-25
Harmonized emission budgets

katarkow:
Update the way we calculate our emission budgets in core/postsolve.gms.

Currently we are calculating our emission budgets from 2010 or 2011 on, which makes things a bit hard to compare to e.g. AR6 budgets that runs from 2020.

Budgets from AR6 for 2020-2100 (including 2020), for 1.5°C: 400 Gt CO2, for 2°C: 1150 Gt CO2 (in the code they're 200Gt higher).

I updated computations for s_actualbudgetCO2 and p_actualbudgetCO2. These are used together with c_budgetCO2 and c_budgetCO2FFI, so since the time frame changed, I also updated default values of these parameters and the corresponding values in the scenario config files (by subtracting 200Gt).


core/postsolve.gms
```diff
@@ -44,47 +44,16 @@ loop(ttot$(ttot.val ge 2005),
   );
 );
 
-
-if(cm_iterative_target_adj eq 1,
```
LaviniaBaumstark, line 48: is there no cm_iterative_target_adj eq 1 any more?

LaviniaBaumstark (commented):
please also adjust the scenario names when you change the budget, could be very misleading otherwise

core/postsolve.gms
LaviniaBaumstark: o.k., I trust Jessi

katarkow:
> please also adjust the scenario names when you change the budget, could be very misleading otherwise

Thanks, I haven't noticed that they're related! 

I also noticed that scripts/output/comparison, scripts/utils/compareParallel.R, and config/multi_comparison_matrix.csv use PkBudg<sth> which I guess correspond to these scenario names that I have updated in the last commit. Do you think I should also update these?

LaviniaBaumstark:
I think, those scripts either read in the scenarios interactively or anyway need some adjustments. So, not part of your PR

LaviniaBaumstark (approved):

PR #642, opened 2022-01-25
Disabling 2020 NPi spv lower bound due to incompatibility with core bounds

Renato-Rodrigues:
* this change is not strictly necessary as explained by Christopher in the REMIND mattermost channel because NPi should always have cm_startyear higher than 2020. However, in order to make this information clear and keep consistency when anybody searching the code for bounds that affect an specific technology, I recommend including the following time conditional `$(t.val gt 2020)` for all bounds set in the NPi module realisation.
* I only included the conditional for spv, as this creates a direct incompatibility with the current core bounds definitions.


Renato-Rodrigues:
> Yes, this is ok with the comment. I wonder if it makes sense to include a comment also into the bounds.gms file, up to you...

comment added to the code: https://github.com/remindmodel/remind/pull/642/commits/e89ff5d46c97c61cfed541b94db5d006cf66a5dd

PR #647, opened 2022-02-02
add module 48_carbonprice

orichters:
For the NGFS project, this PR introduces the module `48_carbonpriceRegi`. It adds a regional markup to the carbon prices that can be used to combine a `45_carbonprice` module to additionally achieve NDC or netZero targets.

1. `core`
- adding `pm_taxCO2eqRegi` as "additional regional CO2 tax path" which is usually set to zero. Its value is added to `pm_pvpRegi(ttot,regi,"perm")` which still covers the full GHG taxation

2. `21_tax/on`
- add `pm_taxCO2eqRegi` to all relevant taxation variables

3. `48_carbonpriceRegi/netZero`
- adaption of a realization implemented by Jerome Hilaire.
- in `datainput.gms`, allows to specify netZero years for regions and brings them to zero in this year by adapting `pm_taxCO2eqRegi`. Before and after this year, uses a triangular shape for tax rates, going to zero in 2100. Can easily be adapted "on the fly" within remind.

4. `48_carbonpriceRegi/NDC`
- uses the `UNFCCC_NDC` database to import both NDC targets and netZero targets from mrremind.
- chooses the years where a target with sufficiently coverage within the region is given. Adapts `pm_taxCO2eqRegi` in all these years such that all targets are met. Uses a linear interpolation between these years.

If any 45 realization is selected, both 48 realizations start only after the 10th iteration, hoping that the global carbon price has converged sufficiently.

Runs with these new realizations can be found in `/p/tmp/oliverr/remind-NGFS/`. If switched off, nothing changes for REMIND users.

The change of file and variable names in `45_carbonprice/NDC` has been already merged with PR #656.


christophbertram (changes requested):
Thanks a lot for the very clear PR explanation:
I found 3 minor places where edits seem required:
modules/21_tax/on/equations.gms L96: would prefer to also here have all 4 components for consistency. Perhaps also worth a short discussion in the next REMIND meeting
modules/21_tax/on/postsolve.gms L22: Here you forgot to add the other 3 components, right?
modules/21_tax/on/presolve.gms L20: Same here

christophbertram (commented):
Looks good to me know, but given the extensive nature of this PR, good to have 2 more pairs of eyes...

christophbertram (approved):

core/datainput.gms
```diff
@@ -21,6 +21,9 @@ pm_globalMeanTemperature(tall)              = 0;
 pm_globalMeanTemperatureZeroed1900(tall)    = 0;
 pm_temperatureImpulseResponseCO2(tall,tall) = 0;
 
+*OR* initialize regional tax that is only required in 48_carbonpriceRegi
+pm_taxCO2eqRegi(ttot,regi)                 = 0;
```
LaviniaBaumstark, line 25: why do you declare it here and not in 48_carbonpriceRegi?

modules/48_carbonpriceRegi/none/realization.gms
```diff
@@ -0,0 +1,12 @@
+*** |  (C) 2006-2020 Potsdam Institute for Climate Impact Research (PIK)
+*** |  authors, and contributors see CITATION.cff file. This file is part
+*** |  of REMIND and licensed under AGPL-3.0-or-later. Under Section 7 of
+*** |  AGPL-3.0, you are granted additional permissions described in the
+*** |  REMIND License Exception, version 1.0 (see LICENSE file).
+*** |  Contact: remind@pik-potsdam.de
+*** SOF ./modules/48_carbonpriceRegi/none/realization.gms
+
+*####################### R SECTION START (PHASES) ##############################
```
LaviniaBaumstark, line 9: I would assume that in this realization the additional taxes are also declared but set to 0

LaviniaBaumstark (commented):
I ike the introduction of pm_taxCO2eqSum

core/datainput.gms
orichters: So you would prefer moving this to 48_carbonpriceRegi/none?

LaviniaBaumstark: yes

Renato-Rodrigues:
Could I ask if there was any previous discussion with a bigger group on the best place to locate this module within the remind code? 
The 47_regipol module was designed to be the last module in chain of modules that change the pm_taxCO2eq behaviour. I am afraid that adding a module just after it will cause undesirable issues for all the code and work related with REMIND-EU targets.
Could we open a discussion on this topic before merging moving on to merge these changes to the REMIND code?

christophbertram:
Hi Renato, 
the idea is that module 48 does not change pm_taxCO2eq, but rather only pm_taxCO2eqRegi. And both are, together with pm_taxCO2eqHist and pm_taxCO2eqSCC components of pm_taxCO2eqSum.
Nevertheless, it's clear that the overall behaviour of pm_taxCO2eqSum since quite some time already is a rather complex mix of different adjustments, and by no means can it be expected that all combinations of settings for modules 45,47, and 48 yield a meaningful result. So the modelers really need to understand the logic behind each realization.
I think that therefore structurally deciding on which module adjusts which parameter is very helpful. With this commit, the logic would be that 45 and 47 both adjust pm_taxCO2eq, with 47 basically overruling adjustments from 45. 48 in turn only adjusts pm_taxCO2eqRegi. Happy to have a dedicated meeting on this next week, might indeed be helpful, and we can hold off merging until then.

PR #660, opened 2022-02-16
new module 46_carbonpriceRegi

orichters:
This PR replaces [PR #647](https://github.com/remindmodel/remind/pull/647) because I wasn't aware of the fact that renaming a branch irrevocably closes the PR…

For the NGFS project, this PR introduces the module `46_carbonpriceRegi`. It adds a regional markup to the carbon prices that can be used to combine a `45_carbonprice` module to additionally achieve NDC or netZero targets.

1. `core`
- adding `pm_taxCO2eqRegi` as "additional regional CO2 tax path" which is usually set to zero. Its value is added to `pm_pvpRegi(ttot,regi,"perm")` which still covers the full GHG taxation

2. `21_tax/on`
- add `pm_taxCO2eqRegi` to all relevant taxation variables

3. `46_carbonpriceRegi/netZero`
- adaption of a realization implemented by Jerome Hilaire.
- in `datainput.gms`, allows to specify netZero years for regions and brings them to zero in this year by adapting `pm_taxCO2eqRegi`. Before and after this year, uses a triangular shape for tax rates, going to zero in 2100. Can easily be adapted "on the fly" within remind.

4. `46_carbonpriceRegi/NDC`
- uses the `UNFCCC_NDC` database to import both NDC targets and netZero targets from mrremind.
- chooses the years where a target with sufficiently coverage within the region is given. Adapts `pm_taxCO2eqRegi` in all these years such that all targets are met. Uses a linear interpolation between these years.

5. `47_regipol/regiCarbonPrice`
- set the `pm_taxCO2eqRegi` to zero for the regions covered by regipol. @Renato-Rodrigues should check that.

6. `start` scripts
- remove warnings if EOF marker in log is missing
- run NDC start scripts also for 46 module, move this after input data is distributed because input data are required

If any 45 realization is selected, both 46 realizations start only after the 10th iteration, hoping that the global carbon price has converged sufficiently.

Runs with these new realizations are were done in `/p/tmp/oliverr/remind-NGFS/`, I do some new runs using `optimization = testOneRegi` to check whether everything works after moving the module from 48 to 46.

The change of file and variable names in `45_carbonprice/NDC` has been already merged with PR #656.


Renato-Rodrigues:
could you also set pm_taxCO2eqRegi to zero around line 260 of the postsolve in regipol, just after the code lines:
```
***--------------------------------------------------
*** Regional carbon princing
***--------------------------------------------------
$IFTHEN.regicarbonprice not "%cm_regiCO2target%" == "off" 
*** display pm_taxCO2eq;
```

LaviniaBaumstark (approved):
I still like the introduction of pm_taxCO2eqSum

christophbertram (commented):
This looks mostly good to me.

I wonder if we should already use this PR to also add the comments into the default.cfg that we discussed yesterday (the more detailed tutorial can come later).

And for the commenting of the pm_taxCO2eqSum in 21_tax/on/postsolve.gms and presolve, I think this should not be under the "save reference level" comment, but rather get a separate line of code

orichters:
@christophbertram: I moved the summation and added some "be careful" to default.cfg, but a more detailed explanation probably has to be written by someone that understood better what regipol actually does under which conditions and switch configurations.

PR #661, opened 2022-02-17
improve tutorials, enable more empty cells in scenario config

orichters:
- For scenario configs, allow that also the columns `model`, `output`, `regionMapping` and `inputData` may have empty cells and use the default
 - clarify a bit the very welcome tutorial by @fbenke-pik on input data update; fix links


fbenke-pik (approved):
Looks good to me!

LaviniaBaumstark (approved):
nice cleaning in start.R!

PR #662, opened 2022-02-17
add constraint to have equal shares of synfuels in biofuel+synfuel total per transport subsectors

fschreyer:
After a recent discussion with Robert, we said we would for now want to have equal shares of synfuels and biofuels per transport subsector (LDV, HDV, Bunkers) as there seems to be no real reason why we should put all biofuels into bunkers and all synfuels into trucks or vice versa which REMIND does quite arbitrarily at the moment. 

So this constraint, makes the synfuel share in carbon-neutral liquids (biofuel+synfuels) the same in ``fepet`` (LDV), ``fedie.ES`` (HDV), and ``fedie.other`` (Bunkers). 

We are planning to soon add reporting of that fuel distinction in the EDGE-T reporting based on the shares in ``vm_demFeSector``.

### DIFF
--- a/core/sets.gms
+++ b/core/sets.gms
@@ -2295,8 +2295,8 @@ alias(esty,esty2);
 alias(rlf,rlf2);
 alias(regi,regi2,regi3);
 alias(steps,steps2);
-alias(all_emiMkt,emiMkt);
-alias(emi_sectors,sector);
+alias(all_emiMkt,emiMkt,emiMkt2);
+alias(emi_sectors,sector,sector2);
 alias(sector_types,type)
 
 ***-----------------------------------------------------------------------------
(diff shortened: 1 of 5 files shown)


Renato-Rodrigues:
Why the shares are calculated on bio and synfuel only? Wouldn't you need to consider fossils also in the equation to enforce the same bio+syn on different modes?

Maybe I am missing something, but at a quick look you are enforcing the share of synfuels in bio+syn to be the same, but you could still have an specialization on for example, having a lot of fossil liquids in one mode, and very little in the other mode, and still respects the shares equation. If this is the intended behavior just disregard my comment.

fschreyer:
> Why the shares are calculated on bio and synfuel only? Wouldn't you need to consider fossils also in the equation to enforce the same bio+syn on different modes?

I would only want to enforce these equal shares for carbon-neutral fuels. Once we use the emissions markets and have different carbon prices for e.g. trucks and bunkers, different shares of fossil fuels per sector are a meaningful endogenous model outcome.

Renato-Rodrigues (approved):

Loisel (approved):
Looks OK. Clever to write it in this way to avoid DIVBYZERO.

fschreyer:
> Looks OK. Clever to write it in this way to avoid DIVBYZERO.

Taken from you from the equation above, Alois :)

PR #665, opened 2022-02-18
adapt module description for 45,46,47

orichters:
move module description and possible conflicts from `default.cfg` to `module.gms`

### DIFF
--- a/config/default.cfg
+++ b/config/default.cfg
@@ -221,7 +221,6 @@ cfg$gms$banking  <- "off"           # def = off
 cfg$gms$carbonprice <- "none"           # def = none
 
 # ***---------------------    46_carbonpriceRegi  ----------------------------------------
-# * The 46_carbonpriceRegi module implements a regional taxCO2eq markup that is added to the tax implemented by 45_carbonprice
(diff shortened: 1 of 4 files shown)


Renato-Rodrigues:
My suggestion was to add the text below directly to the default.cfg file, in the 45_carbonprice description (either in line 2020 o 2022), as this is the first one that deals with the co2 tax determination:

`# Carbon prices are potentially defined by three modules: 
# - 45_carbonprice: define the carbon price necessary to reach global emission targets following specific price trajectories. 
# - 46_carbonRegiPrice: apply markups above 45_carbonprice estimations to reach specific NDC or net-neutral targets. 
# - 47_regipol: under the regiCarbonPrice realisation, define more detailed region or emissions market specific targets, overwriting the carbon price for selected regions.` 

But I am happy also with Christopher suggestion, adding the help text directly to each specific module in the default.cfg file.

orichters:
I now also re-added the text in default.cfg, mangling together the texts by you two.

Renato-Rodrigues (approved):

christophbertram (approved):
Perfect, thanks a lot!

PR #671, opened 2022-02-25
allow to differentiate path_gdx_ref and path_gdx_refpolicycost

orichters:
For the scenario config, a new column name `path_gdx_refpolicycost` is introduced following the debate in https://github.com/remindmodel/development_issues/issues/10:

- `path_gdx_ref` is used as before for fixing until `cm_startyear`,
- `path_gdx_refpolicycost` is used as a baseline for policy cost calculation.

If no column `path_gdx_refpolicycost` is specified, `path_gdx_ref` is used as before, so this PR should not alter the working of any existing scenario config file.

`path_gdx_refpolicycost` is integrated in the determination of subsequent runs, so subsequent runs will wait until the run for comparison of policy cost is finished, although technically the `fulldata.gdx` is only required in the output generation at the end of the subsequent run. If this slows down the process, you can leave the `path_gdx_refpolicycost` cell empty and do the policy cost calculation manually after all runs are finished, or you can add a `_` after the run name in the cell to use an older, already finished run.

### DIFF
--- a/scripts/output/single/reportCEScalib.R
+++ b/scripts/output/single/reportCEScalib.R
@@ -17,8 +17,8 @@ require(colorspace)
 require(gdx)
 require(grid)
 
-gdx_name     <- "fulldata.gdx"        # name of the gdx  
-gdx_ref_name <- "input_ref.gdx"       # name of the reference gdx (for policy cost calculation)
+gdx_name     <- "fulldata.gdx"             # name of the gdx
+gdx_ref_name <- "input_refpolicycost.gdx"  # name of the reference gdx (for policy cost calculation)
 
 if(!exists("source_include")) {
   #Define arguments that can be read from command line
--- a/scripts/output/single/reportCEScalib.R
+++ b/scripts/output/single/reportCEScalib.R
@@ -18,16 +18,13 @@ require(gdx)
 require(grid)
 
 gdx_name     <- "fulldata.gdx"             # name of the gdx
-gdx_ref_name <- "input_refpolicycost.gdx"  # name of the reference gdx (for policy cost calculation)
 
 if(!exists("source_include")) {
   #Define arguments that can be read from command line
    outputdir <- "output/R17IH_SSP2_postIIASA-26_2016-12-23_16.03.23"     # path to the output folder
-   readArgs("outputdir","gdx_name","gdx_ref_name")
+   readArgs("outputdir","gdx_name")
 } 
 gdx      <- file.path(outputdir,gdx_name)
-gdx_ref  <- file.path(outputdir,gdx_ref_name)
-if(!file.exists(gdx_ref)) { gdx_ref <- NULL }
 scenario <- getScenNames(outputdir)
 
 #---------------------------------------------------------------------------
--- a/scripts/output/single/LCOEPlot.R
+++ b/scripts/output/single/LCOEPlot.R
@@ -12,7 +12,7 @@ library(lusweave)
 
 
 ############################# BASIC CONFIGURATION #############################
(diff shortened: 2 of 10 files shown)


scripts/output/single/reportCEScalib.R
```diff
@@ -17,8 +17,8 @@ require(colorspace)
 require(gdx)
 require(grid)
 
-gdx_name     <- "fulldata.gdx"        # name of the gdx  
-gdx_ref_name <- "input_ref.gdx"       # name of the reference gdx (for policy cost calculation)
+gdx_name     <- "fulldata.gdx"             # name of the gdx
+gdx_ref_name <- "input_refpolicycost.gdx"  # name of the reference gdx (for policy cost calculation)
```
LaviniaBaumstark, line 21: why is here this gdx loaded?

bs538 (commented):
Thanks, I like the way this is set up. (Hope it's OK if I leave the technical review to Lavinia).

Is this supposed to be used in the default scenario config? If yes, you'd have to add a column with (SSPx/SDP/SDP_XX)-NPi as path_gdx_refpolicycost.

also putting @johanneskoch94 in the loop as he put together the previous (manual) policycost reporting via Rscript output.R

scripts/output/single/reportCEScalib.R
orichters: In the sense of "why do we need it here anyway?" or "should be input_ref.gdx" as before?

Was introduced here: https://github.com/remindmodel/remind/commit/70d1ebd2d6a7b5d34d7abf06b279521744e5bdd2

orichters:
@bs538: No, all the `path_gdx…` are not in the default config and cannot be, because you need to start with some run at some point before you can use it for some later comparison.

scripts/output/single/reportCEScalib.R
LaviniaBaumstark: in the sense "should be input_ref.gdx" as before. I wondered why we calculate policy costs in this script. And actually, no input_ref.gdx file is used in the script could be deleted.

bs538:
> @bs538: No, all the `path_gdx…` are not in the default config and cannot be, because you need to start with some run at some point before you can use it for some later comparison.

I thought in principle they can, just referring to runs that already happen previously in the cascade (like it's already done for path_gdx_ref, where the corresponding NDC run that has run previously is used). But using that here with NPi as path_gdx_policostref would mean that our cascade would get an additional step Base-NDC-*NPi*-(all other policy runs), which is probably undesirable. @LaviniaBaumstark using NDC (and not NPi) as ref for the fixing was decided on some point based on giving a better near-term picture right? (counter-intuitive as it is though...). perhaps this validation is a good moment to revisit this and check if that's still necessary to be done in this counter-intuitive order?

orichters:
Yes, @bs538, you are right, this would imply a next step in the cascade and as I wrote above in the PR description, you have to decide whether you want to do that. Therefore, I did not change any scenario config files. This PR just offers the possibility to implement it this way or to select an older (already converged) run as refpolicycost run.

scripts/output/single/reportCEScalib.R
orichters: done

LaviniaBaumstark (approved):

LaviniaBaumstark:
I would not use this functionality in the current validation runs as it would slow down the process. We decided to use NDC runs for fixing the first time as those scenarios are in line with some long term climate policy assumptions. To ensure that NCD and NPi are identical during the fixing period we also run the NPi scenario afterwards based on the NDC scenario. I do not think that we should change this.

johanneskoch94 (approved):

Issue #174, opened 2020-05-13
Subsequent runs not starting 


giannou:
This run:
`/p/tmp/giannou/sa2_updated/output/NDC_spvL_CDR_2020-05-12_20.12.24`
has 3 subsequent runs: 
`$subsequentruns
[1] "PkBudg1100_spvL_CDR" "PkBudg900_spvL_CDR"  "PkBudg1300_spvL_CDR"`
They weren't started when the run was finished and they don't start when I restart the basis-run with
`Rscript start.R --restart` from the mail folder. 
Any idea what is going on here? I get no error message and in the `log.txt` file it looks as though the subsequent runs don't exist.

Other runs of mine had the same problem.

giannou:
Ok I think I have an idea what is causing the problem. When a run is preempted, once (automatically) restarted the presence of `full.gms` in the run folder makes the `prepare_and_run.R`script think that it is a run that is being restarted manually and sets the object `start_subsequent_runs`object to `FALSE`

We might want to fix this

johanneskoch94:
This is the specifically set behavior. When restarting a run subsequent runs are not started. Because it doesn't work, or because one doesn't want to do this by default, I don't know. Try changing the flag in line the prepare_and_run.R script.

johanneskoch94:
Oh sry... didn't see your comment ;)

LaviniaBaumstark:
I think for restarted runs we really do not want to also restart the subsequent runs. But for automatically preempted runs we would like to also start the subsequent runs. There is an option to run a few scripts before a run is preempted. Maybe this could be handed there.

PR #672, opened 2022-02-25
enable option to use --restart to also restart subsequent runs

orichters:
This PR enables an option to use --restart to also restart subsequent runs.

- Running `Rscript start.R --restart`, it asks you whether you also want to restart the subsequent runs of the run you want to restart.
- `start.R` stores this in `cfg$restart_subsequent_runs` and saves this to `config.Rdata` in all the selected run folders.
- Note that I also write the new `cfg$results_folder` and `cfg$slurmConfig` into this file, although I'm not sure whether this is correct – the fact that this was missing is probably a bug if indeed (as suggested in the comment) the folder was moved, because `prepare_and_run.R` uses `cfg$results_folder` in several occasions, then using the wrong folder. Probably was never a problem because hardly anyone moves folders and restarts run, I guess. But please check carefully if that is correct. 
- `prepare_and_run.R` normally sets `start_subsequent_runs` to FALSE for restarted runs. But if `isTRUE(cfg$restart_subsequent_runs))` is satisfied, it starts them nevertheless. As `isTRUE(cfg$whatever)` is `FALSE` if `whatever` does not exist, this check should be safe.

A related issue was opened by @giannou: [Subsequent runs not starting #174](https://github.com/remindmodel/remind/issues/174).
But I haven't fully understood what automatically preempted run means. @LaviniaBaumstark there states: "There is an option to run a few scripts before a run is preempted." Basically you just need to set `cfg$restart_subsequent_runs` to `TRUE` in `config.Rdata` in these scripts, wherever they are.

Tested: `/p/tmp/oliverr/remind-smallfix/output/SSP2-Base_2022-02-25_13.16.32` was restarted and successfully started a subsequent run.

(also added some hint on using testOneRegi to tutorial 3)

### DIFF
--- a/scripts/start/prepare_and_run.R
+++ b/scripts/start/prepare_and_run.R
@@ -896,7 +896,7 @@ run <- function(start_subsequent_runs = TRUE) {
   # Use the name to check whether it is a coupled run (TRUE if the name ends with "-rem-xx")
   coupled_run <- grepl("-rem-[0-9]{1,2}$",cfg$title)
   # Don't start subsequent runs form here if REMIND runs coupled. They are started in start_coupled.R instead.
-  start_subsequent_runs <- start_subsequent_runs & !coupled_run
+  start_subsequent_runs <- (start_subsequent_runs | isTRUE(cfg$restart_subsequent_runs)) & !coupled_run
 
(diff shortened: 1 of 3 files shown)


LaviniaBaumstark:
@giannou do I remember correctly that you had an issue with copied output folders which were restarted? Was this fixed and might this change be a problem?

LaviniaBaumstark (approved):

PR #674, opened 2022-02-28
Refactoring and code cleaning of various deprecated developments

fschreyer:
This PR does some code cleaning of previous developments:

* remove constraint on equal shares of biofuels/synfuels in transport (added in this [PR](https://github.com/remindmodel/remind/pull/662), eventually I realized this is not really necessary during optimization and can be added only in reporting -> see this [PR](https://github.com/pik-piam/remind2/pull/198)
* remove deprecated ``cm_regipol_slope_beforeTarget`` switch in regipol module
* clean up CCU module a bit
* some minor clean up in core equations
* remove deprecated share equations in buildings simple
* refactor CES markup ppfen set in buildings simple


LaviniaBaumstark (approved):
I assume that you are sure that the functionality you are deleting in regiCarbonPrice is no longer needed

rkrekeler (approved):

fschreyer:
@Renato:

On your comment. Not sure whether I entirely understand it. We can touch upon it briefly in the other meeting tmrw. I just want to remove this equation again because having sensible bio/synfuel shares etc. is something we can also just do in the reporting. I agree that the best would be to have cost at some point determine this but currently I don't see from a bottom-up perspective yet how favoring one or the other could be justified.

PR #683, opened 2022-03-03
bugfix in start_bundle_coupled.R

orichters:
if input_ref is a run to be waiting… for, but input_bau a fully specified gdx.

The current setup generates `/p/tmp/oliverr/remind-NGFS/output/C_./output/C_SSP2-Base_bIT-rem-3/fulldata.gdx-rem-5/fulldata.gdx`.


christophbertram (approved):
"Bugfix" is a bit strong, as the previous behaviour was actually ok for the standard usage of coupled runs. What we are doing here is kind of a special workaround, and so we now add functionality for using such a workaround...

orichters:
Yes, it is a "bug" in the sense of "didn't work as I expected" :)

LaviniaBaumstark (approved):

PR #685, opened 2022-03-04
add description to model runs

orichters:
This PR is a suggestion developed jointly with @chroetz, namely adding a short description to each REMIND run via the scenario config file.

The description will be saved in the .RData file and will be used by compareScenario2 that prints it on the first pages of the document to allow for a quick understanding of what these scenarios mean. The description is also helpful if one changes small things in a model run to note these things, such that later comparing runs with the same title but slightly different settings is much easier, particularly after a while.

I realized this idea by adding cfg$description to default.cfg. The different start scripts search for a description column in the scenario config files, and if none exist, uses a text summarizing the title and the config file this run was created by.

Note that the way it is implemented now, any change to the description in default.cfg will be overwritten for bundles of runs. Is that a problem?

prepare_and_run.R copies the description to cfg$gms$c_description, and in main.gms this description is written in the explanatory text of a set that contains a hint where to find the description. The reason for this small hack is that the length of strings used in GAMS variables is very limited.

A non-coupled run with this feature is `/p/tmp/oliverr/remind-smallfix/output/SSP2-Base_long_2022-02-28_17.58.17`, whereas `/p/tmp/oliverr/remind-smallfix/output/C_SSP2-Base_bIT-rem-1` shows it is correctly written to the config.RData file also for coupled runs (I killed the run then).

Alternatives or improvements of course warmly welcome.


Renato-Rodrigues (approved):

chroetz (approved):
The R code looks fine to me. I don't know enough about the GAMS part to judge that. Also, I haven't tested a full REMIND run with this version

chroetz:
It would be nice to have to default scenario_configs documented by this description

orichters:
> It would be nice to have to default scenario_configs documented by this description

I agree! I would ask in the remind meeting that everyone tries to document the scenario config files (s)he is supervising. But maybe a good idea if I add the NGFS scenario config as an example already to this PR.

orichters:
I added the description field for the NGFS runs

christophbertram (approved):
Thanks a lot, useful innovation. I guess you have tested the correct functioning of the start....R scripts with the changes.

LaviniaBaumstark (approved):
cool, thanks
I will add descriptions for the validation scenarios soon

PR #687, opened 2022-03-07
Documenting scenario_config_coupled and fixed path_gdx coupling bug

gabriel-abrahao:
This PR makes the behavior of `path_gdx` in `start_bundle_coupled.R` more intuitive, and updates `tutorials/4_RunningREMINDandMAgPIE.md` to include an explanation of the columns in `scenario_config_coupled.csv` and an overview of the files and folders created in a coupled run.


dklein-pik (approved):
Great work! Thank you.

tutorials/4_RunningREMINDandMAgPIE.md
```diff
@@ -64,7 +66,23 @@ See comments in the head section of the file. Most importantly you need to provi
 ### Configure the scenario_config_coupled.csv of your choice
 
 By default these are (A) `scenario_config_coupled_SSPSDP.csv` and (B) `scenario_config_SSPSDP.csv`. A provides some extra information
-for coupled runs (e.g. which run should be started). All other settings are taken from B. Thus every scenario in A must also be present in B.
+for coupled runs (e.g. which run should be started, specific MAgPIE configurations), with one scenario per row and settings on the columns. B is a normal `scenario_config` that defines REMIND settings, as explained in [3_RunningBundleOfRuns](https://github.com/remindmodel/remind/blob/develop/tutorials/4_RunningREMINDandMAgPIE.md). All other settings are taken from B. Thus every scenario in A must also be present in B.
+
+All the columns must be present in the `scenario_config_coupled.csv` file, but most of them can be left blank. The required ones are:
+   - `scenario`: The name of the scenario, must be unique and match the `scenario` column in REMIND's `scenario_config.csv`
```
orichters, line 72: isn't it called `title`?

tutorials/4_RunningREMINDandMAgPIE.md
```diff
@@ -64,7 +66,23 @@ See comments in the head section of the file. Most importantly you need to provi
...
+   - `start`: Defines if a scenario run should be started (1) or not (0). Overrides whatever is set in REMIND's `scenario_config.csv`. If you have an unfinished coupled run, it will automatically try to continue from the last coupling iteration (i.e. REMIND or MAgPIE run). If it already finished (i.e. reached `max_iterations`), it will simply try to restart the last REMIND run.
```
orichters, line 73: I think it is not overwriting, but rather only running models with both start flags set to 1, see https://github.com/remindmodel/remind/blob/develop/start_bundle_coupled.R#L96

tutorials/4_RunningREMINDandMAgPIE.md
```diff
@@ -64,7 +66,23 @@ See comments in the head section of the file. Most importantly you need to provi
...
+   - `qos`: The SLURM qos the coupled runs should be submitted to. Currently there's no default support for running a coupled run locally. Coupled runs usually take very long, with multiple runs of both models being performed sequentially (see below), so it's wise to choose a qos with a long time limit. As of writing, a typical coupled run with 5 iterations takes betwee 16 and 48 hours to complete, depending on various factors.
```
orichters, line 74: you could add: if no `qos` is specified, `short` is used.

tutorials/4_RunningREMINDandMAgPIE.md
```diff
@@ -64,7 +66,23 @@ See comments in the head section of the file. Most importantly you need to provi
...
+
+Other, optional columns allow you to make a run start only after another has finished, set starting conditions, and give you finer control over which data is fed to MAgPIE.
+   - `path_gdx`, `path_gdx_ref` and	`path_gdx_bau`: Override these same settings in REMIND's `scenario_config`, allowing you to set the results of another run to be used as a starting point for the solution (`path_gdx`), a reference to be used for years before `cm_startyear` (`path_gdx_ref`) and a business-as-usual scenario required in some cases (`path_gdx_bau`). 
+      - You can set these switches either to the full path of a `fulldata.gdx` file or simply to the name of another scenario in the file (without the "C_"!). So if you want a certain scenario (say `SSP2-NDC`) to use as starting point the results of a `SSP2-Base` scenario, you can simply set `path_gdx` to `SSP2-Base` and it will automatically locate the last `fulldata.gdx` in `SSP2-Base`, for example `path_remind/C_SSP2-Base-rem-5/fulldata.gdx` if 5 iterations where requested.
```
orichters, line 80: one could add that the feature that automatically older runs are used if `start` is set to `0` as explained in [tutorial 3](https://github.com/remindmodel/remind/blob/develop/tutorials/3_RunningBundleOfRuns.md) is not implemented for coupled runs.

orichters:
Great work, thanks, I added some comments and will have another look at it later!

tutorials/4_RunningREMINDandMAgPIE.md
dklein-pik: This line keeps scenarios only that exist in both (standalone and coupled) scenario_config files. It does not care about the start setting. This is done by this line https://github.com/remindmodel/remind/blob/fe9bf59480ad406c351e12b10eedbee17f237f41/start_bundle_coupled.R#L82. As a results, `start` in `scenario_config_coupled.R` indeed overrides the `scenario_config.csv`.

orichters: thanks!

tutorials/4_RunningREMINDandMAgPIE.md
chroetz: I think the name of the column does not matter. But it must be the first column. This is because the first column is used as row names of the settings objects, see [here](https://github.com/remindmodel/remind/blob/fe9bf59480ad406c351e12b10eedbee17f237f41/start_bundle_coupled.R#L75) the argument `row.names=1`. And the rownames are used as titles of the scenarios

gabriel-abrahao: Makes sense, but I suppose it's useful to list it as `title ` here anyway. Doing it.

tutorials/4_RunningREMINDandMAgPIE.md
gabriel-abrahao: It actually works for coupled runs too, but it uses the `max_iterations`-th iteration of the "parent" run. That could actually lead to bugs if one is to change `max_iterations` between runs. I'll try to document it.

orichters: I'm not so sure. It may work for the `path_gdx` column, but I think not for all the others such as _bau. I just implemented this some month ago for non-coupled runs. But yes, it should be clearly documented.

gabriel-abrahao: Interesting, checking the code, it's suppose to behave like that with all of them *except* `path_gdx`. I suppose it should work like that for it too, do you agree? If so, I can add that.

https://github.com/gabriel-abrahao/remind/blob/03a93d8c0147f7e9e7661517c6e83f81a9024b60/start_bundle_coupled.R#L306-L314

`path_gdx` just replaces the same setting in the regular `scenario_config`:
https://github.com/gabriel-abrahao/remind/blob/03a93d8c0147f7e9e7661517c6e83f81a9024b60/start_bundle_coupled.R#L167-L171

orichters: Sorry, I meant `path_gdx_ref` :/. You are probably right and I really just started looking into coupled runs, so I'm not so sure. I'm just a bit surprised that these things are implemented for coupled runs, but were not for a normal bundle of runs. But of course, there it was a bit more tricky, because the folder names contain time stamps.

I suggest you check with David whether it should be added for path_gdx as well, and possibly also the new `path_gdx_refpolicycost`.

chroetz (approved):

chroetz:
Thanks! This is very helpful

tutorials/4_RunningREMINDandMAgPIE.md
gabriel-abrahao: I did some tests and looks like it was an actual bug, setting up a bad path in case you used a scenario name in `path_gdx`. Fixed it in https://github.com/remindmodel/remind/pull/687/commits/ef43a382389d8c66442e529669accc43ae52606e

I'll talk to him, but I think it's best to keep `path_gdx_refpolicycost` out of this for now since it's a very REMIND-specific postprocessing setting. Should I merge?

orichters: Yes, I think you can merge it. There may be some later improvements and we can add path_gdx_refpolicycost at any time, but I think this should not impede your merge.

dklein-pik: If I understand the discussion correctly, the question is why it was not possible so far to specify the name of a scenario in the `scenario_config.csv` for the `input.gdx`, from which the `fulldata.gdx` is then automatically searched. The answer is: it is possible, but it was [realized differently](https://github.com/remindmodel/remind/blob/219c054766a974f35acf2dfaffd58efd8e701b41/start_bundle_coupled.R#L122). The reason is that coupled runs have no timestamp in the name. So a scenario can always exist only once in the current output folder. Either it already exists. Then it can only be continued if the iteration count increases (mostly because the run was aborted at a low iteration count and you want to continue it). Or it does not exist (you start at iteration 1), then you have to search in the same folder for a scenario with a different name (`oldrun` column in `scenario_config_coupled,csv`), or you have to search elsewhere (`path_remind_oldruns` and `path_magpie_oldruns`) either for the same scenario (leave `path_gdx` column blank) or a scenario with a different name (`oldrun` column).

In other words, there are usually two different cases:

1) Continuing the same scenario in the current output folder: then you don't need to specify anything for `path_gdx` in `scenario_config.csv`. The script looks for the last successful REMIND or MAgpIE iteration in the same folder and continues with the corresponding model.
2) Start at iteration 1 and use an old run from somewhere else: then set `path_remind_oldruns` and `path_magpie_oldruns`. If the old run has a different scenario name, put it in `oldrun`.

The change you suggest only becomes active if no `fulldata.gdx` was found during all the automatic search described above, i.e. if there is still no real path in `cfg_rem$files2export$start['input.gdx']`. This should never be the case, or only if you want to use the `fulldata.gdx` of a run that does not exist yet, but will exist when the current run starts. If that was the intention, we should adopt your suggested changes.

gabriel-abrahao: Thanks for the thorough explanation! Yes, that change was for the case when you want to set a new coupled run to use the `fulldata.gdx` of another run that doesn't exist yet. The use case I foresee is to speed up convergence in chains of runs. 

Say you have an `NDC` scenario a `policy1` one that uses `NDC` as reference, and a `policy2` that also uses `NDC` as a reference but carbon prices from `policy1` (so it has to wait for it to finish). If we do it like we're doing it now, this chain will work fine (with `path_mif_ghgprice_land`), but `policy2` will start from the regular `input.gdx`. With this change, we can set it to start the optimization from `policy1`, or any other run that is likely much closer to the solution, and do it when starting the bundle, before any of this has actually run. 

It's not a big deal, but might help speed things up a bit. Plus it makes the behavior of the  `path_*` columns consistent with each other, for predictability sake.

dklein-pik: I see. Thanks for explainaing! Makes perfect sense.

PR #688, opened 2022-03-08
Fix bad comment that leads to warning

johanneskoch94:


johanneskoch94:
This is the warning that pops up during the preparation of the run:

Warning in gms::singleGAMSfile(mainfile = cfg$model, output = file.path(cfg$results_folder,  :
  Catched a command which could not be translated (!! $include "./modules/37_industry/subsectors/input/vm_cesIO_scales.inc";)

LaviniaBaumstark (approved):

PR #691, opened 2022-03-11
provide config files in call to `compareScenarios2()` via output.R

chroetz:
When calling `compareScenarios2()` via output.R, the function is now told the location of the config.Rdata files for each scenario and the path to `default.cfg`. This enables the new functionality implemented in #685 and https://github.com/pik-piam/remind2/pull/206


orichters (approved):
Thank you!

PR #696, opened 2022-03-14
set SOF and EOF statements to actual filenames

orichters:
to facilitate debugging…

Searched for files missing this information using:
```
grep -L "\*\*\* SOF.*$" modules/*.gms modules/*/*.gms modules/*/*/*.gms
grep -L "\*\*\* EOF.*$" modules/*.gms modules/*/*.gms modules/*/*/*.gms
```
Add `*** SOF` and `*** EOF` and then run the following command on all files:
```
for i in modules/*.gms modules/*/*.gms modules/*/*/*.gms
do 
   sed -i "s>\*\*\* SOF.*$>\*\*\* SOF ./${i}>g;s>\*\*\* EOF.*$>\*\*\* EOF ./${i}>g" $i
done
```

### DIFF
--- a/modules/01_macro/singleSectorGr/realization.gms
+++ b/modules/01_macro/singleSectorGr/realization.gms
@@ -4,7 +4,7 @@
 *** |  AGPL-3.0, you are granted additional permissions described in the
 *** |  REMIND License Exception, version 1.0 (see LICENSE file).
 *** |  Contact: remind@pik-potsdam.de
-*** SOF ./modules/01_macro/singleSectorGr.gms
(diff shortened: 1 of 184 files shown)


0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
# :+1:

Nice.

@LaviniaBaumstark probably this should be checked in [`gms::codeCheck()`](https://github.com/pik-piam/gms/blob/master/R/codeCheck.R), no?

LaviniaBaumstark:
yes, that would be nice!

LaviniaBaumstark (approved):

PR #700, opened 2022-03-17
Use adjusted policy costs mif, if available

johanneskoch94:
The idea is to use the adjusted policy costs mif, if it was created.
It contains all the same variables as the normal mif, with some adjusted and additional policy costs variables.

### DIFF
--- a/scripts/utils/run_compareScenarios2.R
+++ b/scripts/utils/run_compareScenarios2.R
@@ -20,10 +20,16 @@ run_compareScenarios2 <- function(outputdirs, shortTerm, outfilename, regionList
   scenNames <- getScenNames(outputdirs)
   # Add '../' in front of the paths as compareScenarios2() will be run in individual temporary subfolders (see below).
   mif_path  <- file.path("..", outputdirs, paste("REMIND_generic_", scenNames, ".mif", sep = ""))
+  mif_path_polCosts  <- file.path("..", outputdirs, paste("REMIND_generic_", scenNames, "_adjustedPolicyCosts.mif", sep = ""))
   hist_path <- file.path("..", outputdirs[1], "historical.mif")
   scen_config_path  <- file.path("..", outputdirs, "config.Rdata")
   default_config_path  <- file.path("..", "config", "default.cfg")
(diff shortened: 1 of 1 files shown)


scripts/utils/run_compareScenarios2.R
```diff
@@ -20,10 +20,16 @@ run_compareScenarios2 <- function(outputdirs, shortTerm, outfilename, regionList
   scenNames <- getScenNames(outputdirs)
   # Add '../' in front of the paths as compareScenarios2() will be run in individual temporary subfolders (see below).
   mif_path  <- file.path("..", outputdirs, paste("REMIND_generic_", scenNames, ".mif", sep = ""))
+  mif_path_polCosts  <- file.path("..", outputdirs, paste("REMIND_generic_", scenNames, "_adjustedPolicyCosts.mif", sep = ""))
   hist_path <- file.path("..", outputdirs[1], "historical.mif")
   scen_config_path  <- file.path("..", outputdirs, "config.Rdata")
   default_config_path  <- file.path("..", "config", "default.cfg")
 
+  # Use adjustedPolicyCosts mif, if available
+  if (file.exists(mif_path_polCosts)) {
```
chroetz, line 29: `mif_path_polCosts` will be a vector of several paths. Maybe use something like `if (all(file.exitst(mif_path_polCosts)))` or `mif_path <- ifelse(file.exitst(mif_path_polCosts), mif_path_polCosts, mif_path)`?

orichters: @chroetz is right here, I suggest to use the second one. There will always be runs where the policy costs where calculated afterwards and some where the `path_gdx_refpolicycost` run was already finished. Importantly, adding a run without a _adjustedPolicyCosts.mif does then not change the behavior of the rest, which I think is good coding.

johanneskoch94: Thanks for the review! See 9a9f0db.

orichters (approved):
Great, thanks!

chroetz (approved):

PR #667, opened 2022-02-21
Add H2 and electricity HTH trajectories for industry and introduce markup cost for H2 and electricity HTH

fschreyer:
This
* adds small amounts of H2 and HTH electricity in industry sectors for baseline trajectories 
* introduces markup cost on CES nodes of H2 and electricity HTH in industry sector to represent demand-side cost of electrification/H2 technologies

For results and discussion, see this [issue](https://github.com/remindmodel/development_issues/issues/95). 

To be discussed further.

PR #704, opened 2022-03-19
H2 and electricity HTH trajectories and markup cost in industry

fschreyer:
This is can be merged as a pragmatic solution of introducing some H2 and HTH electricity in the baseline, too. Follows up on #667 on which I somehow could not push anymore.  It introduces

* H2 and high-temperature heat electricity trajectories for the baseline as a share of gas/low-temperature heat electricity trajectories from input data -> the share can be controled by the switch ``cm_Ind_H2HTH_Traj``. By default this switch is ``cm_Ind_H2HTH_Traj = "standard"`` which means that the H2/HTH share after 2050 is 30% of gas/WLTH. See default.cfg for how to set the switch specifically. Remember recalibrating after adjusting the switch as the switch only affects the calibration. 
* possibliity of putting a share of FE into the pm_cesdata("offset_quantity") is introduced, but commented out -> could be used future if larger differences between baseline and policy deployment are needed
* CES markup cost on industry H2 and electricity HTH use are introduced to reflect additional technology cost of using those carriers in industry: By default they are set to 200 USD/MWh(Elec) for electricity HTH use, 100 USD/MWh(H2) for H2 use. Again, this can be adapted via a switch ``cm_CESMkup_ind``. This switch can be used to construct specific electrification or hydrogen-focused policy scenarios by changing the markup cost in policy runs. 

Results close to the current setup can be found in ``/p/tmp/schreyer/Modeling/REMIND_H12/Current/compScen-IndTraj_Offset20-2022-03-07_09.16.08-H12.pdf`` -> ``SSP2EU−PkBudg500_H2HTH_offset30_20_Mkup`` runs. 

After merging this PR, recalibration of REMIND is necessary. 

For details, see https://github.com/remindmodel/development_issues/issues/95


0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q (approved):
I think we will need to put portions of the baseline H2/HTHel into `offset_quantity`, but that can be added later.

PR #705, opened 2022-03-22
new flags for start.R: --debug, and --interactive, all working with scenario_config file, --reprepare to support --testOneRegi and --debug for existing runs

orichters:
This PR has turned into a bigger project adding features to start.R.

--testOneRegi with scenario_config.csv
-----
If you run:
```
Rscript start.R --testOneRegi scenario_config.csv
```
until now the scenario_config file was ignored and a default run with `testOneRegi` was started. This is impractical if changes in submodules were to be tested that are not used by the default.cfg.

With this change, the above code starts the runs specified in the `scenario_config.csv`, but sets `optimization` to `testOneRegi`, so only one region and one iteration is performed. This allows not only to check whether a change in a module realization works, but also whether the chain of subsequent runs works flawlessly without waiting for potential long convergence of the normal runs. This testOneRegi version uses the normal scenarioname-timestamp folder names to allow for multiple subsequent runs.

For all runs, the slurmConfig specified by the user is used, overwriting potential specifications in the scenario_config file. The script allows the user to provide the region used for testOneRegi.

A successful set of runs with this setting are the 2022-03-22 runs found in `/p/tmp/oliverr/remind/output`.

Adding --debug option
-----
```
Rscript start.R --debug scenario_config.csv
```
starts the scenario_config.csv in debug mode. Shortcut for `--debug`: `-d`

Adding --interactive option
-----
```
Rscript start.R --interactive scenario_config.csv
```
This shows a list of all the runs (also with `start` = 0) and let the user select the runs he wants to start, instead of tediously editing the scenario_config file. Shortcut for `--interactive `: `-i`

If no scenario config file is provided (thus `Rscript start.R -i`), the script lets you select a scenario config file (not starting with scenario_config_coupled) and then the runs you want to start.

Adding --reprepare option
-----
For debugging, it is sometimes helpful, to restart a run, but in debug or testOneRegi mode. This can be achieved with:
```
Rscript start.R --debug --testOneRegi --reprepare
```
It moves the full.gms to full_old.gms and the fulldata.gdx to fulldata_old.gdx and therefore triggers a reload of full.gms. Therefore, the testOneRegi and debug option can be used to start the run in these modes, and the script allows the user to provide the region used for testOneRegi. They are of course not mandatory, so `Rscript start.R --reprepare` works as well and simply does restart + new generation of full.gms. Shortcut: `--d1R`

Improved --test mode
-----
--test mode does now 'ignore' all errors where a gdx could not be find in the sense that it prints the error message, but doesn't break, and tells you at the end how often that has happened. This way, you can fix all these errors in one go.

--test mode now works with --restart/--reprepare as well. Shortcut: `-t`.

Add short flags as shortcuts:
-----
- -d for --debug
- -i for --interactive
- -r for --restart
- -R for --reprepare ("big restart")
- -t for --test
- -1 for --testOneRegi

So you can test how selecting and starting runs in debug & testOneRegi mode will look like by running:
```bash
Rscript start.R -dit1
```

cleanup of start.R
-----
Tried to reduce redundant definitions

update of tutorials
-----
I added some options to tutorial three, but the --debug and --reprepare option should better be discussed in a yet to be written debug tutorial.

enable start groups
-----
In interactive mode, you can now not only select individual runs, but also groups with the same identifier in the start column. This way, you can still select all with start = 1, but you can also tweak your config file such that groups can be started jointly.

generalize chooseFolder to chooseFromList
-----
so it can be used for --interactive and more generally to select things from a list. I use it both for selecting a scenario config file (with allowempty for default and multiple=FALSE) and for selecting the runs.

fixed wrong comment in Rprofile
-----
- also fixed wrong comment and unclear messaging from .Rprofile: If a snapshot is specified, it doesn't check the never-used library for updating profiles first.


0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
@LaviniaBaumstark, could you approve this merge request, pretty please – with sugar on top!  That would save me a couple of hours of headaches every time I have to figure out whether REMIND or CONOPT is broken …

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
> That would save me a couple of hours of headaches …

Would, that is, if it worked … see review.

orichters:
still can't see a review by @0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q…

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
I can't find any option for me to mark this resolved.  From my experience, you should have a button for that.

orichters:
@0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q: The screenshot states "pending" which implies: ["Before you submit your review, your line comments are pending and only visible to you"](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/reviewing-proposed-changes-in-a-pull-request). And after you submitted it, you can also mark it as resolved. The code works as it should, though I agree that it is not easily understandable. `config.file` was initialized as `argv[1]`, so if the first argument is `scenario_config.csv`, this if-clause evaluates to `negative`, only if `--testOneRegi` is the first argument, it sets `config.file` to NA.

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
> @0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q: The screenshot states "pending" which implies: ["Before you submit your review, your line comments are pending and only visible to you"](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/reviewing-proposed-changes-in-a-pull-request). And after you submitted it, you can also mark it as resolved.

Well that is unintuitive as f$^% …

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
> The code works as it should, though I agree that it is not easily understandable. `config.file` was initialized as `argv[1]`, so if the first argument is `scenario_config.csv`, this if-clause evaluates to `negative`, only if `--testOneRegi` is the first argument, it sets `config.file` to NA. I think if I replaced `config.file` here by `argv[1]`, it may be more intuitive.

I see.  But having options after the operand (`command operand --options`) is highly unusual.  POSIX [explicitly recommends](https://pubs.opengroup.org/onlinepubs/9699919799/) otherwise (`command --options operand`), and I guess most people will expect it to be the other way around (like I did).
Also, the option of starting runs from scenario configuration files is not reflected in the documentation comment at the top of the file.

orichters:
I see, really took a while to understand what goes wrong.

I rewrote the code. The order of arguments now does not matter. The code also checks whether more than one config file is specified and throws an error then. Also, testOneRegi runs with default settings are now logged as testOneRegi, not default and write their settings to default.Rdata. In the tutorial, I switched to the recommended order `Rscript start.R --testOneRegi *.csv`

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q (approved):

gabriel-abrahao (approved):
Worked pretty well. Interactively started runs from `scenario_config` with `--interactive`, turned them into debug runs with `--debug`, and restarted coupled runs in debug mode with `--reprepare --debug` here: `/p/projects/piam/abrahao/scratch/develop/remind-restart`

LaviniaBaumstark (approved):

PR #709, opened 2022-03-24
move 'description' column in scenario_config files to the end

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
This is a suggestion to #685 (and 0ece0db).
The `description` fields in the scenario config files are certainly useful, but for me they get very much in the way of my workflow.  It's one huge column that I'm not particularly interested in (since I know what e.g. `SSP2EU-calibrate` is supposed to do), so I always have to hide it or scroll past it when checking or modifying a configuration.
I suspect most people will use them in a similar manner, looking at it once every couple of weeks for reference and ignoring it otherwise.
So I suggest to move them to the last column of the file.  One can still navigate to them fast if one needs to, but they are out of the way for daily business.
There's a small R script to do this (`/home/pehl/scripts/move_description.R`)
```R
#!/usr/bin/env Rscript
library(tidyverse)

for (f in list.files(path = './', pattern = '^scenario_config.*\\.csv$')) {
  read_delim(file = f, delim = ';', col_types = 'c') %>% 
  select(all_of(setdiff(colnames(.), 'description')), 
         all_of(intersect(colnames(.), 'description'))) %>% 
    write_delim(file = f, delim = ';', na = '')
}
```
to be run from the `./config/` directory.

So discuss and decide if this makes sense to you.  If you very much prefer to have the description in the front of the file, I can also keep hiding it.


orichters:
I'm ok with that in the sense that I'm indifferent.

I use the description field more regularly if I change something in the configuration to put a comment explaining this change so I can understand it later, but it wouldn't bother me too much to scroll to the end to do that.

But I would suggest to wait for @rauner to merge https://github.com/remindmodel/remind/pull/699 and then rerun your script to avoid merge conflicts. (edit: he did)

chroetz (approved):
It sounds reasonable to me to put the arbitrarily wide column `description` at the end

LaviniaBaumstark:
fine for me. Please make sure that no changes in the meantime are lost when merging

LaviniaBaumstark (approved):

orichters (approved):

Issue #173, opened 2020-05-13
Subsequent REMIND runs start even if basis-run fails or does not converge


giannou:
We might want to change this

orichters:
The problem is even worse: Even if an `Execution Error` is raised, the run is counted as `completed` (including sending a mail stating that) as long as some `fulldata.gdx` was created in some iteration and therefore [modelstat reading works](https://github.com/remindmodel/remind/blob/develop/scripts/start/prepare_and_run.R#L893). An example run is `/p/tmp/oliverr/remind-tests/output/SSP2-Base_inv1_welf2_2022-03-18_11.59.07`. I would suggest to add an additional check like that:
```
if (! file.exists("full.log")  || ! grep("*** Status", readLines("full.log"), fixed = TRUE, value = TRUE) == "*** Status: Normal completion"))
   stop("GAMS run did not complete normally")
```
But I have no idea whether this would actually do what I intend or whether it should be implemented differently.

LaviniaBaumstark:
actually, we ant to start subsequent runs even if the reference scenarios did not converge. Sometimes a scenario just is not converging within 100 iterations but only fails in a few markets at the end of the century. Those results should not be used for papers or project submissions but can already be used to understand results and get more insights.

giannou:
The status of unconverged runs is still `Normal completion`, so the formulation with `Execution Errors` is better

orichters:
If unconverged runs return `Normal completion`, then I don't see a difference between the two implementations (as long as there is no third/fourth… status that I haven't seen yet).

giannou:
`Normal completion` means GAMS has done it's part and has finished without `Execution error(s)` or `Compilation error(s)`. Further statuses refer to what CONOPT and REMIND have achieved, these could be checked with `modelstats::getRunStatus`

orichters:
Ok, but then we should stick to the first formulation, where it checks whether the log states "*** Status: Normal completion". This covers both converged and unconverged runs, so this is fine, but Execution or Compilation errors both lead to a stop.

orichters:
I close this. REMIND now stops if run fails, but the start of subsequent runs if non-converged seems intended.

PR #711, opened 2022-03-24
stop prepare_and_run if GAMS failed, improve logging

orichters:
stop if REMIND failed
-----
- currently, if GAMS failed at the start, the lack of a fulldata.gdx created an error while submitting the runstatistics
- but if an earlier iteration converged and REMIND failed then, R continued with output generation and started subsequent runs and send a mail that everything was fine.
- With this PR, prepare_and_run.R checks whether the GAMS log states "Normal completion". If not, GAMS had produced an error and the scripts stops without output generation and starting subsequent runs.
- If REMIND does not converge within the usual 100 iteration, it still finishes with "Normal completion", so this PR is not affecting this case, as discussed here: https://github.com/remindmodel/remind/issues/173
- I also check whether the GAMS files were generated etc, providing a clearer error if something went wrong

logging improvement
----
- I added the modelstat and number of iterations to the Model Summary section of the log.
- Fixed the `Settings are unknown in provided cfg (cfg$remind_folder)!` warning
- set in emulator.R the gginput call to quiet, as it fills `log.txt` with ~35 very similar and not really helpful success stories:
```
        Data successfully prepared for ggplot!
         data columns: .spat1 = region .temp1 = year .data1 = scenario .data2 = sample
         value columns:  .value1 = .value.x .value2 = .value.y
```
- added an extract of nashstat to the log.txt

For `/p/tmp/pascalwe/ariadne/remind/output/Base_2022-03-25_10.36.44`, it prints:
```
Infeasibilities extracted from full.lst with nashstat -F:
itr   region   solvestat              modelstat                 resusd     objval
  1   CHA      NORMAL COMPLETION        LOCALLY INFEASIBLE      0.333    45.0892103043586   F
... 953 infeasibilities omitted, show all with nashstat -a ...
 51   USA   TERMINATED BY SOLVER   INTERMEDIATE INFEASIBLE      0.025   -336.924644266471   F
```

For a run without infeasibilities, it prints:
```
Infeasibilities extracted from full.lst with nashstat -F:
no infeasibilities found
```

On my local machine, it prints:
```
Infeasibilities extracted from full.lst with nashstat -F:
Error: nashstat not found, search for p80_repy in full.lst yourself.
```

This first gives you an overview whether there were INFES during the run, but also helps newbies to find this useful tool.

### DIFF
--- a/scripts/output/single/emulator.R
+++ b/scripts/output/single/emulator.R
@@ -99,13 +99,13 @@ regions <- sort(getRegions(x$supplycurve))
 for (y in years) {
 
   title <- paste0(y) 
-  dat <- gginput(x$supplycurve[regions,y,],scatter="type")
+  dat <- gginput(x$supplycurve[regions,y,], scatter = "type", verbose = FALSE)
   dat$year<-factor(dat$year)
 
   p <- ggplot(dat, aes(x=.value.x,y=.value.y)) +
     geom_line(aes(colour=scenario, linetype=curve)) + #geom_line(size=0.5) + 
-    geom_point(data=gginput(x$rem_point[regions,y,],scatter = "variable"),aes(x=.value.x,y=.value.y,colour=scenario)) +
-    geom_point(data=gginput(x$mag_point[regions,y,],scatter = "variable"),aes(x=.value.x,y=.value.y,colour=scenario),shape=5) +
+    geom_point(data=gginput(x$rem_point[regions,y,],scatter = "variable", verbose = FALSE),aes(x=.value.x,y=.value.y,colour=scenario)) +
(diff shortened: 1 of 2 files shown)


Renato-Rodrigues:
Thank you for implementing those changes Oliver! They look fine for me too.
Maybe it would be nice in the future to have utilities functions like in nashstat or rs2 inside a package, so we can have those tests and reports carried even if we run in our local machines.

LaviniaBaumstark (approved):

giannou:
> have utilities functions like in nashstat or rs2 inside a package

`rs2` is already part of a package. See point 5 in the [documentation](https://pik-piam.r-universe.dev/ui#view:modelstats/rs2.html)

PR #713, opened 2022-03-24
Only set initial values for cumulated capacities if cm_startyear is t0

chroetz:
Capacities for some technologies in 2005 are set to a certain fixed value. This may be in conflict with loaded values from `input_ref.gdx` if `cm_startyear` is greater than 2005, see below. The issue is fixed by adding a condition to the equation for setting these initial values.

One the one hand,`q_capCumNet` sets initial (at time `t0`, which is 2005) values for cumulated capacities `vm_capCum` of learning technologies to a parameter value `pm_data(regi,"ccap0",teLearn)`. `pm_data(regi,"ccap0",teLearn)` is set to  `1/card(regi)*fm_dataglob("ccap0",te)` and `fm_dataglob` is read from `./core/input/generisdata_tech.prn`.

One the other hand, if `cm_startyear > t0`, `vm_capCum` in `t0` may be fixed to a certain value via `input_ref.gdx` and `fixings.gms`. 

If these values differ, `q_capCumNet` is infeasible.

**Example:** Run a baseline (`cm_startyear = 2005`) with `CDR = "off"` and a policy run (`cm_startyear = 2025`) with `CDR = "dac"`. Then, in the result of the baseline, `vm_capCum(2005, regi, "dac")` is 0, as CDR is off. Thus, in the policy run `vm_capCum(2005, regi, "dac")` is also fixed to 0, 2005 is before `cm_startyear` and this value is loaded from the baseline. But `pm_data(regi,"ccap0","dac")` is `1e-6 / 12 = 8.3333e-8`, as this is the respective value listed in `./core/input/generisdata_tech.prn`. Hence, `q_capCumNet(2005, regi, "dac")` is infeasible, as 0 is not equal 8.3333e-8.

I have tested the solution with the CDR module and "dac" and it works. But this change also affects other technologies. So a careful review and comments are appreciated.


LaviniaBaumstark:
In general, I like the solution. I only wonder if there are scenarios where we would like to have exactly the same settings for reference and policy scenario and where we now would not realize the difference? For dac I think the solution is fine. @robertpietzcker @Renato-Rodrigues do you see any problems for other technologies?

LaviniaBaumstark (approved):

Renato-Rodrigues:
I cannot see any issue regarding the change, although I think it kind of hides a issue instead of fixing it.

Either if you run cm_startyear = 2005 or 2025, right now the vm_capCum(2005, regi, "dac") value is being set to the parameter value in the equation. In principle, parameters shouldn't change for years not executed by the model. If we do that, we can create a mess with for example relative policy cost variables that require calculations involving both the baseline and policy run results.

So when you run the model with cm_startyear = 2025, in theory the 2005 values should be fixed for everything, variables and parameters, and the equation should be perfectly compatible with this all times.

If the parameter values are being changed only in the policy run case (e.g. the parameter is changed in ttot if DACs is enabled), I would instead make sure that the parameter does not ever change values for the fixed years (ttot le cm_startyear), and instead only apply change to its values in free years (use set t instead of ttot for any value updates).

chroetz:
@Renato-Rodrigues, @LaviniaBaumstark thanks for the comments. I looked into the reason for this problem some more and have an alternative proposal for a solution. Maybe this fits better to @Renato-Rodrigues's suggestion(?)

`pm_data` is [set](https://github.com/remindmodel/remind/blob/64adce06464fde4091ddb031992d75d84c28e87e/core/datainput.gms#L1072) for `te`. But `q_capCumNet` is [defined](https://github.com/remindmodel/remind/blob/ea141860a408488cfd131c5da1197fa7e64a0cd6/core/equations.gms#L371) for `teLearn`. `teLearn` always [contains](https://github.com/remindmodel/remind/blob/ec1bb3294d714572a8a84e1ad296e7bbfb74da59/core/sets.gms#L1315) `dac`. `te` [does not contain](https://github.com/remindmodel/remind/blob/ec1bb3294d714572a8a84e1ad296e7bbfb74da59/core/sets.gms#L1104) `dac` by default. Only if the `CDR` module is set to [`"DAC"`](https://github.com/remindmodel/remind/blob/ec1bb3294d714572a8a84e1ad296e7bbfb74da59/modules/33_CDR/DAC/sets.gms#L34) or to [`"all"`](https://github.com/remindmodel/remind/blob/ec1bb3294d714572a8a84e1ad296e7bbfb74da59/modules/33_CDR/all/sets.gms#L32) is `dac` included into `te`.

If `dac` is part of `te`, then `vm_capCum(2005, regi, "dac")` is required to be 8.3333e-8.
If `dac` is not part of `te`, then `pm_data(regi,"ccap0","dac")` is not defined. I'm not sure what happens with `q_capCumNet(2005, regi, "dac")` in this case. For some reason `vm_capCum(2005, regi, "dac")` becomes 0.

Instead of adding the `cm_startyear` condition to `q_capCumNet`, we could also make sure that `dac` is in `teLearn` if and only if it is in `te`.  Would that be a better solution?

LaviniaBaumstark:
maybe we should always add `dac` to `te` and set the value for `2005` (This are the historic data?). If `dac` is `off` we can turn it off after `2005`?

chroetz:
After discussion with @LaviniaBaumstark, we decided to merge the preliminary fix and later discuss with @strefler about the possibility yo add `dac` to `te`. See #728

PR #721, opened 2022-03-29
CDR realizations

katarkow:
Enables using different CDR module realizations between the base and policy runs, e.g., switching from `all` to `DAC` only.
 
Variables/equations that are unknown symbols when switching between realizations are handled in `levs.gms`, `fixings.gms` and/or `margs.gms`. Additionally, I used this opportunity to clean up the variables that are not used in a given realization (e.g.  `v33_grindrock_onfield` declared in DAC). I think they were added there so reporting doesn't crash and the values from the base run were set to zero anyways.  

Handling `v33_grindrock_onfield` in the reporting is [here](https://github.com/pik-piam/remind2/blob/a7ccbd0f572b5656bb40828c5840a505e51a3ec0/R/reportFE.R#L1281) and `v33_emiEW` and `v33_emiDAC` [here](https://github.com/pik-piam/remind2/blob/cec0792efd7b9acfdefbd29fbe4a31f02c4fdff8/R/reportEmi.R#L157), so it shouldn't create any problems.

Switching between realizations caused infeasibility in `q_capCumNet`, which was addressed in #713.

To consider: Supposedly we're switching from DAC to weathering, then we won't be able to access variables in DAC from t < cm_startyear. However, the core variables copied from the base run do account for DAC.

### DIFF
--- a/modules/33_CDR/DAC/bounds.gms
+++ b/modules/33_CDR/DAC/bounds.gms
@@ -8,9 +8,6 @@
 vm_emiCdr.fx(t,regi,emi)$(not sameas(emi,"co2")) = 0.0;
 vm_emiCdr.l(t,regi,"co2")$(t.val gt 2020 AND cm_ccapturescen ne 2) = -sm_eps;
 vm_omcosts_cdr.fx(t,regi) = 0.0;
-v33_emiEW.fx(t,regi) = 0.0;
-v33_grindrock_onfield.fx(t,regi,rlf,rlf2) = 0;
-v33_grindrock_onfield_tot.fx(t,regi,rlf,rlf2) = 0;
 if (cm_emiscen ne 1,
     vm_cap.lo(t,regi,"dac",rlf)$(teNoTransform2rlf_dyn33("dac",rlf) AND (t.val ge max(2025,cm_startyear))) = 1e-7;  
 );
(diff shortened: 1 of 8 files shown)


chroetz:
When I go from `CDR="all"` to something else I still get that `q33_omcosts` ("calculates O&M costs for spreading ground rocks on fields") is unknown in `margs.gms`. This probably needs to be added to the manipulateThis variables

katarkow:
> When I go from `CDR="all"` to something else I still get that `q33_omcosts` ("calculates O&M costs for spreading ground rocks on fields") is unknown in `margs.gms`. This probably needs to be added to the manipulateThis variables

Oops, didn't realize the equation name in all doesn't match the one in weathering. I think it makes more sense to change its name in one of the realizations.

Thanks a lot for catching it!

chroetz (approved):
Seems to work now, thanks!
I did run into some infeasibilities when testing it, but they should be solved with #713

chroetz:
I just realized that [`manipulateFile()`](https://github.com/pik-piam/lucode2/blob/7537708ec99f4fea655fb0fc3f404ef0d1c798f9/R/manipulateFile.R#L25) uses regular expressions by default. Thus, either all `.` in the search patterns have to be escaped, or, perhaps more elegantly, `fixed=TRUE` needs to be added to the calls to `manipulateFile()` [here](https://github.com/katarkow/remind/blob/d8c5410f74dc1be60f41e38ac3bf29f20ae32e0f/scripts/start/prepare_and_run.R#L790).

chroetz:
Thanks! Looks good to me :)

chroetz (approved):

LaviniaBaumstark (approved):
thanks for also cleaning up the code. I assume that we need both the fix by Christof and this one right?

PR #725, opened 2022-04-04
Add fixings for the CCU module

katarkow:
Enables switching the CCU module from `on` in a base run to `off` in a policy run.


chroetz (approved):
Perfect, thanks!

PR #727, opened 2022-04-05
for coupled runs, enable all path_gdx... to work as subsequent runs

orichters:
I reworked start_bundle_coupled.R to integrate several of the features added to start.R a while ago:
- path_gdx_refpolicycost can be used. If it is not specified, fall back to path_gdx_ref. The coupled config overwrites the non-coupled config.
- an automated check of the column names is introduced, plus some outdated columns let the run fail
- all path_gdx... can be used as subsequent runs, so the next run is only started if all gdx are ready. It is implemented in the same way as in start.R that the Rdata files in the remind repository keep track of the files, and if through replacing a run name with the corresponding fulldata.gdx completes the list of files, the run is started
- A number of successful runs can be found at `/p/tmp/oliverr/remind/output`, for example `C_run6test-rem-1` (see `less /p/tmp/oliverr/remind/C_run6test.log` which successfully used the results of `C_run5test-rem-2` as input to all the gdxes. Note that it is a coupled run with all 12 regions, calling remind twice and magpie once, but running each remind run with only one iteration.
- There is a bit of redundancy now between start.R and start_coupled_bundle.R, but I think this should better be solved with the new remindStart package by @johanneskoch94

### DIFF
--- a/start_bundle_coupled.R
+++ b/start_bundle_coupled.R
@@ -4,6 +4,8 @@
 # |  AGPL-3.0, you are granted additional permissions described in the
 # |  REMIND License Exception, version 1.0 (see LICENSE file).
 # |  Contact: remind@pik-potsdam.de
+library(stringr)
+
 ########################################################################################################
 #################################  U S E R   S E T T I N G S ###########################################
 ########################################################################################################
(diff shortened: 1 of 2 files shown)


dklein-pik (approved):
Great work. Thank you!

LaviniaBaumstark (approved):
thanks, it's also great bringing coupled and stand alone scripts closer. I agree that we need to restructure and clean up all starting scripts and go into more R-packages.

PR #729, opened 2022-04-06
New default config

chroetz:
This is a draft for a proposal to replace default.cfg by defaultConfig.Rmd.
(Requires a change in the `gms` R package to execute https://github.com/chroetz/gms/tree/defaultCfg)

The new defaultConfig.Rmd (in the config folder) allows for better documentation of the parameters previously in default.cfg. It enables markdown for writing the description and a HTML output for viewing it. The output is styled by defaultConfig.css. It has a table of contents for a comprehensive overview over the available parameters. It allows grouping into sections to structure the list of parameters. Hyperlinks between different parameter descriptions enable more efficient navigation.

As defaultConfig.Rmd has a formal structure, it is machine readable. In particular, this allows to auto-generate the switches and flags parts of main.gms. Therefore, when adding a new switch or flag or changing a default value, only the defaultConfig.Rmd has to be changed (and not main.gms).


orichters:
Thanks, will have a look later. Just a short comment: Other places may need updates as well, for example https://github.com/pik-piam/remind2/search?q=default.cfg

main_template.gms
```diff
@@ -0,0 +1,247 @@
...
+***------------------------------------------------------------------------------------------------
+***------------------------------------------------------------------------------------------------
+*** WARNING *** WARNING *** WARNING *** WARNING *** WARNING *** WARNING ***
+***------------------------------------------------------------------------------------------------
+***                        START OF WARNING ZONE
+***------------------------------------------------------------------------------------------------
+***
+*** PLEASE DO NOT PERFORM ANY CHANGES IN THE WARNING ZONE! ALL SETTINGS WILL BE AUTOMATICALLY
+*** SET BY submit.R BASED ON THE SETTINGS OF THE CORRESPONDING CFG FILE
+*** PLEASE DO ALL SETTINGS IN THE CORRESPONDING CFG FILE (e.g. config/default.cfg)
```
orichters, line 135: change filename at the end :)

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
> The new defaultConfig.Rmd (in the config folder) allows for better documentation of the parameters previously in default.cfg. 

All GAMS parameters in `default.cfg` are in `main.gms`, too, which can be documented using [goxygen syntax](https://github.com/pik-piam/goxygen).  Other can be added just for documentation.  We would just need to do that.

> It enables markdown for writing the description and a HTML output for viewing it. The output is styled by defaultConfig.css. It has a table of contents for a comprehensive overview over the available parameters. It allows grouping into sections to structure the list of parameters. Hyperlinks between different parameter descriptions enable more efficient navigation.

Is there a need for sections and cross links?  I suspect most people look for something specific and are fine with using the search feature of whatever software they are using.  This is a whole additional layer of complexity to be supported.

> As defaultConfig.Rmd has a formal structure, it is machine readable. In particular, this allows to auto-generate the switches and flags parts of main.gms. Therefore, when adding a new switch or flag or changing a default value, only the defaultConfig.Rmd has to be changed (and not main.gms).

The same is true for `default.cfg`.  It's just an R script generating a nested list, which can be written out as a series of GAMS statements.

Currently this can't be checked, since `readCfgFromRmd.R` is not sourced everywhere it needs to be and there are still references to `default.cfg`.  But if you go ahead with this, please make sure that `main.gms` remains in a state for `/p/projects/rd3mod/tools/gamscompile` to run.  Development turnover would suffer quite awfully otherwise.

> Unfortunately, the length of the file increases from 1145 to 5781 lines, which is quite substantial and will maybe reduce the fun editing this file.

I would say the file is positively unwieldy.  For example, the `CES_calibration` switches form a logical unit and are often modified and reviewed together.
```
$ grep -n CES_calibration config/default_old.cfg 
401:cfg$gms$c_CES_calibration_iterations         <-  10       # def <- 10  number of calibration iterations
402:cfg$gms$c_CES_calibration_new_structure      <-   0       # def <-  0, switch to 1 if you want to calibrate a CES structure different from input gdx
403:cfg$gms$c_CES_calibration_write_prices       <-   0       # def <-  0, switch to 1 if you want to generate price file, you can use this as new p29_cesdata_price.cs4r price input file
404:cfg$gms$cm_CES_calibration_default_prices    <-   0.01    # def <-  0.01 lower value if input factors get negative shares (xi), CES prices in the first calibration iteration
405:cfg$gms$c_CES_calibration_industry_FE_target <-   1
```
Six consecutive lines
```
$ grep -n CES_calibration config/defaultConfig.Rmd 
4351:### c_CES_calibration_iterations {-}
4356:cfg$gms$c_CES_calibration_iterations <- 10
4370:### c_CES_calibration_iteration {-}
4375:cfg$gms$c_CES_calibration_iteration <- 1
4408:### c_CES_calibration_new_structure {-}
4413:cfg$gms$c_CES_calibration_new_structure <- 0
4428:### c_CES_calibration_write_prices {-}
4433:cfg$gms$c_CES_calibration_write_prices <- 0
4448:### cm_CES_calibration_default_prices {-}
4453:cfg$gms$cm_CES_calibration_default_prices <- 0.01
4467:### c_CES_calibration_industry_FE_target {-}
4472:cfg$gms$c_CES_calibration_industry_FE_target <- 1
```
turned into 121 lines, which would be at least twice the size of the normal terminal window.
The configuration is much more often edited than studied for reference, I would argue.  Its structure should reflect that.

chroetz:
@orichters, @0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q Thanks a lot for the feedback, this is much appreciated!

Note that this is currently just a draft and a proposal and if it is agreed that this should eventually be merged there are a lot of things to be done before actually merging it.

> Unfortunately, the length of the file increases from 1145 to 5781 lines, which is quite substantial and will maybe reduce the fun editing this file.

True, that is a drawback. But the unstructured 1145 lines are also not particular fun to edit. Using folding and navigation by the structure induced by the sections might bring some fun back.

> Will the main-template.gms allow the situation that starting the model, the template is not touched? It sometimes annoys me that running REMIND, main.gms is changed by the process, blurring the changes I introduces to main.gms that should be committed with those automatically done by replacing the config parameters. If this could be separated this way, this would be very much appreciated!

The main-template.gms is not changed when starting a model. It is used to create main.gms from it (and the defaultConfig.Rmd).

> All GAMS parameters in default.cfg are in main.gms, too, which can be documented using [goxygen syntax](https://github.com/pik-piam/goxygen). Other can be added just for documentation. We would just need to do that.

One part of the motivation for the new defaultConfig is that introducing or changing a switch currently requires to edit main.gms and default.cfg at several places and partially just copy-pasting stuff. It would be better if we had all in one place and then create everything else automatically. The proposal in this pull request is centered around a default config file with R-code and Rmarkdown from which main.gms can be created. One could instead use main.gms with goxygen as the root and create a default.cfg from it (one would have to deal with the section of R parameters somehow). I think this would be a good alternative.

> Is there a need for sections and cross links? I suspect most people look for something specific and are fine with using the search feature of whatever software they are using. This is a whole additional layer of complexity to be supported.

I think links are an improvement (not strictly necessary, though). Eg, the description of `45_carbonprice` references many other parameters, to which you can navigate easily with the links. I have added the HTML-output `config/defaultConfig.html` if you want to take a look at. 
It is not difficult to support. You can even automatically add the links as you just have to replace mentions of a known variable `variableName` by `[variableName]`. This will create the link.

> The same is true for default.cfg. It's just an R script generating a nested list, which can be written out as a series of GAMS statements.

You can get the names of the variables form the `cfg` object and the values (potentially except for `cfg$repositories <- getOption("remind_repos")` but thats not a GAMS parameter). You cannot infer from the `cfg`-object whether the parameter is a switch or a flag. Furthermore, you cannot extract the description. But for automatically generating main.gms you need a one-line description. Currently, I don't see how to generate main.gms from default.cfg.
The new defaultConfig.Rmd would define a parameter as switch or flag by putting it in the right section. As each subsubsection of defaultConfig.Rmd has the structure title - one line description - R-code-chunk to define its default - further description, one can parse the file and extract the one-line description and whether something is a switch or a flag.  

> Currently this can't be checked, since readCfgFromRmd.R is not sourced everywhere it needs to be and there are still references to default.cfg. But if you go ahead with this, please make sure that main.gms remains in a state for /p/projects/rd3mod/tools/gamscompile to run. Development turnover would suffer quite awfully otherwise.

Thanks for noting that. I will make the corrections. Eventually the functions in script/config/ should be put into a package, probably `gms`. I will check gamscompile before even thinking of merging this :)

> I would say the file is positively unwieldy. For example, the CES_calibration switches form a logical unit and are often modified and reviewed together.

It would be possible to put all theses variables into one subsubsection with one R-code chunk. Then it can be 6 lines again.

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
> One part of the motivation for the new defaultConfig is that introducing or changing a switch currently requires to edit main.gms and default.cfg at several places and partially just copy-pasting stuff. It would be better if we had all in one place and then create everything else automatically. The proposal in this pull request is centered around a default config file with R-code and Rmarkdown from which main.gms can be created. One could instead use main.gms with goxygen as the root and create a default.cfg from it (one would have to deal with the section of R parameters somehow). I think this would be a good alternative.

Which made me think: we have a `default.cfg` because the original use case was to have different configuration files to be passed as a parameter to `start.R`.  But this was superseded by the scenario config files.  Nobody uses .cfg files anymore to run REMIND.  
How do I know?  Because they don't work anymore.  
```
$ ./start.R config/default.cfg 

Attaching package: ‘dplyr’

The following objects are masked from ‘package:stats’:

    filter, lag

The following objects are masked from ‘package:base’:

    intersect, setdiff, setequal, union

Loading required package: stringr

Reading config file config/default.cfg 
~/PIK/Remind (develop)$ 
```
Maybe since the latest changes to `start.R`, maybe longer.

So we might as well do away with them (it).

Have start functions modify `main.gms` with scenario config settings (if provided) and run straight `main.gms` otherwise.  Document it properly and add the non-GAMS stuff (input data revision, region setting, etc.) in special comment fields.

Escalating to @LaviniaBaumstark as patron saint of all things REMIND.

PR #730, opened 2022-04-06
Update NGFS scenario config

orichters:
what I did:
- Edge-T scenarios: `Mix3` for h_ndc, h_cpol and d_delfrag, `Mix4` for o_1p5c, o_2c, d_rap; delete entries for runs using `complex`
- join all runs from the normal config file and the `_damages` file
- harmonize cm_DiscRateScen, c_shBioTrans
- specify start groups for bIT and damage which can be conveniently started with the new interactive start mode
- cleanup of coupled document
- in 46_carbonpriceRegi, `p46_emi_actual_iter` was designed to track `p46_emi_actual` (and the last forced push fixed that p46_emi_actual does not depend on t)


christophbertram (approved):
Looks good. Now that we have set c_shBioTrans consistently to 1 (in line with default), it would be great if both of us remember to check for bioliquid shares in transport after we have run new runs... just to be sure.

PR #733, opened 2022-04-06
add assertion of qm_energy_limits for the calibration

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:


silviamade (approved):

LaviniaBaumstark (approved):
I am not happy about the extra treatment of industry and the new interfaces between industry and calibration but if it solves calibration problems I like it

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
It's either special treatment of industry in CES_parameters, or special treatment of CES_parameters in industry, no way around it.  But logically one would look for this stuff in the calibration.

PR #737, opened 2022-04-07
for MAgPIE coupling, enable to set cm_nash_autoconvergence differently

orichters:
- for MAgPIE coupling, enable to set cm_nash_autoconvergence differently for last REMIND run
- harmonize use of flags in start.R and start_bundle_coupled.R: Now, both support both --test and -t for the test mode. For backward compatibility, `Rscript start_bundle_coupled.R test` still works
- allow to specify scenario_config_coupled file as command line argument and simplify editing assuming that the REMIND config is the coupled config without the `_coupled`.
- rewrite tutorial, delete filenames that do not exist anymore


start_coupled.R
```diff
@@ -149,6 +149,11 @@ start_coupled <- function(path_remind,path_magpie,cfg_rem,cfg_mag,runname,max_it
       cfg_rem$output <- cfg_rem_original
     }
 
+    # change precision only for last run if setup in coupled config
+    if (i < max_iterations && ! is.null(cfg_rem$cm_nash_autoconverge_lastrun)) {
```
dklein-pik, line 153: I'm sorry if I misunderstand but shouldn't it be `if (i == max_iterations ...` ?

orichters: No, you are completely right. That was a really stupid mistake :) Thanks for spotting it, I fixed it.

Test runs with `max_iterations = 1` are here, check the config.Rdata:
- `/p/tmp/oliverr/remind/output/C_run7test-rem-1`: `cfg$cm_nash_autoconverge_lastrun` is `0` as specified in `scenario_config_coupled_test.csv`. `cfg$gms$cm_nash_autoconverge` is 0. Good!
- `/p/tmp/oliverr/remind/output/C_run8test-rem-1`: `cfg$cm_nash_autoconverge_lastrun` is `2` as specified in `scenario_config_coupled_test.csv`. `cfg$gms$cm_nash_autoconverge` is 2, differently to what is specified in `scenario_config_test.csv`. Good!
- `/p/tmp/oliverr/remind/output/C_run9test-rem-1`: `cfg$cm_nash_autoconverge_lastrun` is null because the column was renamed in `scenario_config_coupled_test.csv`. `cfg$gms$cm_nash_autoconverge` remains 0 as specified in `scenario_config_test.csv`. Good!

I killed the runs after the config file was written, it seems to work as intended.

PR #738, opened 2022-04-07
description must be strictly limited to 255 characters to avoid GAMS fails

orichters:
- (also resetting main.gms to default)

### DIFF
--- a/main.gms
+++ b/main.gms
@@ -83,9 +83,9 @@
 * 
 * Regionscode: 62eff8f7
 * 
-* Input data revision: 6.284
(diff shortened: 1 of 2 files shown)


LaviniaBaumstark (approved):
thanks!

orichters:
@LaviniaBaumstark: well, we worked a bit simultaneously apparently :). With a fresh morning mind, I had a different idea that keeps the full description for compareScenario2 etc., but only trunkates the one in GAMS.

LaviniaBaumstark (approved):
ure even better

PR #740, opened 2022-04-08
first main.gms relief: update tmp file instead

orichters:
Sitting in boring DB trains, I thought about a simple relief for the main.gms update issue.

What the script does:
- to construct `tmpmodelfile`, take `cfg$model` and add `cfg$title` before the `.gms`, so: `main.gms` -> `main_SSP2-Base.gms`.
- copy `cfg$model` to `tmpmodelfile` and do all the substitutions etc. there
- once model unlocked, move `tmpmodelfile` as `main.gms` to the output folder. Filename may not always be "correct" as `cfg$model` may not be `main.gms`. One could also simply delete the file, it is not needed afterwards anyway.
- `main.gms` in REMIND folder remains unchanged by the scripts
- only if `cfg$title` is `default`, it does the operations on the actual `main.gms`, so the default settings can be updated – then, `main.gms` will be copied, not moved to the output folder.

Does not solve the rest the updating problem completely, but I think it is a simple fix as a first relief. I may not be fully aware of side effects of this (standalone runs, etc.), so please check carefully or tell me what else to test.

- A run where it worked but that I killed after REMIND starting running is `/p/tmp/oliverr/remind/output/SSP2-Base_bIT_2022-04-08_19.36.59`.
- A default run that updated `c_description`  in `main.gms` according to my edit in `default.cfg` is `/p/tmp/oliverr/remind/output/default_2022-04-08_19.38.53`


0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
Tagging remindmodel/development_issues#100

> only if `cfg$title` is default, it does the operations on the actual `main.gms`, so the default settings can be updated – then, `main.gms` will be copied, not moved to the output folder.
  
Leaves a way to use `gamscompile`. :+1:

`main.gms` being updated never bothered me as much personally.  But hey, there seems to be demand.

johanneskoch94:
This made me think: one advantage of using a package is that using "tests" would make it a lot easier to, well, test ;)

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
> This made me think: one advantage of using a package is that using "tests" would make it a lot easier to, well, test ;)

Provided there are meaningful unit tests that can be defined for the problem at hand.

orichters:
I use this script successfully also for the NGFS coupled runs. It works nicely, though I'm not sure whether its necessary to store the `main.gms` file in the output folder, could also simply be deleted.

LaviniaBaumstark:
looks like a good solution. I agree that we do not need to copy main.gms to the output folder. The only thing I am wondering is whether we should add the time stamp to the main_*.gms for several scenarios with the same name?

orichters:
In this setup, a timestamp is not needed, because the model will still be locked, then the tmp file is created and then moved away to the output folder, so there can never be two files with the same name. And I would either move the tmpmodel to the output folder, or delete it, but in any case not leave it in the main folder.

I only think we may change the name to `main_tmpmodelfile_SSP2-Base` or so to make it less probable that the setup overwrites some `main_Base.gms` file or so somebody has lingering around in the folder.

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
Question: is this going to be _fait accompli_, determining outcomes of the May 3rd meeting?  Because depending on decisions taken then, this might be moot.  Or decisions might not be taken, to preserve this.

Issue #741, opened 2022-04-11
#705 breaks REMIND for local use


0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
at least for those lowly individuals among us not using slurm on our laptops.
```
~/PIK/Remind ((HEAD detached at 808a3b9b))$ ./start.R --testOneRegi

Attaching package: ‘dplyr’

The following objects are masked from ‘package:stats’:

    filter, lag

The following objects are masked from ‘package:base’:

    intersect, setdiff, setequal, union

Loading required package: stringr

All flags: --testOneRegi

Which region should testOneRegi use? Type it, or leave empty to keep settings:
Examples are CAZ, CHA, EUR, IND, JPN, LAM, MEA, NEU, OAS, REF, SSA, USA.
EUR

 testOneRegi 
   Writing cfg to file testOneRegi.RData
   Creating results folder output/testOneRegi 
    Deleting results folder because it already exists: output/testOneRegi 
   Writing cfg to file output/testOneRegi/config.Rdata 
   Executing prepare_and_run.R for output/testOneRegi 
sh: 1: sbatch: not found
Warning message:
In system(paste0("sbatch --job-name=", cfg$title, " --output=log.txt --mail-type=END --comment=REMIND --wrap=\"Rscript prepare_and_run.R \" ",  :
  error in running command
```
```
~/PIK/Remind ((HEAD detached at 808a3b9b))$ ./start.R 

Attaching package: ‘dplyr’

The following objects are masked from ‘package:stats’:

    filter, lag

The following objects are masked from ‘package:base’:

    intersect, setdiff, setequal, union

Loading required package: stringr

All flags: 

 default 
   Writing cfg to file default.RData
   Creating results folder output/default_2022-04-11_11.05.09 
   Writing cfg to file output/default_2022-04-11_11.05.09/config.Rdata 
   Executing prepare_and_run.R for output/default_2022-04-11_11.05.09 
sh: 1: sbatch: not found
Warning message:
In system(paste0("sbatch --job-name=", cfg$title, " --output=log.txt --mail-type=END --comment=REMIND --wrap=\"Rscript prepare_and_run.R \" ",  :
  error in running command
```
[...]

PR #742, opened 2022-04-11
reenable direct option for testOneRegi and output, fixing 741

orichters:
fixing issue #741.
- make sure that some slurmConfig is always specified
- add direct mode in output.R
- ask for testOneRegi region only if runs are restarted, where no easy option to specify / change it exists.


0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q (approved):
Seems to work. :+1:

PR #747, opened 2022-04-12
Bugfixes in the CDR module

katarkow:
Fixing a few bugs in the CDR module: 
- `fedie.cdr` was missing in `entyFe2Sector`, which led to wrong computations in `reportFE`,
- `q33_grindrock_onfield_tot`: missing one step for timesteps that are even numbers,
- in `all`, computations in `q33_capconst_dac` and `q33_capconst_grindrock` summed over the whole mapping `teNoTransform2rlf_dyn33`, which resulted in capacity * number of all elements in that mapping, instead of only the ones for a given technology,
- `q33_DacFEdemand_heat`: to compute DAC heat demand, all CDR emissions were used instead of DAC emissions only,
- in the `DAC` realization, DAC technology is added to `teAdj` and in realization `all` it is not.

### DIFF
--- a/core/sets.gms
+++ b/core/sets.gms
@@ -1988,6 +1988,7 @@ entyFe2Sector(all_enty,emi_sectors) "final energy (stationary and transportation
 		fehes.cdr
                 fegas.cdr
                 feh2s.cdr
+                fedie.cdr
 /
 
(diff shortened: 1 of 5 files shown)


orichters:
What is the state of this PR? Are these substantial bugs or just minor issues? Thank you

amerfort (approved):
Looks good! Thanks for fixing these.

PR #754, opened 2022-04-14
Added switch for aborting after consecutive fails.

PhilippVerpoort:
The creation of this new switch (called `cm_abortOnConsecFail`) was discussed in the REMIND channel. When set to a
non-zero integer value N, it will cause REMIND to abort its run in Nash mode after at least one region failed to solve N times consecutively.

The default value is zero for the moment, which means it has to be actively turned on to have any effect.

I successfully tested the behaviour of this new switch. Most importantly, it doesn't affect any runs when set to zero, hence ensuring back-compatibility and avoiding unexpected behaviour for users of the develop branch.

### DIFF
--- a/config/default.cfg
+++ b/config/default.cfg
@@ -294,6 +294,7 @@ cfg$gms$codePerformance <- "off"        # def = off
 #                          SWITCHES
 #-----------------------------------------------------------------------------
 cfg$gms$cm_iteration_max       <- 1       # def <- 1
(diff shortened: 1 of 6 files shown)


Renato-Rodrigues (approved):

PhilippVerpoort:
A new PR for setting the default value of this new switch to 5 can now be found here:

https://github.com/remindmodel/remind/pull/928

PR #755, opened 2022-04-19
Default edge transport scenario fixed.

MariannaR:


orichters (approved):
Thanks!

orichters:
Would be great if this line was adapted twice as well: https://github.com/remindmodel/remind/blob/develop/main.gms#L503

orichters:
Thanks!

giannou:
shall we update/correct the `scenario_config_###` files here or open a new PR?

MariannaR:
I am always cautious with `scenario_config_###` files, as people might be using them for their projects and would be on them to define which option is representing their scenario. What is your take on this? Shall we substitute all of them?

orichters:
I was told by Lavinia that if I change REMIND code, I should correct all scenario config files in the config folder such that they continue to work. And you and the other transport people probably know best how to substitute a specific scenario, for example `ElecEraWise` -> `Mix4`.

Here is how I did it with renaming NDC2018 to NDC: https://github.com/remindmodel/remind/pull/581

So if you think all `ElecEraWise` should be `Mix4` from now on, go to the folder and run:
```
grep ";ElecEraWise;" scenario_confi*.csv
```
and if you think all the places found should be replaced, continue with:
```
sed -i "s/;ElecEraWise;/;Mix4;/g" scenario_confi*.csv
```
In the best case, the PR also contains information how to update other config files that people may use but have not merged them into the develop branch. If there is no clear 1:1 relation between old and new scenarios, it becomes a bit more difficult, I understand, then maybe you have to check who was editing the scenario config files last and inform them they have to decide that for themselves.

MariannaR:
it is not so straightforward from the perspective of the scenario implications - the previous ones were "technology-based" (e.g. a battery electric world, a H2 world...) while the current scenarios are "budget-based" (e.g. a stringent policy world versus a more conservative world). I will tentatively match the old and new scenarios: 

`ConvCase`, `ConvCaseWise`->`Mix1`
`ElecEra`, `HydrHype`, `ElecEraWise`, `HydrHypeWise` -> `Mix3`

do the other transport experts @Loisel and @johannah-pik  agree?

orichters:
From my perspective, it would be perfectly fine to replace them with something that is "quite similar", and maybe provide an explanation in the next REMIND meeting why you did what you did so people can adjust later.

MariannaR:
The last commit includes the scenario config renaming.

PR #761, opened 2022-04-20
update lower bound to vm_cesIO(201[05],SSA,enhgab) to accomodate low c_CES_calibration_default_prices

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
- if `c_CES_calibration_new_structure` is set, all prices are set to
  `c_CES_calibration_default_prices` (0.01)
- since the demand for `fegab` and `feh2b` in `SSA` is so low, the
  resulting value (quantity &times; price) of `enhgab` is lower than the
  default lower bound on `vm_cesIO` of 1e-6
- this applies only to the first calibration iteration, since "normal"
  prices are a magnitude larger
- `enhgab` inputs are fixed in the first calibration iteration to target
  values, to facilitate the CONOPT solution process

### DIFF
--- a/modules/01_macro/singleSectorGr/bounds.gms
+++ b/modules/01_macro/singleSectorGr/bounds.gms
@@ -11,6 +11,19 @@ vm_cons.lo(t,regi)     = 1e-3;
 vm_cesIO.lo(t,regi,in)$( NOT in_industry_dyn37(in) ) = 1e-6;
 vm_cesIOdelta.lo(t,regi,in_putty) = 1e-6;
 
+* 201[05]/SSA/enhgab seems to need lower bounds
+$ifthen.calibration "%CES_parameters%" == "calibrate"   !! CES_parameters
+$ifthen.first_CES_itr "%c_CES_calibration_iteration%" == "1"   !! c_CES_calibration_iteration
+loop ((t,regi)$( t.val ge 2010 AND t.val le 2015 AND sameas(regi,"SSA") ),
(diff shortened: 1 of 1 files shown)


LaviniaBaumstark (changes requested):
you should use a more general formulation for "sameas(regi,"SSA")"

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
> you should use a more general formulation for "sameas(regi,"SSA")"

No.  This applies only to `SSA`, in either H12 or EU21 configurations.  It should be specific, not general.

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
OK, this merge request is moot, since this appeared in a second calibration iteration, too.  See [here and following](https://mattermost.pik-potsdam.de/rd3/pl/d814j5xgcpbu3ebseya8t5m58o).

The culprit in this case seems to be a surprisingly low `feh2b` price returned from the ESM in the first calibration iteration:
```
$ dumpgdx non_optimal.gdx p29_CESderivative "2010,SSA,inco,feh2b"
2010  SSA  inco  feh2b  0.0091652306420884
```

So I will go for a general solution along the lines of Renato-Rodrigues/remind#40.

Not sure how this performs in baseline/policy runs, though.  We should consider adopting more differentiated bounds on `vm_cesIO`, as we will be running into more and more of these problems with more detailed regions and CES structures.

PR #762, opened 2022-04-20
update exoGAINS/postsolve to use input.gdx in first NASH iteration

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
- exoGAINS calls the script exoGAINSAirpollutants.R, which calls
  remind2::reportFE, which uses vm_otherFEdemand, which might not be
  available during the first NASH iteration
- therefore, the input.gdx is copied to fulldata.gdx before
  exoGAINSAirpollutants.R is called
- fulldata.gdx is subsequently overwritten in core/loop

### DIFF
--- a/modules/11_aerosols/exoGAINS/postsolve.gms
+++ b/modules/11_aerosols/exoGAINS/postsolve.gms
@@ -11,12 +11,28 @@
 *--------------------------------------------------------------------------
 *** write the fulldata.gdx file after each optimal iteration
 ***AJS* in Nash status 7 is considered optimal in that respect (see definition of o_modelstat in solve.gms)
-if((o_modelstat le 2),
+if (iteration.val eq 1,
+  !! in the first iteration, use the input.gdx, since some output data is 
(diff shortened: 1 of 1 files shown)


LaviniaBaumstark (commented):
I think, Rscript exoGAINSAirpollutants.R failed because of a reporting error.  When I tried it, copying input.gdx did not work. Are you sure that your code works?

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
> When I tried it, copying input.gdx did not work. 

Where?  How did it fail?

> Are you sure that your code works?

Worked here `/p/tmp/pehl/Remind/output/SSP2EU-calibrate_2022-04-20_09.28.06/full_01.log` and here `/p/tmp/pehl/Remind/output/EU21reg-SSP2EU_calibrate_2022-04-20_08.49.28/full_01.log`, among others.

But seems to have been addressed in pik-piam/remind2#231.  Scuttling.

PR #763, opened 2022-04-20
add modification of ppf prices to ensure labour xi >= 0.1

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
- If the sum of the products of target quantities and calibration prices
  of all `ppf` less labour exceed `inco`, the calibration would assign
  a negative price to labour to even the Euler identity on the root CES
  level, resulting in an aborted calibration.
- Therefore it is checked that labour has an income share of at least
  10 %, and all other `ppf` prices are scaled down as needed to ensure
  that.
- Theoretically, this should not occur in later calibration iterations.
  But for now, the test for that is deactivated to observe the
  calibration behaviour.

### DIFF
--- a/modules/29_CES_parameters/calibrate/preloop.gms
+++ b/modules/29_CES_parameters/calibrate/preloop.gms
@@ -114,9 +114,7 @@ p29_CESderivative(t,regi_dyn29(regi),out,ppf_29(in2))$(
 p29_cesdata_load(t,regi_dyn29(regi),"inco","price") = 1;   !! unit price
 
 *** Transfer prices
-
-pm_cesdata(t,regi_dyn29(regi), in, "price") =
- p29_CESderivative(t,regi,"inco",in);
+pm_cesdata(t,regi_dyn29(regi),in,"price") = p29_CESderivative(t,regi,"inco",in);
 
 option
   p29_CESderivative:3:3:1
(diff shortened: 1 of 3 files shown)


modules/29_CES_parameters/calibrate/preloop.gms
```diff
@@ -257,6 +257,35 @@ if (sm_tmp eq 1,
...
+$ontext
+  if (%c_CES_calibration_iteration% gt 3,   !! c_CES_calibration_iteration
+    execute_unload "abort.gdx";
+    abort "Too low labour xi after third CES iteration.  CALIBRATION PANIC!";
+  );
+$offtext
+
+  display "after ppf price multiplicator",
+    p29_ppf_price_multiplicator, pm_cesdata;
+);
```
LaviniaBaumstark, line 288: is there a mechanism to stop this price_multiplicator? How to ensure that other consistency checks do not fail because of the price shift?

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q: > is there a mechanism to stop this price_multiplicator?

See
> Theoretically, this should not occur in later calibration iterations.
> But for now, the test for that is deactivated to observe the
> calibration behaviour.

and [lines 279-84](
https://github.com/remindmodel/remind/pull/763/files/ce3ae4b5caafe7b99087a66069626ef3697ab60c#diff-54a5960804a16adc77b7ddc96e32be89e442abbd1f89b0fa5cd8e7c5a1b174b7R279-R284)
```
$ontext
  if (%c_CES_calibration_iteration% gt 3,   !! c_CES_calibration_iteration
    execute_unload "abort.gdx";
    abort "Too low labour xi after third CES iteration.  CALIBRATION PANIC!";
  );
$offtext
```

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q: > How to ensure that other consistency checks do not fail because of the price shift?

Prices are modified directly after derivation, so all calculations the consistency checks are checking are based on these modified prices.

LaviniaBaumstark: ah, you plan to activate lines 279-84, thanks

LaviniaBaumstark (approved):

LaviniaBaumstark:
what is the status of this PR? will you continue working on it?

PR #768, opened 2022-04-22
fixing bug for equations incompatible with policy runs

Renato-Rodrigues:
This PR fixes the incompatibility of recent introduced equations with model runs that set cm_startyear greater than 2005


orichters (approved):
Thanks, Renato!

PR #772, opened 2022-04-23
updating CES and GDX version for latest calibration files

Renato-Rodrigues:
updating CES and GDX version for pack including the latest SSP2EU H12 and EU21 calibrations files.


orichters (approved):
we are already using these files for our H12 SSP2-Base NGFS run

PR #773, opened 2022-04-24
allow not to set cm_nash_autoconverge_lastrun for some runs

orichters:
- bugfix on #737
- make the logging more readable


christophbertram (approved):
Looks good, thanks also for taking care of nicer helper displays...

PR #774, opened 2022-04-25
adding the coal policy switch to bounds preventing coal to gas switch for DEU

cchrisgong:
apparently the gas bounds was activated also under no coal phaseout policies.

### DIFF
--- a/modules/47_regipol/regiCarbonPrice/bounds.gms
+++ b/modules/47_regipol/regiCarbonPrice/bounds.gms
@@ -84,7 +84,7 @@ vm_emiTeDetail.up(t,regi,peFos,entySe,teFosCCS,"cco2")$((sameas(regi,"DEU")) AND
 *** limit German CDR amount (Energy system BECCS, DACCS, EW and negative Landuse Change emissions), conversion from MtCO2 to GtC
 vm_emiCdrAll.up(t,regi)$((cm_deuCDRmax ge 0) AND (sameas(regi,"DEU"))) = cm_deuCDRmax / 1000 / sm_c_2_co2;
 
-
+$IFTHEN.CoalRegiPol not "%cm_CoalRegiPol%" == "off"
 *** adaptation of power system for Germany in early years  to prevent coal to gas switch in Germany due to coal-phase out policies
 loop(regi$(sameAs(regi,"DEU")),
 vm_deltaCap.up("2015",regi,"ngcc","1") = 0.002;
@@ -93,5 +93,6 @@ vm_deltaCap.up("2025",regi,"ngcc","1") = 0.0015;
 *** limit early retirement of coal power in Germany in 2020s to avoid extremly fast phase-out
 vm_capEarlyReti.up('2025',regi,'pc') = 0.65; 
 );
+$ENDIF.CoalRegiPol
 
 *** EOF ./modules/47_regipol/regiCarbonPrice/bounds.gms


Renato-Rodrigues:
Felix was the one that added the code. You should include him as a reviewer of the PR.
Also, without digging deeper, these affect historical years so at least part of it should be by default in all runs, so hiding behind the switch wouldn't be my approach to the situation.

I would instead: 
(1) split the historical years behavior related bounds from the policy years. 
(2) Keep the historical bounds active at any time.
(3) Hide the policy years bounds under the switch.
(4) make the switch on by default as this should represent legislation that is already at place and it should be enforced anyway.

cchrisgong:
> Felix was the one that added the code. You should include him as a reviewer of the PR. Also, without digging deeper, these affect historical years so at least part of it should be by default in all runs, so hiding behind the switch wouldn't be my approach to the situation.

Felix is on parental leave, might not be able to respond to this. But I tagged him just in case. I'm also not very familiar with this implementation, all I know was that it was done for Ariadne....

> Keep the historical bounds active at any time.

I agree the historical bounds should be on by default, provided they are not there to accompany non-historical bounds (for example if without the historical bounds, one get infes; i.e. the bounds in itself is not based in real-world observations, they are there for model behavioral reasons). Without Felix's input, it is hard to know which one this is

> make the switch on by default as this should represent legislation that is already at place and it should be enforced anyway.

I think there could be reason to make even legislation in place non-default since 
1) legislations can be reverted. For example due to the energy crisis and Russia-Ukrain conflict right now, coal plants could see longer time under baseline case. By adding a switch, modellers can make a conscious decision to check whether this policy is still in place; 
2) if the definition for "baseline" is "no additional policies", then baseline should include policies that are in place, I agree, but in most occasions we also don't have such policies or goals in place: for example, [DEU's nuclear phaseout is also behind a switch](https://github.com/remindmodel/remind/blob/develop/modules/47_regipol/regiCarbonPrice/bounds.gms#L16), as are goals for [EU's electrolysers, China's SOE goals for nuclear addition (both are not implemented) ](https://github.com/remindmodel/development_issues/issues/69), goals for EVs (or even current sales) are not implemented in EDGE-T, etc. It would be nice to have a group meeting/task force to systematically create a guideline for what type of policies should be on under "baseline", what is not.

cchrisgong:
On the "which time-span of which policy should be under a switch" question, here is a matrix I made to organize this better, what do you think?:
image

cchrisgong:
Thanks for the comments. I think we could think this way:

> (1) split the historical years behavior related bounds from the policy years.

not sure the deltacap("2015") is quite historical bound, since it is the capacity addition for "2015", hence impacting "2020", which is not historcal? (but I'm not sure)

> (4) make the switch on by default as this should represent legislation that is already at place and it should be enforced anyway.

Should this be on under baseline? I think in that case then we also should have ETS on under baseline. Either baseline is no policy, or no additional policy

I think we can just do it like I originally proposed. Splitting out the historical part doesn't follow what seems to be Felix's logic, i.e. line [85](https://github.com/remindmodel/remind/blob/develop/modules/47_regipol/regiCarbonPrice/bounds.gms#L88)-95 belong to one block. Nor should it be on by default, as that would also turn it on under baseline... (and updating all baselines in config tables to have it off is a hassle)

Renato-Rodrigues:
> not sure the deltacap("2015") is quite historical bound, since it is the capacity addition for "2015", hence impacting "2020", which is not historcal? (but I'm not sure)

2020 is a historical year. Its numbers should reflect 2018-2022 values, and the majority of it is already history.

> Should this be on under baseline? I think in that case then we also should have ETS on under baseline. Either baseline is no policy, or no additional policy

The baseline does not change what happened in the past. Both these and the ETS historical CO2 prices are considered also under the baseline for historical years, e.g.: https://github.com/remindmodel/remind/blob/7d1a32acf79e071a3b5008287f5bbdb00527c614/modules/47_regipol/regiCarbonPrice/postsolve.gms#L26

for later years, it is a matter of the scenario definition, and it is not related with the code change you proposed here.

I would request to remove this PR and move this discussion to a issue instead. If you want to keep the PR change it to a draft instead, as your proposal for changes is not desirable, as historical coal behavior should be detached from policy decisions from the future anyway.

fschreyer:
This change to historic capacity additions is active in baseline and policy runs. It is also in the regipol ``none`` [realization](https://github.com/remindmodel/remind/blob/c99ee0c4f08235a437c34bfe7403a0b48c995805/modules/47_regipol/none/bounds.gms#L32). I added it in regipol ``none`` and ``regiCarbonPrice`` to have them always on. The reason why these changes are in ``regipol`` is because Lavinia does not want to have code with hard-coded regions outside of the ``regipol`` module. I agree, this is a rather confusing solution.

PR #777, opened 2022-04-26
Bugfix to exchange of GHG prices in coupling. Remove deprecated "LU_pricing" column

dklein-pik:


start_coupled.R
```diff
@@ -82,6 +82,8 @@ start_coupled <- function(path_remind,path_magpie,cfg_rem,cfg_mag,runname,max_it
   cfg_mag$sequential <- TRUE
   cfg_mag$force_replace <- TRUE
   cfg_mag$output     <- c("rds_report") # ,"remind","report") # rds_report: MAgPIE4; remind,report: MAgPIE3 (glo.modelstat.csv)
+  # if provided use ghg prices for land (MAgPIE) from a different REMIND run than the one MAgPIE runs coupled to
+  use_external_ghgprices <- if(is.na(cfg_mag$path_to_report_ghgprices), FALSE, TRUE)
```
orichters: This shouldn't work this way, it should be `ifelse`

orichters (approved):
ok, after the if -> ifelse it looks good. Thanks for fixing it.

PR #779, opened 2022-04-26
New dist heat scen

robinhasse:
This PR adds an extra district heating scenario to the ARIADNE config. The new scenario together with the existing Ariadne scenarios will be used in the ECEMF T2.2 on buildings.


pweigmann:
It is a bit hard to see, but I assume only the addition of line 16 is the change committed here?

robinhasse:
> It is a bit hard to see, but I assume only the addition of line 16 is the change committed here?

Exactly. I Added a description to the scenario in a new column which adds semicolons to the end of each line. That's why all of them are changed. Apart from that, all lines are identical.

pweigmann (approved):

PR #787, opened 2022-04-28
report correct run names for testOneRegi

giannou:
make starting a testOneRegi run by editing default.cfg report correct names, use inclusive language


orichters (approved):
Thanks for fixing that! But as far as I know, this was "always wrong" and not introduced by me recently, or did I fail again?

giannou:
Indeed not your fault, it has been broken for a very long time

Issue #728, opened 2022-04-06
CDR off to CDR dac is solved preliminary


chroetz:
See discussion in #713

Issue #771, opened 2022-04-23
CDR = off doesn't work in NDC scenarios


orichters:
We were trying to run `45_carbonprice = NDC` runs with `CDR = off`, but this creates infeasibilities where `vm_deltaCap` for `dac` goes to 1e+10 for some regions and timesteps. There seem to be bounds missing somewhere? @christophbertram, can you provide a bit more information?

christophbertram:
Yes, the issue is that unlimited consumption is possible, as regions seem to benefit via `SubsidizeLearning` from installing the maximum possible deltaCap of "dac" in 2020 and 2025. Presumably, this is without cost, which is why it pays off to do so... These regions then become `UNBOUNDED`...

robertpietzcker:
If I understood correctly, dac is missing from te2rlf. 
If that is the case, the model sees no costs, which is dangerous. 
if a technology is in te, it should be in te2rlf, or it has to forbidden by setting both cap and deltacap to 0 (as is done for solhe, for example)

orichters:
I don't think this is the case, because the `DAC` realization of `CDR` works without problems. The `CDR = off` realization must implement the `off` somehow, that is it.

robertpietzcker:
if you check in "DAC" realization, you will see that it is added in "sets.gms" to te2rlf.
you have to add DAC to te2rlf also in the off case, otherwise the model has a loophole, and it will very likely use that loophole when it least expects it. 
we do this for other tech that are not used in the model but exist in te exactly the same way ("solhe")

katarkow:
The unlimited consumption was a consequence of `dac` always being part of `teLearn` used in `SubsidizeLearning`, while, if CDR is off, it is not in `te` nor `te2rlf`.  After speaking to Jessica, we agreed that `dac` can be added to `teLearn` only when it is expected to be deployed like in the case of `te` and `te2rlf`. See #789.

PR #789, opened 2022-04-29
Add DAC to teLearn from the CDR module

katarkow:
Fix to #771 - installing maximum deltaCap of DAC, when the CDR module is off.

After discussing with Jessica, we agreed that `dac` can be added to `teLearn` from the CDR module only when DAC is used. This way, when there is no DAC, it will not be accounted for in `SubsidizeLearning`.

Additionally, this could fix issue #728, which was introduced in #713 - I removed the condition in `q_capCumNet`, since if `dac` is not in `teLearn`, we'll not run into infeasibilities.

### DIFF
--- a/core/equations.gms
+++ b/core/equations.gms
@@ -368,7 +368,7 @@ qm_deltaCapCumNet(ttot,regi,teLearn)$(ord(ttot) lt card(ttot) AND pm_ttot_val(tt
 ***---------------------------------------------------------------------------
 *' Initial values for cumulated capacities (learning technologies only):
 ***---------------------------------------------------------------------------
-q_capCumNet(t0,regi,teLearn)$(cm_startyear LE t0.val)..
+q_capCumNet(t0,regi,teLearn)..
   vm_capCum(t0,regi,teLearn)
   =e=
   pm_data(regi,"ccap0",teLearn);
--- a/core/sets.gms
+++ b/core/sets.gms
@@ -1312,7 +1312,6 @@ $IFTHEN.WindOff %cm_wind_offshore% == "1"
         storwindoff "storage technology for wind offshore"
 $ENDIF.WindOff
         storcsp     "storage technology for csp"
-        dac         "direct air capture"
         elh2        "hydrogen elecrolysis"
 /
 
--- a/modules/33_CDR/DAC/sets.gms
+++ b/modules/33_CDR/DAC/sets.gms
@@ -26,12 +26,18 @@ adjte_dyn33(all_te)           "technologies with linearly growing constraint on 
 /
       dac
 /
+
+teLearn_dyn33(all_te)         "learning technologies"
+/
+      dac
+/
 ;
 
 ***-------------------------------------------------------------------------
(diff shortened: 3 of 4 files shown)


orichters:
Sorry, I have to admit that I don't understand enough to approve this PR, I neither worked with the TE specifications nor CDR.

But thanks for your work, I appreciate it again!

chroetz:
If I understand it correctly, after this change, `dac` is in `te` if and only if `dac` is in `teLearn` if and only if `CDR` is either `"all"` or `"dac"`.

`dac` is still always a [part](https://github.com/remindmodel/remind/blob/8eb0bab5c18525d4236e21ec01bcddcb292789c1/core/sets.gms#L275) of `all_te` and `vm_capCum` is [defined](https://github.com/remindmodel/remind/blob/ec1bb3294d714572a8a84e1ad296e7bbfb74da59/core/declarations.gms#L335) on `all_te`. Is this a problem? In particular, does `vm_capCum(2005, regi, "dac")` come up in the fixings of a policy run, e.g.,  with `CDR="dac"`, when the baseline run has `CDR="off"`?

katarkow:
@chroetz To your first question: Variables in gams cannot be declared over dynamic sets e.g., `te`, so for all the variables that are declared over technologies, `all_te` is used. [See documentation](https://www.gams.com/latest/docs/UG_DynamicSets.html#UG_DynamicSets_EquationsDefinedOverTheDomainOfDynamicSets). 

As for the second question: If there is no CDR and `dac` is not in `teLearn` in the baseline run, then (I think) `vm_capCum(ttot, regi, 'dac')` is copied from the input file and then it is not touched in the main code. These copied values would be then used for a policy run with e.g., `DAC` or some other realization. I'm looking into this now. Thanks for the comment! :)

LaviniaBaumstark:
regarding the question of DAC in base vs. policy scenarios: this has two aspects: first it might lead to compilation errors if the file with the fixing information contains sets which are not declared in the policy scenarios (seem not to be a problem here) and second we might get infes if the amount of DAC in the first time steps is inconsistent to the usage in later time periods

chroetz (approved):
I tested all combinations of realizations of the CDR module for one iteration and all worked without an error.

I think, it is not desirable that DAC capacity values are copied from the input.gdx if `CDR="off"` is chosen (instead of constraining it to some level so that it is independent of the input.gdx). But this is a different issue.

orichters:
I think fixing DAC capacity values to zero for `CDR = off` should be done. I see no sense in using the data from the base run at this point. @chroetz: is this a different issue in its colloquial meaning, or is there actually a remindmodel issue somewhere?

chroetz:
@LaviniaBaumstark Yes, it works!
@orichters Now there is remindmodel issue (#816)

LaviniaBaumstark:
sorry, was nut aware that you were waiting for me

katarkow:
Thanks for all the comments and @chroetz for all the testing! In this case, I'll merge this PR and continue with #816

PR #790, opened 2022-05-02
add automatic switch to Nash debug mode for testing

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:


robertpietzcker (approved):
It seems in principle great, I just think the "cleaning up formulations" by replacing all ' with " went a little bit to far and changed some don't to don"t - which might be a problem in gams statements that are terminated by "

display "#### We can"t accept this solution, because it is non-optimal, and, in addition, too far away from the last known optimal solution. ";

it would be great to quickly remove the don"t / doesn"t instances

PhilippVerpoort (approved):

PR #791, opened 2022-05-02
allow non-existing columns for cm_nash_autoconvergence_lastrun

orichters:
check more precisely whether cm_nash_autoconvergence_lastrun is defined


merfort (approved):
Thanks for the quick fix!
Looks all good to me.

PR #792, opened 2022-05-02
update EDGE-T scenarios in automated model tests (AMT)

giannou:
The values for the `Mix#` setting were chosen so that they correspond to the ones found in the file `config/scenario_config_rev3p0p0.csv`. Are they correct?


LaviniaBaumstark:
more or less, some are still preliminary

LaviniaBaumstark (approved):

giannou:
@pfuehrlich-pik I guess you can run tests now

PR #793, opened 2022-05-02
get rid of magpie-dependency and use magpie4 instead

LaviniaBaumstark:


dklein-pik (approved):

pfuehrlich-pik (changes requested):
`magpie` is still needed, see [here](https://github.com/LaviniaBaumstark/remind/blob/c7981bcfe542100142431b68fcab1c3653e45a57/scripts/output/comparison/CO2price.R#L13) and [here](https://github.com/LaviniaBaumstark/remind/blob/c7981bcfe542100142431b68fcab1c3653e45a57/scripts/output/comparison/oneRegComp.R#L9)

pfuehrlich-pik:
You can check if magpie is still needed with `x <- renv::dependencies(); x[x["Package"] == "magpie",]`. Also `EDGETransportComparison.Rmd` is broken, there are two lines which have a superfluous `)`, see warning generated by `renv::dependencies()`

LaviniaBaumstark:
I know that there is still magpie used at some places. Those are currently discussed and hopefully will be solved soon. I will add it to this PR.

pfuehrlich-pik (approved):

PR #794, opened 2022-05-02
Parallel mode for coupled runs

orichters:
This PR introduces the parallel mode for coupled runs. Using the flag `--parallel` or `-p` for `Rscript start_bundle_coupled.R` allows to speed up the total calculation time substantially.

Imagine you want to run a `Base` run, a `NDC` run and a `Policy` run. NDC uses `Base` as `path_gdx_ref` and `path_gdx_bau`. `Policy` uses `Base` as `path_gdx_bau`  and `NDC` as `path_gdx_ref`. The idea is to run the coupling with `5` REMIND runs and `4` MAgPIE runs.

In non-parallel mode, the current line is like that:
```
Base-rem-1 -> Base-rem-2…4 -> Base-rem-5 -> NDC-rem-1 -> NDC-rem-2…4 -> NDC-rem-5 -> Policy-rem-1 etc.
```
This is a nuisance, because running `rem-1` until `rem-5` requires ~36 hours which either impedes you from using the `priority` slurm or forces you to regularly restart something. So a bundle of subsequent runs with depth 4 requires easily a week.

This PR enables to change it as such:
```
Base-rem-1 -> Base-rem-2   -> Base-rem-3   -> Base-rem-4
            ↳ NDC-rem-1    -↳ NDC-rem-2    -↳ NDC-rem-3    -↳ NDC-rem-4
                            ↳ Policy-rem-1 -↳ Policy-rem-2 -↳ Policy-rem-3 -↳ Policy-rem-4
```
`NDC-rem-1` starts already after `Base-rem-1` is finished, and `NDC-rem-2` waits for `Base-rem-2` as `path_gdx_ref` and `path_gdx_bau`, and `NDC-rem-1` as `path_gdx`.

The advantage is:
- substantial speed-up of the overall process, because many of the runs can be calculated simultaneously, not serially. If a single REMIND + MAgPIE run takes 6 hours, you do 5 REMIND-MAgPIE iterations, and you have a chain of four runs, the current setup takes 5 * 4 * 6 = 120 hours = 5 days. With the new setup, each is (5 + 3) * 6 hours = 48 hours = 2 days.
- each REMIND-MAgPIE run is a slurm batch thus `rem-1 + mag-1`, `rem-2 + mag-2`, each taking less than 24 hours, so priority/short slots can be used.
- For each REMIND run (including the subsequent MAgPIE run), a single log file in `output` is written, making everything a bit more tracktable

I did some tests with a REMIND and MAgPIE "emulator" essentially consisting of a "model" file containing `execute 'cp ../fulldata.gdx fulldata.gdx';` to avoid tedious waiting times, and ran a complete set of NGFS runs with this setup, fixing several bugs in the process, in particular for restarting the bundle at different states.

the changes in detail
----
- replace `read_all` from deprecated `magpie` package by `readAll` from `remind2` as [discussed here](https://github.com/pik-piam/discussions/discussions/14)
- in the usual mode, you have a single `RData` file in the REMIND folder (such as `C_SSP2-Base.RData`) and then `start_coupled.R` iterates through `rem-1` until `rem-5`, at the end starting the subsequent runs (policy run, for example)
- in parallel mode, you have one `RData` file for each REMIND run (such as `C_SSP2-Base-rem-1.RData`), and `start_coupled.R` executes REMIND and MAgPIE only once, then starting new runs (including its successor `C_SSP2-Base-rem-2`). This allows also to edit the `RData` files after starting the runs, if `qos` or other settings are to be adapted.
- in `serial` mode, `start_bundle.R` loops through the REMIND and MAgPIE iterations and at the end, starts the subsequent runs. In `parallel` mode, `start_bundle.R` executes only one single run for REMIND and MAgPIE each (Say: `rem-1` and `mag-1`), and then directly starts the subsequent runs which includes the `rem-2` and `mag-2` run.
- the logs are moved into the REMIND output folder: `output/log_scenario-rem-3.txt` logs `rem-3` and `mag-3`. If at some point only `mag-3` is restarted, to avoid overwriting it, save its log as `log_scenario-mag-3.txt`.
- Add `errorsfound`, `startedRuns`, `waitingRuns` and `knownRefRuns` to provide a clearer summary of which runs are started and waiting (also fixing the issue that the script told it was waiting for runs that weren't actually started)
- Runs to be waiting for are displayed in blue in the start_bundle_coupled.R logging
- add PIK cluster specific warning on priority slots
- search for finished runs first in the current REMIND and MAgPIE folders and try to continue them. Only if nothing is found there, go to the oldruns folder. This makes sure that the `rem-x` within the output folder are consistent.
- minor correction of a stated "not clean, but sufficiently clean" solution `nr_of_regions` (actually: number of CPUs) for subsequent runs. Doesn't use the number of regions of the current run as before, but of the run to be started
- after the last REMIND run of each scenario is finished, a compareScenario2 is started comparing all `rem-*`.
- tutorial updated

Ideas for enhancement were moved to issue #834.


dklein-pik (approved):
Great, thank you!

gabriel-abrahao (approved):

merfort:
Thanks a lot for the great work! Looks all good to me.

I did some test comparing coupled REMIND-MAgPIE runs in `serial` and `parallel` mode and results indicate that the policy runs in `rem-5` do not differ significantly between `serial` and `parallel` mode, see here:
```
/p/tmp/merfort/NAVIGATE_industry_MIP_comp_v2_v3/remind/output
```
For my tests REMIND and MAgPIE did not fully converge after 5 iterations (both for `serial` and `parallel` mode), but the convergence progress was the same.

orichters:
We have new code currently that is not extremely well checked in the sense of "people reading it", but we had @abrahao, @dklein, @merfort and myself using it to successfully run different bundles of scenarios, and several special cases and problems were identified and fixed within this process. We will probably identify further problems at some point (probably because the runs failed in a new situation), but I'm quite confident these can be either solved quickly, or mitigated by just manually deleting the output folder with the semi-finished run and restarting then (which works reliably).

As we have to rework all the scripts anyway in the future with the new start scripts, and the current setup is broken (read_all stuff), I decided to merge it now.

PR #796, opened 2022-05-03
Changes in capacity trade realisation

PhilippVerpoort:
Merging back several new recent developments in the `capacity` realization of the `24_trade` module.


LaviniaBaumstark:
you renamed some sets. Did you test that the reporting and all scripts run between iterations are still working?

PhilippVerpoort:
The changes I made to the sets are all in relation to the new trade realisation, which are not referenced by any reporting or intermediate scripts yet, so this shouldn't cause any problems.

Nonetheless, I just completed two REMIND runs that compare the develop branch against this PR. They seem to work fine, and I would like to run a compareScenarios2 next to demonstrate that my PR doesn't introduce any unexpected changes.

I am however currently struggling with some other reporting issue, which is apparently unrelated to the commits I made (i.e. occurs even when cloning the develop branch). I'll investigate and will let you know.

PhilippVerpoort:
I now successfully compiled compareScenarios2, which shows no difference between the develop branch and my branch containing the changes in this PR.

I therefore believe it is safe to merge.

modules/23_capitalMarket/debt_limit/equations.gms
```diff
@@ -23,7 +22,6 @@ q23_limit_surplus_growth(t,regi)..
   + sum(tradePe, (pm_pvp(t,tradePe)/(pm_pvp(t,"good")+0.000000001))*(vm_Mport(t,regi,tradePe)- vm_Xport(t,regi,tradePe))) 
   + (pm_pvp(t,"perm")/(pm_pvp(t,"good")+0.000000001)) * (vm_Mport(t,regi,"perm") - vm_Xport(t,regi,"perm"))
   + sum(tradeSe, pm_MPortsPrice(t,regi,tradeSe) * vm_Mport(t,regi,tradeSe)) - sum(tradeSe, pm_XPortsPrice(t,regi,tradeSe) * vm_Xport(t,regi,tradeSe)) 
-  + vm_budgetTradeM(t,regi) - vm_budgetTradeX(t,regi)
```
LaviniaBaumstark, line 26: did you discuss this with Marian?

modules/80_optimization/nash/equations.gms
```diff
@@ -13,7 +13,7 @@ q80_budg_intertemp(regi)..
   + SUM(ttot$(ttot.val ge 2005),
      pm_ts(ttot)
       * (
-        SUM(trade$(NOT tradeSe(trade)),
+        SUM(trade$(NOT tradeSe(trade) and NOT tradeCap(trade)),
```
LaviniaBaumstark, line 16: is it correct that the new tradeCap-set is not included in the Nash convergence?

modules/80_optimization/testOneRegi/equations.gms
```diff
@@ -13,14 +13,17 @@ q80_budg_intertemp(regi)..
 0 =e=  
  SUM(ttot$(ttot.val ge 2005), 
     pm_ts(ttot) 
-       * SUM(trade$(NOT tradeSe(trade)), 
+       * (
+         SUM(trade$(NOT tradeSe(trade)),
```
LaviniaBaumstark, line 17: why do you use here tradeSe and not tradeCap?

LaviniaBaumstark (commented):
a few questions

modules/23_capitalMarket/debt_limit/equations.gms
PhilippVerpoort: These are budget terms I previously added when I first merged my initial draft of the capacity realisation. I didn't discuss adding them or removing them now with anyone, because so far they are all set to zero in all realisations except my new one.

In other words: this is always zero in `standard` trade, so it doesn't matter.

modules/80_optimization/nash/equations.gms
PhilippVerpoort: First of all, it doesn't matter for anyone except me, as the `tradeCap` set is empty in all realisations other than `capacity`. So any normal user working with `standard` trade will be unaffected by this change.

As an explanation why this is removed: it did not have to be there in the first place, because currently the `trade` set (which is calculated automatically by REMIND) simply doesn't contain `tradeCap`. The question whether in the long run `tradeCap` should be part of `trade` is one that needs to be determined. Renato opted to include `tradeSe` in `trade` but then exclude it explicitly in a few places. I have so far not done this but instead explicitly added `tradeCap` in any sum where it was needed.

modules/80_optimization/testOneRegi/equations.gms
PhilippVerpoort: First, I didn't change this, this specific line is obviously just adding a bracket. The actual change is further below, where I'm adding new terms.

Second, as explained above, Renato chose to add `tradeSe` to `trade` so he had to exclude it at that point. `tradeCap` doesn't need to be included at this point, there should be further terms below this line.

LaviniaBaumstark (approved):

orichters:
The renaming of set elements unfortunately breaks the fixing to runs some weeks old, see `less /p/tmp/oliverr/remind/output/d_delfrag_bIT_2022-05-19_12.10.46/full.lst`. Can this be also fixed [somewhere here](https://github.com/remindmodel/remind/blob/develop/scripts/start/prepare_and_run.R#L527) or is it just impossible to reuse older runs?

orichters:
I'm now quite confident that renaming the set elements in `core` causes the problems. Why did you rename `gas_pipe` to `pipe_gas`? The affected variables are: `v_adjFactor.L`, `v_adjFactorGlob.L`, `v_costInvTeDir.L`, `vm_capDistr.L`, `vm_capFac.L`, `vm_deltaCap.L`, `vm_costTeCapital.L`, `vm_capEarlyReti.L`.

orichters:
I'm now quite confident that renaming the set elements in `core` causes the problems. Why did you rename `gas_pipe` to `pipe_gas`? The affected variables are: `v_adjFactor.L`, `v_adjFactorGlob.L`, `v_costInvTeDir.L`, `vm_capDistr.L`, `vm_capFac.L`, `vm_deltaCap.L`, `vm_costTeCapital.L`, `vm_capEarlyReti.L`.

PhilippVerpoort:
Hi @orichters,

thanks for bringing this up. It does make sense that the renaming of control set elements related to trade capacities caused your observed issue.

My apologies, I was unaware that this might cause trouble for people trying to reuse older runs.

It's unfortunate that — despite the modularisation of REMIND — I can't seem to make changes in my realisation of the code that nobody else uses without still causing trouble for others (since the technologies also need to be defined in core).

The easiest way to fix this (i.e. without trying to fix the old runs) would be to simply add the old trade technologies back into `all_te`. We could in principle add these back to the `develop` branch, but this effectively means we can never remove old/obsolete control set elements from core, which would results in unnecessary clutter in the long run. Perhaps an better way would be to add these locally for you specific runs. Therefore if you just add these set elements

```
gas_pipe
lng_liq
lng_gas
lng_ves
coal_ves
```

to `all_te` in core, it should work just fine. Maybe that's enough so you can complete whatever work you're doing with your old runs?

orichters:
Don't worry, I don't expect anyone to actually understand with precision the interplay of all these model elements.
I think the alternative would be to add to rename the technology names with `levs_manipulateThis`, `margs_` and `fixings` [as I did it here](https://github.com/remindmodel/remind/blob/develop/scripts/start/prepare_and_run.R#L777-L779), except not renaming the parameters, but the set elements. I don't know if we are allowed to do that, but I will try it out and let me know if it works.

PR #797, opened 2022-05-04
fix bugs in the damage modules

fpiontek:
There were still some bugs in some of the other damage modules from all the renaming which are fixed now.


orichters (approved):
Thanks, Franziska!

PR #800, opened 2022-05-04
adding switch control to edge-t non-energy transportation costs scaling

Renato-Rodrigues:
set cm_transpGDPscale to "off" to disable edge-t non-energy scaling


MariannaR (approved):
Thank you for including the switch! We could probably switch to default `off` once the new version of the `edgeTrpLib`  and `edgeTransport` libraries is merged into master and a new input data revision is available

Renato-Rodrigues:
Is the new input data revision already the default one?
If not, this should be switched to off when the input data revision is updated.
If yes, juts let me know and I will change it to off by default.

MariannaR:
No, this is not the case yet. I will take care of switch it to `off` when the data is available and the revision is included in REMIND as default.

PR #802, opened 2022-05-05
add a copy of the scenario config csv to the run folder

cchrisgong:
I've added this to the list of files to be copied to the output folder, so there is a record of what the REMIND cascade looked like in the run folder, and allow users to be able to compare scenario config from finished runs
(this will complement cfg.txt, which only merges the default.cfg and config.csv, but is agnostic about what were in the config csv) 

It seems that [many people want this feature](https://mattermost.pik-potsdam.de/rd3/pl/wx6nfs57gjnu9r3uddskkywese) in the group channel, but hopefully it won't cram up the run folder too much

### DIFF
--- a/start.R
+++ b/start.R
@@ -230,7 +230,7 @@ configure_cfg <- function(icfg, iscen, iscenarios, isettings) {
     names(gdxlist) <- path_gdx_list
 
     # add gdxlist to list of files2export
-    icfg$files2export$start <- c(icfg$files2export$start, gdxlist)
+    icfg$files2export$start <- c(icfg$files2export$start, gdxlist, config.file)
 
     # add table with information about runs that need the fulldata.gdx of the current run as input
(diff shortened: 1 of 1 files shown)


giannou (approved):
Very useful, thanks. Was the functionality tested, for e.g. missing `config.file`, coupled runs, etc.?

orichters:
Missing config.file shouldn't bother, because configure_cfg is only called if `! is.na(config.file)`. Cannot work for coupled runs like this because they don't call `start.R`. But I may add that to #794. But I hope @cchrisgong tested it a bit…

cchrisgong:
Thanks, both. I tested the no config.file base, so basically only with Rscript start.R which starts a default run. It looks okay to me, of course, no config table is copied. I tested a normal run case and checked that the csv is copied. Should I do more tests (what type of tests do you mean, @orichters )?

With regard to coupled run, I trust Oliver will make it work on his side - if I understand your comment correctly, Oli

orichters (approved):
That is fine…

PR #803, opened 2022-05-06
Safe files for extended reporting

johannah-pik:
Safe additional RDS files to be used in the new extended EDGE-T reporting that will be implemented soon.
Namely:
- logit_data.RDS
- logit_params.RDS


Loisel (approved):
Should be completely harmless.

Issue #785, opened 2022-04-28
Upgrade to remind2


johanneskoch94:
https://github.com/remindmodel/remind/blob/65b8a204109e9bea8b99585eea6835063020cd2c/scripts/output/comparison/policyCosts.R#L307
Here, `remind2` should be used instead of `remind`.

PR #807, opened 2022-05-10
update dependency list

pfuehrlich-pik:
I just ran `renv::dependencies()` to auto-detect dependencies. I manually added `goxygen` which is not used in R files, but it is mentioned in the readme to generate documentation. The full command I used before copy-pasting:
```
cat(paste0('"', sort(unique(renv::dependencies()$Package)), '"', collapse = ",\n"))
```


README.md
```diff
@@ -99,20 +99,46 @@ variable. After that you can run the following lines of code:
...
+          "lusweave",
+          "magclass",
+          "magpie4",
+          "mip",
+          "mrremind",
+          "mrvalidation",
+          "optparse",
+          "plotly",
+          "quitte",
+          "remind",
```
mikapfl, line 132: Do we still depend on remind as well as remind?

mikapfl:
There is another dependency list in https://github.com/remindmodel/remind/blob/develop/tutorials/1_GettingREMIND.md, probably it also needs updating? Or both need merging?

README.md
johanneskoch94: Nope, #785

pfuehrlich-pik: Did you already test the suggestion from the issue? Can I just add your suggestion to this PR?

johanneskoch94: I have not tested it, but I think it will work.

pfuehrlich-pik: Could you test it? I think this is much quicker than if I try to find out how to call the function now

johanneskoch94: I tested it, looks fine to me.

pfuehrlich-pik:
I replaced `remind::` with `remind2::` as suggested by Johannes, so the remind dependency is gone now

pfuehrlich-pik:
I think we can merge now, but I still need approval :wink:

johanneskoch94 (approved):

PR #808, opened 2022-05-10
Renv support

pfuehrlich-pik:
This PR adds a renv based workflow that should replace the snapshot machinery. This will only work with R > 4.0 (piam/1.24), because with R 3.6 renv crashes when packages should be installed that require R > 4.0.
- [x] There's a so-called "main renv" in the root folder of the remind repo
- [x] R package dependencies can be easily installed via a script -> run `Rscript setup.R`
- [x] Whenever a new runfolder is created, the main renv is updated and the main renv is then copied to the runfolder.
- [x] auto-updates can be disabled permanently in a non git-tracked way -> add `options(autoRenvUpdates = FALSE)` to your `~/.Rprofile`
- [x] remind AMTs run successfully with renv
- [x] `Rscript output.R` uses the main renv
- [x] `Rscript output.R --renv=X` uses the renv at path X
- [x] `Rscript output.R --update` updates the main renv before using it
- [x] renvs are stored in an archive and can easily be restored -> `scripts/utils/{archiveRenv,restoreRenv}.R`
- [x] snapshots can still be activated and work as before (the plan is to remove all snapshot related code soon though) -> rename `snapshot.Rprofile` to `.Rprofile`
- [x] renv workflows are properly documented and existing tutorials are updated -> see [tutorial](https://github.com/pfuehrlich-pik/remindmodel/blob/renv-support/tutorials/11_ManagingRenv.md)


0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
> Which packages (CRAN/pik packages) should be updated how (manually/automatically)?

None.  That's the whole point of `renv`, to have a stable package library that does not change without user interaction.

>  My inclination would be to update only pik packages automatically, 

So far, REMIND has only been trash-canned by updates of pik packages.  So those should definitely not be updated automatically.

> and only in the root renv and not in the renvs in the results folders.

The output directories should not automatically be tempered with on third count: for providing reproducible output.  We have to archive those for publication runs, nobody needs to the extra work to go back and check whether their state changed between run and archiving.

.Rprofile
```diff
@@ -1,45 +1,22 @@
...
+                    "mrdrivers", "mredgebuildings", "mrfable", "mrfeed", "mrfish",
+                    "mrland", "mrmagpie", "mrremind", "mrsoil", "mrtutorial", "mrvalidation",
+                    "mrvalidnitrogen", "mrwaste", "mrwater", "mstools", "piamModelTests",
+                    "piktests", "quitte", "regressionworlddata", "remind", "remind2",
+                    "remulator", "rmndt", "shinyresults", "trafficlight")
+
+
+# TODO should this be on or off by default?
+# comment out the following line to disable auto-updates
+renv::update(intersect(utils::installed.packages()[, "Package"], pikPiamPackages))
```
0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q, line 17: This line defeats the whole purpose of the commit.

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q (changes requested):

.Rprofile
pfuehrlich-pik: I'm quite sure this is not problematic unless multiple R sessions are running in parallel. Updating package x only breaks R when the update is installed while x is already loaded. When updating at the start of the session via .Rprofile no packages are loaded yet.

pfuehrlich-pik:
What do you think about a question whenever you start a remind run interactively? Something like "The following packages can be updated: x 1.2.3 -> x 1.3.4, y 1.2 -> y 1.5  --- update now? (y/N)"
This should prevent people from unknowingly using outdated packages.

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
> What do you think about a question whenever you start a remind run interactively? Something like "The following packages can be updated: x 1.2.3 -> x 1.3.4, y 1.2 -> y 1.5 --- update now? (y/N)" 

Personally I'm strongly opposed to the computer pestering me continuously with pointless questions.

> This should prevent people from unknowingly using outdated packages.

What is your definition of "outdated"?  The only way packages could be outdated is if people manually update REMIND code that depends on some new package version.  Which happens once in a blue moon that happens to also be a Friday.

.Rprofile
0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q: No.  I did not write the line does not work.  

> This PR replaces the snapshot machinery in .Rprofile with an renv based approach.

If the `renv` library gets updated every time R is started in this particular directory, that is the exact opposite of what the "snapshot machinery" should provide – stable package versions.

.Rprofile
```diff
@@ -1,45 +1,22 @@
...
-# to the model output folder. Otherwise, the run itself will again use the
-# default library set!
-
-# snapshot <- "/p/projects/rd3mod/R/libraries/snapshots/2022_03"
-# snapshot <- latest_snapshot
-
-if(exists("snapshot") && file.exists(snapshot)) {
-  message("libPaths was manually set to: ",snapshot)
-  .libPaths(snapshot)
-}
```
orichters, line 28: These lines should definitely stay in some way. I use both the option to fix to a specific snapshot or to use the latest to guarantee reproducibility and also to avoid errors in longer runs if somebody breaks something in between

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q: `renv` supersedes snapshots.  
Each REMIND directory has its own private R package library, that is constant and does not see unsupervised changes (pending changes discussed above).  Each REMIND output directory gets an independent copy of that library.
So "fixing to a specific snapshot" becomes just "not changing the `renv` package library".

orichters: Ok, thanks, this wasn't clear from the PR description. I also regularly set different REMIND repos to the same library snapshot for testing, so it would be great if this could easily be achieved.

orichters: And, sorry for asking as I haven't understood the renv approach: In NGFS, we ran REMIND with a certain set of libraries, and later had to rerun the reporting with a more recent library version. Will there still be an easy way to get older versions back once the libraries were updated within the repo?

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q: > I also regularly set different REMIND repos to the same library snapshot for testing, so it would be great if this could easily be achieved.

Should suffice to copy the `renv.lock` file from one directory to the other to make the target behave like the source.

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q: > In NGFS, we ran REMIND with a certain set of libraries, and later had to rerun the reporting with a more recent library version. Will there still be an easy way to get older versions back once the libraries were updated within the repo?

See [here](https://rstudio.github.io/renv/articles/renv.html#workflow).  You basically work with different versions of the `renv.lock` file, however you choose to do that (`git`, `copy/rename`, …).

pfuehrlich-pik: Every `renv` is represented by a simple text file called `renv.lock`. In order to get the exact same `renv` in different places you could just copy `renv.lock` and then `renv::restore` it. In the same way you can save an `renv.lock` file anywhere, maybe in a `renv_locks` folder in your home dir, and re-use them later by copying and restoring.

orichters: Thanks, Pascal. We have [this quite PIK cluster centered explanation](https://github.com/remindmodel/remind/blob/develop/tutorials/4_RunningREMINDandMAgPIE.md#create-snapshot-of-r-libraries) in the REMIND-MAgPIE-tutorial, which would then require an update, I guess.

pfuehrlich-pik:
We just discussed the remind renv integration in the RSE team, and here's the plan for now:
There's a so-called "main renv" in the root folder of the remind repo. Whenever a new runfolder is created, the main renv is updated (this can be disabled via `default.cfg`) and the main renv is then copied to the runfolder. For reportings (`Rscript output.R`): `single` uses the corresponding runfolder renv, `comparison` throws a warning and uses the runfolder renv of the first selected run (similar to how multiple historical.mif are handled). When running `Rscript output.R --update` the main renv is first updated and then used. Also, there will be a transition phase where the renv workflow is the default, but the snapshot workflow can still be used instead.
I'll also present this in the remind meeting on Thursday, and we can discuss this there.

orichters:
> For reportings (`Rscript output.R`): `single` uses the corresponding runfolder renv, `comparison` throws a warning and uses the runfolder renv of the first selected run (similar to how multiple historical.mif are handled). When running `Rscript output.R --update` the main renv is first updated and then used.

Thanks for the explanation. This means that if I want to rerun the reporting with updated libraries (which was my usual reason for changing libraries somewhere within the process), I would need to update the renv of all the individual runs?

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
> Whenever a new runfolder is created, the main renv is updated (this can be disabled via `default.cfg`)

So we have to keep `default.cfg` modified and get merge conflicts on every other pull in order to use `renv`?

pfuehrlich-pik:
> Thanks for the explanation. This means that if I want to rerun the reporting with updated libraries (which was my usual reason for changing libraries somewhere within the process), I would need to update the renv of all the individual runs?

No in that case you'd use `Rscript output.R --update` which uses the main renv (in the repo root, independent of individual runfolder renvs) after updating all packages there.

.Rprofile
pfuehrlich-pik: Thanks, I'll make sure it gets an update.

LaviniaBaumstark:
> > For reportings (`Rscript output.R`): `single` uses the corresponding runfolder renv, `comparison` throws a warning and uses the runfolder renv of the first selected run (similar to how multiple historical.mif are handled). When running `Rscript output.R --update` the main renv is first updated and then used.
> 
> Thanks for the explanation. This means that if I want to rerun the reporting with updated libraries (which was my usual reason for changing libraries somewhere within the process), I would need to update the renv of all the individual runs?

no, you would only run `Rscript output.R --update` which updates the Renv in the main remind folder and use this for the scripts you are running. All Renv files in the output folders would not be changed to ensure reporodicability

LaviniaBaumstark:
> > Whenever a new runfolder is created, the main renv is updated (this can be disabled via `default.cfg`)
> 
> So we have to keep `default.cfg` modified and get merge conflicts on every other pull in order to use `renv`?

For now, we only add the possibility to disable Renv updates via a set like all the other sets in REMIND are handled. For the future, this is part of the updated structure of the starting scripts, where we plan to only keep `main.gms` and handle all changes directly in the `full.gms` which will avoid any merge conflicts

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
> For the future, this is part of the updated structure of the starting scripts, where we plan to only keep `main.gms` and handle all changes directly in the `full.gms` which will avoid any merge conflicts

How so?  Users would still have to change a file under version control locally (whether `main.gms` or `default.cfg` is immaterial) to achieve desired behaviour, resulting in merge conflicts whenever the file is changed upstream.

LaviniaBaumstark:
'main.gms' would alwaiys contain the default settings and you would change it for your scenarios via `scenario_config_*.csv`

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
> 'main.gms' would alwaiys contain the default settings and you would change it for your scenarios via scenario_config_*.csv

Which presupposes that a scenario config file is used.  During development, that is often not the case.

The case remains that automatic updating of package libraries is contrary to the very intent of `renv` and the snapshot mechanism it is to replace.
Nowhere in the current `.Rprofile` is there any automated updating of the library snapshot.  Nobody asked for this functionality.  Why introduce it?

pfuehrlich-pik:
The goal is to create a workflow that allows both, stable package environments but also convenient access to package updates. Using outdated (defined as "an update is available") packages for longer periods of time is not recommended as this increases the chance of accumulating multiple overlapping incompatibilities with up-to-date packages that are then harder to debug. Also bugfixes are not used as soon as they are available. For this reason I'd argue that automatic package updates should in some way be enabled by default, but can be turned off. Default settings should work for most (and especially less experienced) users, even if that means they are less suitable for expert level users, as they can just change settings.

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
> The goal is to create a workflow that allows both, stable package environments but also convenient access to package updates. 

And this doesn't.  If users have to manually change some setting that is likely to be overwritten on the next pull to not have their package library changed without any interaction, that is not a stable package environment.

> Using outdated (defined as "an update is available") packages for longer periods of time is not recommended as this increases the chance of accumulating multiple overlapping incompatibilities with up-to-date packages that are then harder to debug. 

What is that sentence even meant to mean?  The `renv` library is supposed the provide the same functionality as the current library snapshots do.  Do you regularly go out and delete old snapshots because of "accumulating multiple overlapping incompatibilities with up-to-date packages"?
The package library is self-contained.  If it works, it works.  And will do so in the future.  If users selectively update only single packages which depend on updated versions of other packages in the library, that is their doing.  And the way to address this would be to include version requirements in dependencies, not force-updating everything.

> Also bugfixes are not used as soon as they are available. 

Same goes for bugfixes in the REMIND code.  You also wouldn't put `system('git pull --force')` in the `.Rprofile` so that users get "all the bugfixes", would you?  REMIND user are well-capable of keeping their working environment up to date.

> For this reason I'd argue that automatic package updates should in some way be enabled by default, but can be turned off. Default settings should work for most (and especially less experienced) users, even if that means they are less suitable for expert level users, as they can just change settings.

Two minimal condition on that:
1. A way to consistently opt out.  Either a environment variable for a global setting, or a marker file that is ignored by git for a per-repository setting.  Or both.
2. A way to roll changes back.  Package updates are as likely to be detrimental to REMIND runs as they are to improve them.  Users will need a way to undo automatic updates that turned out to be buggy.  Otherwise we end up in the same state of "nobody can work until some error way over there crashing everything is fixed" sooner or later again.

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
Come to think of if, I'm not convinced putting `renv.lock` under version control is really beneficial here.
- With automatic updates, it will be in a constant modified state, something we want to get rid of.
- With automatic updates, this is bound to result in constant merge conflicts, which people won't want to resolve manually.
- Who's responsible for updating it?

I'd argue to best have it ignored by git.

pfuehrlich-pik:
Having a way to disable auto-updates with a non-git-tracked file/environment variable should be possible.

> A way to roll changes back

Good idea. How would you implement that? After every update copy the renv.lock into a renv_history folder (ignored by git) and rename that renv.lock so it includes the date?

> I'd argue to best have it ignored by git.

I agree. For [releases](https://github.com/remindmodel/remind/releases) it might make sense to include the renv.lock though.

LaviniaBaumstark:
I never expected the renv.lock being part of git, but for releases this would be really a good idea

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
> > A way to roll changes back
> 
> Good idea. How would you implement that? After every update copy the renv.lock into a renv_history folder (ignored by git) and rename that renv.lock so it includes the date?

- `renv::snapshot()`
- save `renv.lock` with time stamp (not just date) to some ignored location
- `renv:update()`
- `renv::snapshot()`

Plus a script to choose some `renv.lock_time` file, move it back, and call `renv::restore()`

.Rprofile
```diff
@@ -1,5 +1,17 @@
 source("renv/activate.R")
 
+options(repos = c(pikruniverse = "https://pik-piam.r-universe.dev", CRAN = "https://cran.rstudio.com/"))
+
+# bootstrapping, will only run once after remind is freshly cloned
+if (identical(rownames(installed.packages(priority = "NA")), c(Package = "renv")) &&
```
0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q, line 7: Just wondering: what kind of error does that throw if the RSE package repository is not configured correctly?  Maybe capture that and give an explicit hint on how to do that.

output.R
```diff
@@ -21,6 +21,17 @@ options(error = quote({
   q()
 }))
 
+argv <- get0("argv", ifnotfound = commandArgs(trailingOnly = TRUE))
+
+# run updates before loading any packages
+if ("--update" %in% argv) {
+  stopifnot(`--update must not be used together with --renv=...` = !any(startsWith(argv, "--renv=")))
```
mikapfl, line 28: This would work with standard double quotes `"` instead of backticks, right?

scripts/utils/archiveRenv.R
```diff
@@ -0,0 +1,12 @@
+# copies current renv.lock to the archive and appends current timestamp
+local({
+  stopifnot(`No renv active. Try starting the R session in the repo root.` = !is.null(renv::project()))
+
+  # update renv.lock
+  renv::snapshot(prompt = FALSE)
+
+  dir.create(here::here("archive"), showWarnings = FALSE)
+  datetime <- format(Sys.time(), "%Y-%m-%dT%H%M%S")
+  file.copy(renv::paths$lockfile(), here::here("archive", paste0(datetime, "_renv.lock")))
```
mikapfl, line 10: `here::here` fails in `output/*/magicc` because those directories are special-cursed and contain a `.git` folder. Could that be a problem?

scripts/utils/restoreRenv.R
```diff
@@ -0,0 +1,24 @@
+# restores current renv to the state described in selected lockfile e.g. from the archive, updates current renv.lock
```
mikapfl, line 1:
```suggestion
# restores current renv to the state described in selected lockfile e.g. from the archive, and writes the restored state to renv.lock
```

tutorials/4_RunningREMINDandMAgPIE.md
```diff
@@ -49,7 +49,7 @@ bash /p/projects/rd3mod/R/libraries/Scripts/create_snapshot_with_day.sh
 
 ### Activate snapshot for REMIND and MAgPIE
 
-Direct the models to the snapshot you just created above by editing [`.Rprofile`](../.Rprofile) in both REMIND's main folder. This file will be copied to the MAgPIE main folder automatically. Uncomment these line by deleting `#` and change the date to today (the `_R4` is necessary if you run `R 4.0` or later):
+Direct the models to the snapshot you just created above by first renaming [snapshot.Rsprofile](../snapshot.Rprofile) to `.Rprofile`. The default `.Rprofile` is activating renv, but coupled runs must use a snapshot at the moment. Then edit this new `.Rprofile` in both (TODO what? two REMIND folderS?) REMIND's main folder. This file will be copied to the MAgPIE main folder automatically. Uncomment the following line by deleting `#` and change the date to today (the `_R4` is necessary if you run `R 4.0` or later):
```
mikapfl, line 52: TODO still needs to be resolved

mikapfl (changes requested):
Looks generally good to me! Huge change and important milestone for reproducibility of results! My smoke testing looked good, the auto-setup worked for me.

I have questions and ideas for changes all over the place.

renv/activate.R
```diff
@@ -0,0 +1,933 @@
+
```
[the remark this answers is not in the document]
pfuehrlich-pik: This file is auto-generated when creating an renv, so I'd not change it. This file is sourced in `.Rprofile`, it is bootstrapping renv and loading it.

renv/activate.R
```diff
@@ -0,0 +1,933 @@
...
+    if (!is.null(override))
+      return(override)
+
+    # next, check environment variables
+    # TODO: prefer using the configuration one in the future
+    envvars <- c(
+      "RENV_CONFIG_AUTOLOADER_ENABLED",
+      "RENV_AUTOLOADER_ENABLED",
+      "RENV_ACTIVATE_PROJECT"
+    )
```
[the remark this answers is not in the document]
pfuehrlich-pik: No idea, I never read the file really :grimacing: Not my TODO anyway :stuck_out_tongue:

.Rprofile
```diff
@@ -1,65 +1,21 @@
...
-# Snapshots must be compatible to the R version used. If you are using R 4.1
-# make sure the selected snapshot's name ends with '_R4'.
-
-# snapshot <- "/p/projects/rd3mod/R/libraries/snapshots/2022_05_31_R4"
-# snapshot <- latestSnapshot
-
-if (exists("snapshot")) {
-  activateSnapshot(snapshot)
+if (!"https://rse.pik-potsdam.de/r/packages" %in% getOption("repos")) {
+  options(repos = c(getOption("repos"), pik = "https://rse.pik-potsdam.de/r/packages"))
```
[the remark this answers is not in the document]
pfuehrlich-pik: The auto-setup will fail without it, because it cannot find all the PIK packages. People from PIK will have this configured probably, so it's just a fallback really, but for people outside PIK it should help.

renv/activate.R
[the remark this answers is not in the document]
mikapfl: Hmm, I see. But I think we need some indication somewhere that this file is auto-generated. Otherwise, people will be confused as I was and start reviewing it or so.

output.R
```diff
@@ -21,6 +21,17 @@ options(error = quote({
...
+  source("scripts/utils/updateRenv.R")
```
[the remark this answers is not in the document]
pfuehrlich-pik: CRAN packages should only be updated manually, or when a PIK package depends on a newer version, yes. It's similar to what we have now: PIK packages are always kept up-to-date and CRAN packages are updated manually or when a PIK package requires it. The main difference would be that updates are no longer shared, everyone has to update CRAN packages on their own.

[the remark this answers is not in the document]
mikapfl: Did we speak about that already somewhere? Maybe that is the optimal policy, maybe not. Because "everyone has to update CRAN packages on their own" is a big change in policy. What we have now is (implicitly): "We'll do CRAN updates for you, don't worry." with a fuzzy definition of "we".

scripts/start/prepare_and_run.R
```diff
@@ -1173,6 +1173,11 @@ run <- function(start_subsequent_runs = TRUE) {
   # Postprocessing / Output Generation
   output    <- cfg$output
   outputdir <- cfg$results_folder
+
+  # make sure the renv used for the run is also used for generating output
+  stopifnot(is.null(renv::project()) || normalizePath(renv::project()) == normalizePath(outputdir))
+  argv <- c(get0("argv"), paste0("--renv=", renv::project(), recycle0 = TRUE))
```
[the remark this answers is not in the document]
pfuehrlich-pik: Well, I'm pretty sure this is an underappreciated feature :wink: Although it's the default for file.path:
```
> file.path(NULL, 'asd')
character(0)
> paste(character(0), 'asd')
[1] " asd"
```
Anyway, I can rewrite with if.

scripts/start/prepare_and_run.R
```diff
@@ -1173,6 +1173,11 @@ run <- function(start_subsequent_runs = TRUE) {
   # Postprocessing / Output Generation
   output    <- cfg$output
   outputdir <- cfg$results_folder
+
+  # make sure the renv used for the run is also used for generating output
+  stopifnot(is.null(renv::project()) || normalizePath(renv::project()) == normalizePath(outputdir))
```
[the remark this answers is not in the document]
pfuehrlich-pik: This is more of a documentation thing that should help with understanding the code, if this still fails I messed up. I added an error message anyway, just in case

renv/activate.R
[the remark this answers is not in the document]
pfuehrlich-pik: > some indication somewhere

seems easy to miss, any concrete ideas where to put that?

output.R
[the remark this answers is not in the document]
pfuehrlich-pik: Hm that's not how I understand our current policy, I'd say it's closer to "never update CRAN packages, unless really necessary". Anyway it's certainly very fuzzy, and I'm open for a discussion on this.

scripts/utils/archiveRenv.R
pfuehrlich-pik: I don't think so. This would only cause problems if the working directory was in a magicc folder when here::here is called, but that should never be the case. Did you actually run into this problem or did you just anticipate this problem from reading the here::here documentation?

tutorials/1_GettingREMIND.md
```diff
@@ -1,98 +1,54 @@
...
+
+REQUIREMENTS
+------------
 
 **Please note: You need input data to run REMIND. At present, there are still legal issues with sharing this data, so we can not publish them yet (but we are working on it, please bear with us). If you have access to the necessary data sources (IEA, etc.) you can generate the input data on your own, we are happy to assist you in doing so.**
 
+- input data
+- git
+- GAMS >= 39.1.0 with CONOPT license
+- R >= 4.0
```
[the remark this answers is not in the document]
pfuehrlich-pik: I removed recycle0

renv/activate.R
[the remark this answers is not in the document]
mikapfl: Put autogenerated files in a folder `autogenerated`. That way you don't have to touch the file itself, and it is hard to miss.

scripts/utils/archiveRenv.R
mikapfl: I anticipated the problem because the `here::here` logic seems a little like black magic. It effectively means if you delete the `.git` folder, it stops working, right? So we shouldn't distribute REMIND releases as tarballs in the future, for example. I mean, I like the idea of `here::here`, and maybe if it works, it works, but it seems a bit fickle to me. Maybe we need a `.here` file in the remind main folder, just to be safe?

output.R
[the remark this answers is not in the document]
pfuehrlich-pik: After discussing with the RSE group again we decided to leave the CRAN package update strategy as it is, and readjust if outdated CRAN packages become a problem.

renv/activate.R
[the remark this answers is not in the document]
pfuehrlich-pik: Actually the whole `renv` folder is auto-generated and completely managed by renv, so it might make things confusing for experienced renv users if we relocate this. I'm still in favor of just leaving this as it is

scripts/utils/archiveRenv.R
pfuehrlich-pik: I added calls to [here::i_am](https://here.r-lib.org//reference/i_am.html), now it works even if `.git` is not present, and it will fail with a somewhat helpful error when here::here is called from outside of a remind (sub)folder

pfuehrlich-pik: Also it now even works when called from `output/*/magicc`

renv/activate.R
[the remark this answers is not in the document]
mikapfl: But, the rest of the `renv` folder is not added to git, right? Having autogenerated stuff around is common, having it checked into version control is not.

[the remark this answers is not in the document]
mikapfl: I mean, it is not a huge issue, I would also be fine with it as-is, as long as this file changes almost never, so it doesn't pollute diffs.

mikapfl (approved):

tutorials/1_GettingREMIND.md
```diff
@@ -1,98 +1,54 @@
...
-  "stringr",
-  "tidyr",
-  "tidyverse",
-  "withr",
-  "yaml"
-)
-install.packages(pkgs)
+```sh
+Rscript checkSetup.R
 ```
```
0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q, line 54: > Navigate to the REMIND folder that you cloned earlier. Always start R scripts and sessions related to REMIND from this path. To install all required R packages and check if your system is ready to run REMIND execute the following:
> 
> ```sh
> Rscript checkSetup.R
> ```

This does not work for new users.  There is no mention anywhere of the RSE repository and how to set it.  The local `.Rprofile` would set it only after hydration, which will have failed by then.  Moving that up (see above) resolves this.

.Rprofile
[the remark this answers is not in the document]
0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q: See comment on `1_GettinREMIND.md`.  This has to go before the bootstrapping, else new user (everybody that does not have the required packages installed) will be left with a defunct REMIND installation.

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q (changes requested):
- `checkSetup.R` should go into `./scripts/utils/`
- `.Rprofile` needs adjustments to work for new users

.Rprofile
```diff
@@ -1,65 +1,21 @@
+source("renv/activate.R")
```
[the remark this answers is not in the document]
pfuehrlich-pik: I added a note in 1_GettingREMIND.md

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q (approved):

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
Not necessarily relevant for this merge request, but still seems the best place to put it.
How would one install a specific package version using `renv`?  REMIND code depends on RSE package versions, and these dependencies are not stated explicitly.  If one wants to run a specific REMIND commit (current example [`aac2264`](https://github.com/remindmodel/remind/commit/aac2264b821e3540678649df9944acc22eb8e2ec), which is the last know version definitely known to be working thanks to the [AMT](https://gitlab.pik-potsdam.de/REMIND/testing_suite/-/blob/32ba26e9146426284d4d3abc8bc62b08a35fc2b9/README.md)), that will break with versions of the `edgeTransport` package lower than 0.19.0 (for [reasons](https://github.com/remindmodel/remind/issues/945)).
The [RSE wiki](https://redmine.pik-potsdam.de/projects/mo/wiki/Configuration_of_repositories_-_Add_the_PIK_repository#CRAN-like-repositories-at-PIK) states that 
> `https://rse.pik-potsdam.de/` […] supports installing old package versions with renv

1. The documentation should lay out how this is done.
2. A tool/script for installing all RSE packages in the versions at a specific point in time would be nifty.

pfuehrlich-pik:
1. Good point, I added `renv::install("package@1.2.3")` to the [RSE wiki](https://redmine.pik-potsdam.de/projects/mo/wiki/Configuration_of_repositories_-_Add_the_PIK_repository#CRAN-like-repositories-at-PIK). This is also explained in `11_ManagingRenv.md`, section `renv functions`.

2. I guess there are a couple ways to do this:
* It might just be enough to have the renv.lock files from the personal archive. If the package env is broken in some way it's also easy to share a working renv.lock file with everyone.
* We could create centralized renv.lock files with timestamps for {each update of any package,each day,each week}. This could also define a standard in which to test stuff before pushing, and it would make it more likely that the renv cache already has a package even before anyone installs it manually.
* The hard way would be to analyze timestamps of git commits and find out which package versions were active at the given time, so they can be installed manually via renv.

I'd go with the first option for now.

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
> […] `renv::install("package@1.2.3")` […] is also explained in `11_ManagingRenv.md`, section `renv functions`.

I missed that while skimming it.  :+1: 

For the "back in time" functionality, we will need option three at some point.  Else we are dependent on people having "accidentally" created a `renv.lock` on just the right day.  And that will not be the case just when we need it.

pfuehrlich-pik:
For the back-in-time feature I created a [lockfile archive repo](https://github.com/pik-piam/lockfile-archive), and Anastasis is helping me to set up a cronjob updating the lockfiles each day.

PR #812, opened 2022-05-12
avoid inconsistent policyCosts if reporting is rerun, fix identical logs for output slurms

orichters:
- in output.R, if multiple single output scripts were selected, they were writing to the same log. Now, save the log in the outputfolder such as `log_validation.txt` or `log_reporting.txt`, also print the log name and use it as slurm name so it can be found easier
- in policyCosts.R, copy the fulldata.gdx of the reference run to input_refpolicycost.gdx in the output folder to get consistency
- in reporting.R, delete the adjusted policy mif. Because input_refpolicycost.gdx was overwritten, the normal mif file then contains the correct policy comparison
- bit of cleanup, better logging including time stamps

### DIFF
--- a/scripts/output/comparison/policyCosts.R
+++ b/scripts/output/comparison/policyCosts.R
@@ -7,16 +7,16 @@
 
 # policyCosts.R
 #
-# This script produces a pdf in which the policy costs of policy-runs with 
+# This script produces a pdf in which the policy costs of policy-runs with
 # respect to specified reference runs are displayed.
-# It can be called via the output.R script. If that is the case, the order 
+# It can be called via the output.R script. If that is the case, the order
 # in which the runs are selected is important:
 #   1: policy run 1
 #   2: reference run 1
 #   3: policy run 2
 #   4: reference run 2
-# and so on... 
-# 
+# and so on...
+#
 
 suppressPackageStartupMessages(library(tidyverse))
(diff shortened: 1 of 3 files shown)


scripts/output/comparison/policyCosts.R
```diff
@@ -233,102 +233,108 @@ happy_with_input <- FALSE
...
+      message(crayon::red("Stopping execution now.\n\n"))
+      stop("I can't figure this **** out. I give up.")
     }
   } else {
     happy_with_input <- TRUE
   }
 }
 
 
+message("Copy fulldata.gdx of policy refs to input_refpolicycost.gdx in output folders, eventually overwriting these files.")
```
johanneskoch94, line 306: I would say "overwriting these files if they exist" instead of eventually. (Eventually a false-friend ;) )

johanneskoch94 (commented):
Looks good to me. I think you can merge as soon as the conflicts are fixed.  Thanks!

scripts/output/comparison/policyCosts.R
orichters: Haha, thanks for spotting that :)

johanneskoch94 (approved):

LaviniaBaumstark (approved):

PR #817, opened 2022-05-14
Update scenario_config_AMT.csv

giannou:


orichters (approved):
haven't tested it, but if this breaks something, we can burn the whole model down :)

giannou:
For documentation purposes: adding the `-Base` string to a run name is important, especially because of the dash `-`, otherwise the REMIND testing suite cannot process the list of tested scenarios properly

PR #632, opened 2022-01-18
Enable divergent scenarios for NGFS v3

orichters:
realization `21_tax/on`
----------------
- we introduce a switch `cm_CO2TaxSectorMarkup` that allows to add a markup to the CO2 tax in buildings and transport sector. It can for example be specified as `GLO.build 1, USA_regi.trans 0.25, EUR_regi.trans 0.25` to have 100% more CO2 tax in buildings sector globally, and 25% higher CO2 taxes in transport sector of USA and Europe.
- `vm_emiCO2Sector` contains the CO2 emissions of build and trans sectors, as provided by @fschreyer.
- With `v21_taxrevCO2Sector` and `p21_taxrevCO2Sector0`, we implement an iterative process to realize budget neutrality, but marginal effectiveness of the tax

realization `45_carbonprice/NDC`
--------------
- add switch `cm_NDC_divergentScenario` so one can choose how quickly the carbon prices converge after the NDC targets. The default remains 2100. I decided to implement it like that and not allowing to choose the convergence year directly, because we may change further things for these scenarios.

PR #826, opened 2022-05-19
add regional ccs differentiation

orichters:
This commit implements a regional ccs differentiation required for the NGFS scenarios to avoid excessive use of CCS in the delayed-transition scenario.

The switch `c_ccsinjecrateRegi` allows to specify for each region, or globally, that the injection rate factor is adapted. The factor `sm_ccsinjecrate` is simply multiplied by `pm_ccsinjecrateRegi(regi)` everywhere. So setting it to `1` (as in the default case where it is set to `off`) has no impact, while setting it to 0.6 reduces CCS injection by 40%.

The input uses the same principle as used already for [`p21_CO2TaxSectorMarkup`](https://github.com/remindmodel/remind/blob/72791c0ab406fa09173cc55b7e28775d6da1451c/modules/21_tax/on/datainput.gms#L231-L242) which was introduced in #632 and was based on a similar code in `47_regipol` developed by @Renato-Rodrigues.

Examples with this feature are currently running at: `/p/tmp/oliverr/remind/output/d_delfrag_bIT_2022-05-19_12.32.20` and `d_delfrag_bIT_ccs2_2022-05-19_12.32.21`.

I add @christophbertram for the NGFS, @Renato-Rodrigues for maybe having a short look at the code in `core/datainput.gms` and @LaviniaBaumstark as the queen of REMIND. Who else should be added? CCS experts?

<details>
<summary>Click here for file comparison: ../remind-develop/config/scenario_config_NGFS_v3.csv -> config/scenario_config_NGFS_v3.csv</summary>

```
Using configfile config/default.cfg as default.
Columns deleted: -
Columns added:   c_ccsinjecrateRegi
Renamed columns: -
Renamed rows:    -

Changes in the scenarios:
~ o_1p5c_bIT:
    slurmConfig: 14 -> 5 (default: )
~ o_1p5c_bIT_46NDC:
    slurmConfig: 14 -> 5 (default: )
~ o_2c_bIT:
    slurmConfig: 14 -> 5 (default: )
~ d_delfrag_bIT:
    slurmConfig: 14 -> 5 (default: )
~ d_delfrag_bIT_ccs2:
    slurmConfig: 14 -> 5 (default: )
~ d_delfrag_bIT_46NDC:
    slurmConfig: 14 -> 5 (default: )
~ d_rap_bIT:
    slurmConfig: 14 -> 5 (default: )
```
</details>


christophbertram (approved):
This looks very solid to me, thanks.

core/datainput.gms
```diff
@@ -239,6 +239,19 @@ if (c_ccsinjecratescen eq 3, sm_ccsinjecrate = sm_ccsinjecrate *   1.50 ); !! Up
 if (c_ccsinjecratescen eq 4, sm_ccsinjecrate = sm_ccsinjecrate * 200    ); !! remove flow constraint for DAC runs
 if (c_ccsinjecratescen eq 5, sm_ccsinjecrate = sm_ccsinjecrate *   0.20 ); !! sustainable estimate
 
+*** OR: regional reduction of ccs injection rate
+$ifThen.c_ccsinjecrateRegi not "%c_ccsinjecrateRegi%" == "off"
+Parameter
+  p_extRegiccsinjecrateRegi(ext_regi) "regional factor that limits CCS injection: 1 means no change, 0.6 means 40% reduction of CCS injection rate factor in respective region (extended regions)" / %c_ccsinjecrateRegi% /
```
LaviniaBaumstark, line 245: is there a reason, why you put the declaration after the if statement? I would prefer to always declare this parameter, also to avoid problems for fixing times and in the reporting

LaviniaBaumstark:
added Jessi for review

core/datainput.gms
orichters: good question. I did it as I did in the tax module, and I think that I discussed it with @Renato-Rodrigues then and there was a reason for it, but I do not remember exactly and maybe I misunderstood him. But I hope he can clarify.

orichters:
What I suggest: set `c_ccsinjecrateRegi` in `default.cfg` not to `off`, but to `GLO 1`. This way, it is easier for everyone to slightly adapt the CCS injection rate and we don't have this split into `off` and `not off`. Still, waiting for @Renato-Rodrigues do weigh in.

core/datainput.gms
Renato-Rodrigues: Placing the declaration inside the if statement is a way of not adding unnecessary content and size to the gdx file. 
It is usually more justifiable when you have a higher dimensional parameter, but I also don't see any issue on doing in this case. 
This would only throw an error in the fixings, if you are running an older remind version with a newer gdx. This is not good practice anyway, as you should instead in this case use the older gdx file (as the calibration for your model version was made for the old gdx), or update your code instead, in which case you wouldn't have any fixing error anyway.

Renato-Rodrigues: ON unrelated note, Why use a multiplicative factor in this case?
Wouldn't it be better to simply set the regions values directly. I think that using the multiplicative factor just creates confusion on understanding the actual values that is being applied to the regions that have the value overwritten.

Renato-Rodrigues: I would instead change the sm_ccsinjecrate to a parameter dependable of regi, assign the value of sm_ccsinjecrate to all regions by default, and then overwrite any region that containes values in the c_ccsinjecrateRegi.

This way we would have in the final gdx exactly the value that is being applied as injection rate directly in a parameter, instead of having to recalculate it from two different swicthes whenever we want to check results.

orichters: so you mean. turn Scalar `sm_ccsinjecrate` into a parameter `sm_ccsinjecrate(regi)` and set it to the scenario value times the regional setting and that is it?

orichters: to the first comment: wouldn't this throw an error also if you used a run that used the regional differentiation and therefore contains `p_extRegiccsinjecrateRegi` as fixing for a run that has it switched `off` and therefore never declared?

Renato-Rodrigues: yes, but I wouldn't set also sm_ccsinjecrate(regi) as a multiplicative term, but instead the absolute value. So:

pm_ccsinjecrate(regi) = sm_ccsinjecrate;
pm_ccsinjecrate(regi)$sm_ccsinjecrate(regi) = sm_ccsinjecrate(regi);

Renato-Rodrigues: also, if you want to be more general, you should use instead the ext_regi in the c_ccsinjecrateRegi switch. 
That way you can define values overwrites to region groups, instead of only remind regions, increasing compatibility with REMIND-21 regions. 
This allows you for example to set a value to "EUR_regi", instead of "EUR", and the same switch value would be compatible with REMIND-H12 and REMIND-21 region runs.

If you want a n example of how to use this let me know and I can show you quickly in a short call.

Renato-Rodrigues: > to the first comment: wouldn't this throw an error also if you used a run that used the regional differentiation and therefore contains `p_extRegiccsinjecrateRegi` as fixing for a run that has it switched `off` and therefore never declared?

Yes, but is this case even feasible or a design error? If your intention is to apply this as a baseline policy, this should never be disabled in the first place.
In this case simply remove any if conditional applied to it.
If you intend this as a policy scenario, then the error case you mention should not happen by logic design.

orichters: 1.
> If you intend this as a policy scenario, then the error case you mention should not happen by logic design.
Content-wise, I agree, but I prefer to try and write the code such that it doesn't break if somebody sets up scenarios that I haven't thought about

2.
> should use instead the ext_regi in the c_ccsinjecrateRegi switch.
That is what I do right now, don't I? Or do you mean I should maintain that even if I adapt the code to implement the other suggested changes?

3.
> I wouldn't set also c_ccsinjecrateRegi(regi) as a multiplicative term, but instead the absolute value
I'm not sure. I see the advantage of the advantage that you can set the injection rate explicitly. But: Let us say you generally want to have less CCS in Europe, but with different scenario settings. I think in this case, it is more transparent to have always `EUR_regi 0.9` in your scenario config instead of different values that than have to be compared to the scenario value in `default.cfg`. And if somebody ever changes the injection rates associated with `c_ccsinjecratescen`, then all the regi values have to be adapted to keep the order of the CCS injection rates fixed. But in the end, I think having it as absolute or multiplicative is rather a matter of taste. Maybe @LaviniaBaumstark can weigh in?

Renato-Rodrigues: 2. Sorry about that, you are perfectly right. I didn't check the code in detail before making my comment.

3. While debugging, it is always preferable to have the final values used in the model directly available, then forcing the person debugging to check two parameters and to make a multiplication to figure out which value is being actually applied to the model as injection rate. That is why I don't think this is a matter of taste.
Also, I see as a undesirable effect changing a region specific injection rate if you change the global value.

orichters:
I updated the PR according to the requests. Simply switching to `pm_ccsinjecrate(regi)` is very useful, I think.

Now, `c_ccsinjecrateRegi` has to be specified not relative to the value determined by `c_ccsinjecratescen`, but rather as an absolute value. @Renato-Rodrigues emphasizes the advantages in terms of debugging. According to my taste, an input such as `GLO 0.000625, CAZ_regi 0.00225, CHA_regi 0.002, EUR_regi 0.00225, IND_regi 0.002, JPN_regi 0.001, USA_regi 0.001` is less comprehensible compared to the older `GLO 0.25, CAZ_regi 0.9, CHA_regi 0.8, EUR_regi 0.9, IND_regi 0.8, JPN_regi 0.4, USA_regi 0.4`, but I don't see an option without drawbacks.

Renato-Rodrigues (approved):

LaviniaBaumstark (approved):

config/default.cfg
```diff
@@ -951,6 +952,9 @@ cfg$gms$c_BaselineAgriEmiRed <- 0
 #  (3) upper estimate: 0.075
 #  (4) unconstrained: 1
 #  (5) sustainability case: 0.001
+# c_ccsinjecrateRegi  "regional upper bound of the CCS injection rate, overwrites settings set with c_ccsinjectratescen"
+#  ("off") no regional differentiation
+#  ("GLO 0.000625, CAZ_regi 0.00225, CHA_regi 0.002, EUR_regi 0.00225, IND_regi 0.002, JPN_regi 0.001, USA_regi 0.001"): "example that is taylored such that NDC goals are achieved without excessive CCS in a delayed transition scenario with c_ccsinjecratescen = 1. Globally, 75% reduction, 10% reduction in CAZ etc."
```
strefler, line 957: This comments sounds like the default value would be 0.0025, but that is the lower estimate. The default is 0.005.

strefler (commented):
I'm a bit indifferent between absolute and relative values. But a comment on the default value might help to assess the values (which I think is in there already, just maybe a bit hidden, and it relates to the lower estimate instead of the default).

PR #828, opened 2022-05-20
add fixings/levs/margs manipulations for trade and transport

orichters:
such that we can reuse runs some weeks old as fixings after #796 and #800


PhilippVerpoort:
Looks good to me. Thanks!

PhilippVerpoort (approved):

PR #829, opened 2022-05-23
increase mem for tasks-per-node=1 runs

orichters:
to avoid that output generation fails, see https://github.com/pik-piam/discussions/discussions/7

Also, log the correct title if run is started with `Rscript start.R` but `cfg$title` was changed manually.


LaviniaBaumstark (approved):

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q (approved):

orichters:
Proof of work, see `/p/tmp/oliverr/remind/output/`:
- testOneRegi-12nodes_2022-05-23_11.25.33: insignificant SLURM OOM event
- testOneRegi-1node_2022-05-23_11.31.52: reporting fails
- testOneRegi-1node-mem8000_2022-05-23_11.25.56: insignificat SLURM OOM event

PR #830, opened 2022-05-23
Energy Security Scenario

pweigmann:
Introduces a new switch `cm_EnSecScen` to enable a PE tax on gas in Germany between 2020 and 2050.


Renato-Rodrigues:
If this is a German only tax, its code should be included in the regipol module and not in the core.

Also, did you considered using the already existent pe2se tax for that instead? 

If you add a tax at the PE level, you would need to make sure this does not cause undesirable effects in the trade formulation.
As we already have taxes applied at the pe2se level, maybe it would be easier and simpler from the model perspective to adjust the taxes in this level instead to achieve the same goal. Did you considered the option?

main.gms
```diff
@@ -85,9 +85,9 @@
 * 
 * Regionscode: 62eff8f7
 * 
-* Input data revision: 6.298001
+* Input data revision: 6.304
 * 
-* Last modification (input data): Mon Apr 11 14:41:17 2022
+* Last modification (input data): Fri May 20 12:34:46 2022
 *
```
fschreyer, line 91: In general, we usually don't modify the run names and settings of the main.gms. You can avoid committing all of a file by using someting like ``git add -pu``. Easiest way to get default settings in main.gms is to run ``Rscript start.R`` and commit the resulting main.gms.

modules/21_tax/on/datainput.gms
```diff
@@ -139,9 +139,24 @@ $offdelim
...
 *LB* use 0 for all regions as default
-p21_tau_xpres_tax(ttot,regi,all_enty) = 0;  
+p21_tau_xpres_tax(ttot,regi,all_enty) = 0;
+
+
+*** -------------------------Primary Energy Tax--------------------------
+*** initialize tax to zero
+p21_tau_pe_tax(ttot,regi,all_enty) = 0;
+
+*PW* choose tax path for gas in Germany
```
fschreyer, line 149: Well, Renato is right. We have this convention of only using regional names in the regipol module. So this should be moved to ``regipol/regiCarbonPrice/datainput.gms``

modules/21_tax/on/equations.gms
```diff
@@ -105,6 +106,14 @@ v21_taxrevNetNegEmi(t,regi) =e= cm_frac_NetNegEmi * pm_taxCO2eqSum(t,regi) * v21
...
 v21_emiALLco2neg(t,regi) =e= -vm_emiAll(t,regi,"co2") + v21_emiALLco2neg_slack(t,regi);
 
+***---------------------------------------------------------------------------
+*'  Calculation of PE tax: tax rate times primary energy
+*'  Documentation of overall tax approach is above at q21_taxrev.
+***---------------------------------------------------------------------------
+q21_taxrevPE(t,regi)$(t.val ge max(2010,cm_startyear))..
+v21_taxrevPE(t,regi) =e= p21_tau_pe_tax(t,regi,"pegas") * vm_prodPe(t,regi,"pegas")
+                          - p21_taxrevPE0(t,regi);
+
```
fschreyer, line 116: That should be formulated generally for all PEs. That way, other people can use the tax for other PE energy carriers and purposes. I will dm you about it.

fschreyer (changes requested):

modules/21_tax/on/datainput.gms
orichters: Alternatively, you could try to define it without a specific region and specify it in the config file as only applying to Germany.

fschreyer:
> Also, did you considered using the already existent pe2se tax for that instead?

Yes Renato, we thought about that. The current ``pe2se`` tax is placed on ``vm_prodSe``. However, we need the tax to take effect on the PE price.

> If you add a tax at the PE level, you would need to make sure this does not cause undesirable effects in the trade formulation. 

We can keep it in mind but I don't think it would. For now, it is just applied for one specific scenario for Germany. However, it could be adapted later for more general purposes.

modules/21_tax/on/datainput.gms
Renato-Rodrigues: > Alternatively, you could try to define it without a specific region and specify it in the config file as only applying to Germany.

Ideally, any switch that include regional specification should also belong to the regipol module. This is not always possible because how the code is currently organized, but we should aim for that as much as we can.

Renato-Rodrigues:
> > Also, did you considered using the already existent pe2se tax for that instead?
> 
> Yes Renato, we thought about that. The current `pe2se` tax is placed on `vm_prodSe`. However, we need the tax to take effect on the PE price.
> 
> > If you add a tax at the PE level, you would need to make sure this does not cause undesirable effects in the trade formulation.
> 
> We can keep it in mind but I don't think it would. For now, it is just applied for one specific scenario for Germany. However, it could be adapted later for more general purposes.

Just be careful because any tax in PE level, acts as double accounting in any other tax defined afterwards. Ultimately you apply double taxation, which is undesirable and avoided at a legislation level. So using this for a one case policy scenario is ok, but generalizing PE taxation is quite a dangerous path.

modules/21_tax/on/datainput.gms
pweigmann: Thanks for the comments, will move it tomorrow!

PR #831, opened 2022-05-24
remove unused input variable from output.R

pfuehrlich-pik:
`cfg['input']` is `NULL`, but R automagically expands `cfg$input` to `cfg$inputRevision` (partial matching) so it did not crash. It seems like `input` is not used anyway so I removed it. This might even fix a problem in shiny app.

### DIFF
--- a/output.R
+++ b/output.R
@@ -293,7 +293,6 @@ if (comp == TRUE) {
     if (exists("cfg")) {
       title <- cfg$title
       gms <- cfg$gms
-      input <- cfg$input
       revision <- cfg$inputRevision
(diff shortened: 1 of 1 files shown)


giannou (approved):
Approving but inviting also others to review

orichters (approved):
output.R is "Based on the Version 2.2 of same file in the MAgPIE main folder", and cfg$input is used in MAgPIE. So probably a legacy from copying this file.

chroetz:
Maybe `options(warnPartialMatchDollar = TRUE)` can be set as a default to catch such weird behavior?

PR #833, opened 2022-05-25
avoid that reporting is called with low memory

orichters:
the normal output reporting cannot be run with the standard memory setting. After this PR, the low memory settings are not offered anymore, if `reporting` is selected.

### DIFF
--- a/output.R
+++ b/output.R
@@ -137,8 +137,10 @@ choose_mode <- function(title = "Please choose the output mode") {
   return(comp)
 }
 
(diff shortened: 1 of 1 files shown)


robinhasse (commented):
~~I am not sure if I got your reasoning behind this PR. What would we loose if we remove the options with low memory all together?~~

robinhasse (approved):
Okay, got it now. Looky good to me.

orichters:
Many reporting scripts don't require high memory settings, and it is not necessary to block much memory if it isn't needed.

PR #835, opened 2022-05-25
allow using oldruns for MAgPIE only, move log to output folder

orichters:
- The new parallel mode has a problem if the oldrun feature should be used for MAgPIE (specifying a folder that contains run), but not for the REMING gdx. With this PR, these two choices are independent, so you can use MAgPIE or REMIND or both or none.
- fix log file name for subsequent runs which ended on `.txt.txt`
- for REMIND-MAgPIE in parallel mode, move log to output folder. This requires that `submit()` tolerates that the folder already exists. Maybe @LaviniaBaumstark can check this part.
- add `Sys.sleep(10)` between starting of subsequent runs to maybe reduce the chances of lock conflicts
- make logging of directory change more readable
- replace the tedious list of things to be saved to the RData file from the environment that has to be updated regularly (now: run_compareScenarios was missing) to simply all things from the new environment. That is what environments are made for, after all :)
- improve the call to compareScenarios2: skip the shortTerm version and drop the timestamp, reduce log clutter
- save both compareScenarios2 logs (stderr and stdout) to same file as agreed with @chroetz 

For review, make sure you set up "hide whitespace" by clicking at the gear wheel or [clicking this link](https://github.com/remindmodel/remind/pull/835/files?w=1)
image


start_bundle_coupled.R
```diff
@@ -100,6 +100,7 @@ parallel <- ifelse("--parallel" %in% argv, TRUE, FALSE)
 errorsfound <- 0
 startedRuns <- 0
 waitingRuns <- 0
+deletedFolders <- 0
```
orichters, line 103: to be more transparent, the start script tells users about folders that have been (in test mode: would be) deleted.

start_bundle_coupled.R
```diff
@@ -474,7 +477,12 @@ for(scen in common){
     if (i == start_iter_first && ! start_now && all(file.exists(cfg_rem$files2export$start[path_gdx_list]) | unlist(gdx_na))) {
         start_now <- TRUE
     }
-
```
orichters, line 477: delete folders for runs that it wants to start. This can happen only if these folders could not be used to continue the scenario from there, for some example they don't contain fulldata.gdx or mif file

start_bundle_coupled.R
```diff
@@ -520,7 +528,8 @@ for(scen in common){
   if (start_now){
       startedRuns <- startedRuns + 1
       if (! "--test" %in% argv) {
-        logfile <- file.path("output", paste0("log_", sub("-rem-", if(start_magpie) "-mag-" else "-rem-", fullrunname), if (! parallel) stamp, ".txt"))
+        logfile <- if (parallel) file.path("output", fullrunname, paste0("log", if (start_magpie) "-mag", ".txt"))
```
orichters, line 531: in parallel mode, move the log to output subfolder

scripts/start/submit.R
```diff
@@ -19,7 +19,7 @@
 
 ############## Define function: runsubmit #########################
 
-submit <- function(cfg, restart = FALSE) {
+submit <- function(cfg, restart = FALSE, stopOnDeleteError = TRUE) {
```
orichters, line 22: the submit function must tolerate that the output subfolder already exists if the slurm job should continue writing the log there.

LaviniaBaumstark (approved):
I assume that for coupled runs you managed to ensure that no existing folders are overwritten when you turn on stopOnDeleteError = FALSE

orichters:
> I assume that for coupled runs you managed to ensure that no existing folders are overwritten when you turn on stopOnDeleteError = FALSE

In serial mode, `cfg_rem$force_replace  <- TRUE` always was and still is the default, so if folders exist, they are overwritten anyway. It is implemented such that `start_bundle_coupled.R` checks for existing folders and sets `start_iter` such that no folder containing meaningful data is overwritten.

In the new parallel mode, the deletion is now performed not by setting `cfg_rem$force_replace <- TRUE`, but very explicitly in `start_bundle_coupled.R`, which has the advantage that the script (1) tells you about it and (2) reports that already in `--test` mode so you can check whether that is intended. But then, it creates the output subfolder to write the log inside.

Maybe I introduced some confusion here calling it a `DeleteError`, when it is actually a `CreateError`: Usually, if the folder already exists and `cfg$force_replace = FALSE`, it throws stops because the folder can't be created. But in the new parallel mode, the slurm job should already be writing the log to the folder, so it must accept that the folder already exists.

What I did in my latest commit: I renamed it from `DeleteError` to `CreateFolderError`, and accept that only if the folder contains only files starting with `log`. This should be safe now.

orichters:
I tested it in parallel mode, it works as intended. If a folder exists already, but contains no fulldata.gdx or miffile, it will be deleted, if you restart the same scenario. This works as before.
Then the folder is recreated and the log is written and everything works as intended. But if I quickly copy a different file into this folder, it then runs into an error. So given your approval, the additional safety feature and tests, I will merge it now.

PR #837, opened 2022-05-27
read modelstat also from non_optimal to avoid that it is always 2

orichters:
Writing the debugging tutorial, I struggled a bit with `modelstat`, because it never seemed to help me in any way. `rs2` notoriously printed either `NA` or `2` to me, whatever infeasibilities I had, and also the modelstat output I added to the REMIND log showed the same behavior.

I understand now why: Both extract the value from `fulldata.gdx` which is a file that by definition is only written if `modelstat = 2`, which is a bit pointless.

With this PR, it now extracts `o_modelstat` and `o_iterationNumber` from `fulldata.gdx` and `non_optimal.gdx`. It then selects the modelstat value from the gdx file which has the higher iteration number and prints that to the author and also submits it to `runstatistics`. I added a small explanation also what the modelstat means, which then looks like that:

```
Model summary:
  gams_runtime is 0.6 secs.
  full.gms exists, so the REMIND GAMS code was generated.
  full.log and full.lst exist, so GAMS did run.
  abort.gdx does not exist, a file written automatically for some types of errors.
  non_optimal.gdx does not exist, a file written if at least one iteration did not find a locally optimal solution.
  fulldata.gdx exists, so at least one iteration was successful.
  Modelstat after 38 iterations: 2 (Locally Optimal)
  full.log states: *** Status: Normal completion
```
We might what to adapt that also in rs2.

### DIFF
--- a/scripts/start/prepare_and_run.R
+++ b/scripts/start/prepare_and_run.R
@@ -993,6 +993,19 @@ run <- function(start_subsequent_runs = TRUE) {
     cat("\nREMIND was compiled but not executed, because cfg$action was set to 'c'\n\n")
   }
 
+  explain_modelstat <- c("1" = "Optimal", "2" = "Locally Optimal", "3" = "Unbounded", "4" = "Infeasible",
+                         "5" = "Locally Infeasible", "6" = "Intermediate Infeasible", "7" = "Intermediate Nonoptimal")
+
+  modelstat <- NA
+  max_iter <- 0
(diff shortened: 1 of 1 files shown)


giannou (approved):
nice catch!

LaviniaBaumstark (approved):

orichters:
I moved the message out of the `if (file.exists("fulldata.gdx"))` block, because that was a bit pointless as well :)

PR #839, opened 2022-05-31
R4 snapshot support

pfuehrlich-pik:
This PR makes sure that snapshots work with R 4.1

### DIFF
--- a/.Rprofile
+++ b/.Rprofile
@@ -28,6 +28,7 @@ activateSnapshot <- function(snapshot) {
     # setting include.site to FALSE makes sure that only the snapshot and system libraries are used
     .libPaths(snapshot, include.site = FALSE)
   }
+  message("libPaths was set to: ", snapshot)
 }
 
 # Just uncomment the following line and set the snapshot path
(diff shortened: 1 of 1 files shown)


.Rprofile
```diff
@@ -1,45 +1,65 @@
...
-             path = '/p/projects/rd3mod/R/libraries/snapshots',
-             pattern = '^20[0-9]{2}_[0-9]{2}(_[0-9]{2})?$',
-             full.names = TRUE), decreasing = TRUE)
-latest_snapshot <- ifelse (length(snapshots) > 0, snapshots[[1]], NA)
+snapshots <- sort(list.files(path = '/p/projects/rd3mod/R/libraries/snapshots',
+                             pattern = paste0('^20[0-9]{2}_[0-9]{2}(_[0-9]{2})?', maybeR4, '$'),
+                             full.names = TRUE), decreasing = TRUE)
+latestSnapshot <- if (length(snapshots) > 0) snapshots[[1]] else NA
+
+activateSnapshot <- function(snapshot) {
```
orichters, line 21: Wouldn't it be clever to check here again whether the snapshot contains or does not contain `_R4`, depending on `R.version$major`? Because certainly someone (me) will try to specify a snapshot manually that doesn't fit the R version and I would prefer that the script checks this.

orichters (approved):
I tested it and it seems to work.

PR #842, opened 2022-06-01
bugfixes and code cleaning in damage modules

orichters:
This PR:
- fixes two actual bugs, were a `);` was still there that finishes an `if` statement that was outcommented. This then ended the iteration loop and killed everything: [modules/51_internalizeDamages/KW_SEitr/postsolve.gms](https://github.com/remindmodel/remind/compare/develop...orichters:NGFS_updates#diff-f5e4f9cc2525dc97ba91ac5008910cf1718c1c683273983fc505187017db95ae), [modules/51_internalizeDamages/LabItr/postsolve.gms](https://github.com/remindmodel/remind/compare/develop...orichters:NGFS_updates#diff-b01b25989e497c24295b4bacb8a981d3a114af428124a24c6caf5f2d971a87ab): 
- fixes several a bit overengineered statements with a for-loop, where in fact a simple attribution of the value is no problem, because `ttot` is a subset of `tall`.
- added the files that are read by [50_damages/TC/datainput.gms](https://github.com/remindmodel/remind/tree/develop/modules/50_damages/TC/datainput.gms) and [50_damages/KWTCint/datainput.gms](https://github.com/remindmodel/remind/tree/develop/modules/50_damages/KWTCint/datainput.gms) to the `input/files`. They have to be added to the input data such that nothing fails: `TC_df_parameters_const.csv`, `TC_df_parameters_tasK.csv`, `gdp_countryFrac_ann.csv`. Before, they must be renamed to `f50_*` I guess. Or are these [`f50_TC_df_const.cs4r` and `f50_TC_df_tasK.cs4r`](https://github.com/pik-piam/mrremind/blob/master/R/fullREMIND.R#L67-L68) from mrremind? And maybe we should not hardcode `gdp_SSP2EU` [here](https://github.com/remindmodel/remind/blob/develop/modules/50_damages/TC/datainput.gms#L43)

I ran the `testOneRegi` runs in `/p/tmp/oliverr/remind/config/scenario_config_damage_tests.csv`, and they all ended successfully, results can be found in `/p/tmp/oliverr/remind/output/testOneRegi-*` with most recent time-stamp


fpiontek:
Looks good, thanks for cleaning up! The f50_TC... files from mrremind are indeed the input files already, they code in the TC module has to be adjusted for their format though.

fpiontek (commented):
Looks good, thanks for cleaning this old code up.

PR #843, opened 2022-06-01
Replace edgeTrpLib references with edgeTranport ones.

Loisel:
This is part of an effort to merge `edgeTrpLib` and `edgeTransport` to reduce the organizational overhead generated by the parallel development of two packages.

### DIFF
--- a/scripts/iterative/EDGE_transport.R
+++ b/scripts/iterative/EDGE_transport.R
@@ -17,7 +17,7 @@ opt = parse_args(opt_parser);
 library(data.table)
 library(gdx)
 library(gdxdt)
-library(edgeTrpLib)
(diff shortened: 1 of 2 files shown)


LaviniaBaumstark:
does it mean `edgeTrpLib` substitutes whole functionality of `edgeTransport`? What does this mean for the input data and input data generation?

Loisel:
the other way round. `edgeTrpLib` will be deprecated. In `mrremind`, only `edgeTransport` functions are called, so no problem. Calls have to be replaced in REMIND and edgeTransport itself, see https://github.com/pik-piam/edgeTransport/pull/160

LaviniaBaumstark (approved):

PR #845, opened 2022-06-02
Add HydrHype4 scenario to the set of available EDGE-T scenarios.

Loisel:


modules/35_transport/edge_esm/sets.gms
```diff
@@ -122,7 +122,7 @@ ConvCase
 ConvCaseWise
 ElecEra
 ElecEraWise
-HydrHype
+HydrHype4
```
LaviniaBaumstark, line 125: does this mean that it only works with new input data? Does it give an error for input data containing "HydrHype" ?

Loisel: `HydrHype` is deprecated and not produced by `mrremind` since quite a while. Can be a problem if one tries to use new REMIND with old input data. Does this happen?

orichters: And there are still config files that use it, see: `grep ";HydrHype;" config/*.csv config/*/*.csv`.

orichters:
I'm fine with the change. My impression is that in general we have no workflow that makes sure that the scenario config files in the repository don't become outdated, and don't even have warnings for old settings apart from [some very specific cases](https://github.com/remindmodel/remind/blob/develop/start.R#L388-L391) that I implemented a while ago. Ideally, each time a deprecated setting is used, some warning should show up explaining what to do in this case. But I don't think this PR necessarily is the right place to discuss such a general question, so go ahead and merge it if Lavinia is fine with it…

LaviniaBaumstark (approved):

PR #846, opened 2022-06-02
Generate DEU pdf for EUR in compareScenarios2

fbenke-pik:
Generate an additional pdf for EUR region with main region DEU and exclude reference models that are meaningless in this context (e.g. due to coarse regional disaggregation).


fschreyer (approved):
Looks good but I am no expert on the code here. If tested that it doesnt break usual mifs with 12 regions, then all is good.

fbenke-pik:
@chroetz Any thoughts on how to handle the passing of the optional parameter `modelsHistExclude` in a more robust and elegant manner?

chroetz:
> @chroetz Any thoughts on how to handle the passing of the optional parameter `modelsHistExclude` in a more robust and elegant manner?

Hm, I don't see an elegant way. I guess we should not call a model `'"; rm -rf *;` ... 
The alternative would be to save all arguments for cs2() in a file, just pass the name of this file in the command line, then load all arguments from the file when actually calling cs2(). Also not really elegant

chroetz (approved):
Also tested it for H12 (but not for EUR). Works without problems.

PR #851, opened 2022-06-07
actually use polCosts in cs2

orichters:
There was a stupid error: If the output directories were given as relative paths (such as using `output.R`), the `_adjustedPolicyCost.mif` was never used because the `file.exists` check happened while in the wrong working directory.


chroetz (approved):
Makes sense. 
It is a good idea to print out which mifs are used. Thanks!

PR #853, opened 2022-06-08
allow to simply use AMT runs as path_gdx...

orichters:
If you specify something such as `SSP2-EU` in your scenario config as `path_gdx_...` and `SSP2-EU` is a scenario that is **not** started, `start.R` picks the `fulldata.gdx` of the latest REMIND run with that name that contains `"*** Status: Normal completion"` in its log. This PR extends this feature to the REMIND modeltests: If nothing is found this way, the script looks in `cfg$modeltests_folder` for such a run.

Like this, you can simply specify `SSP2EU-AMT-Base` in your scenario config file and it will always pick the newest AMT run, which facilitates testing.

Output of start.R:
```
h_ndc_bIT
   Configuring cfg for h_ndc_bIT
   Use newest normally completed run for path_gdx_ref = SSP2EU-AMT-Base:
     /p/projects/remind/modeltests/output/SSP2EU-AMT-Base_2022-06-07_09.57.59
   Use newest normally completed run for path_gdx_refpolicycost = SSP2EU-AMT-Base:
     /p/projects/remind/modeltests/output/SSP2EU-AMT-Base_2022-06-07_09.57.59
   Use newest normally completed run for path_gdx_bau = SSP2EU-AMT-Base:
     /p/projects/remind/modeltests/output/SSP2EU-AMT-Base_2022-06-07_09.57.59
   Run can be started using 3 specified gdx file(s).
     input_ref.gdx: /p/projects/remind/modeltests/output/SSP2EU-AMT-Base_2022-06-07_09.57.59/fulldata.gdx
     input_refpolicycost.gdx: /p/projects/remind/modeltests/output/SSP2EU-AMT-Base_2022-06-07_09.57.59/fulldata.gdx
     input_bau.gdx: /p/projects/remind/modeltests/output/SSP2EU-AMT-Base_2022-06-07_09.57.59/fulldata.gdx
   Writing cfg to file h_ndc_bIT.RData
   If this wasn't --test mode, I would submit h_ndc_bIT.
```

For review, I suggest to use the ["hide whitespace" mode](https://github.com/remindmodel/remind/pull/853/files?diff=unified&w=1)


giannou (commented):
Very nice, thanks!
Two things:
- Will this work (or not crash) on a machine that does not see `/p/projects/remind/modeltests/output`?
- I am a bit puzzled by the `Normal completion` condition to identify successful runs: as this GDX will be used for fixings (or am I wrong?) it has to be a converged run, not simply one that simply completed the GAMS part (`Normal completion` might still be unconverged, no?)

orichters:
>     * Will this work (or not crash) on a machine that does not see `/p/projects/remind/modeltests/output`?

I tested it with `cfg$modeltest_folder <- "C:\Windows\` on the cluster: `Sys.glob("C:\Windows\…\fulldata.gdx")` or so simply returns `character(0)` in the same way as if you search for a scenario that does not exist. I tested it with  on the cluster and it did not crash :)
 
>     * I am a bit puzzled by the `Normal completion` condition to identify successful runs: as this GDX will be used for fixings (or am I wrong?) it has to be a converged run, not simply one that simply completed the GAMS part (`Normal completion` might still be unconverged, no?)

Agree on the second part, and if we can sharpen the condition further, I agree with you. Still, I hope nobody will ever use that for any production run but rather for quick tests, and I don't see a general problem why I should not use the `AMT-testOneRegi` run that is certainly not converged for fixing another testOneRegi run.

orichters:
Yes, the output of start.R is printed above, so people see it there, and `log.txt` also states the origin:
```
Try to copy /p/projects/remind/modeltests/output/SSP2EU-AMT-Base_2022-06-07_09.57.59/fulldata.gdx to input_ref.gdx.
Try to copy /p/projects/remind/modeltests/output/SSP2EU-AMT-Base_2022-06-07_09.57.59/fulldata.gdx to input_refpolicycost.gdx.
Try to copy /p/projects/remind/modeltests/output/SSP2EU-AMT-Base_2022-06-07_09.57.59/fulldata.gdx to input_bau.gdx.
```

giannou (approved):

orichters:
In the last force-push, I cleaned a bit and generalized the search for fully specified folders to AMT.

LaviniaBaumstark:
nice feature!
I only wonder if it might lead to many infes because of inconsistent scenarios which are used for fixing. But we can give it a try.

LaviniaBaumstark:
Maybe we should use a switch for this feature. We should protect production runs from using "wrong" gdx-files by chance and not seen by the user?

orichters:
> Maybe we should use a switch for this feature. We should protect production runs from using "wrong" gdx-files by chance and not seen by the user?

Note that the whole feature is only active if you choose a scenario name in your `path_gdx…` columns that is a title of an AMT scenario, such as as `SSP2EU-AMT-PkBudg670`, which shouldn't be done on production runs, and if you added that deliberately to test something with AMT runs, chances are you have the switch set to on as well.

What maybe better: I implemented [this summary](https://github.com/remindmodel/remind/blob/9da6c6f5b3ee65327d9886867aadfb4a9badf242/start_bundle_coupled.R#L567:L568) in `start_bundle_coupled.R`. Similarly, the start script could tell you at the end how many AMT runs he used for fixing, so people may notice that something is wrong if the are running production runs.

orichters:
Now, `start.R` prints to the user at the end:
```
Finished: 32 runs started. 61 runs are waiting. 3 GDX files from modeltests selected.
You are in --test mode. Rdata files were written, but no runs were started. 50 errors were identified.
```
I think I have to clean-up the NGFS scenario config file at some point :)

I also moved the "lock" warning to the bottom where people may actually read it.

LaviniaBaumstark (approved):

PR #857, opened 2022-06-10
fix R4.0 error that starts subsequent coupled runs with 1 CPU

orichters:
because strings are not read as factors anymore, `levels` returns `0` and the expression evaluates to `1` :(

also, fix that the model lock message and the counting of runs also works for restarting runs.

### DIFF
--- a/start.R
+++ b/start.R
@@ -297,6 +297,9 @@ modeltestRunsUsed <- 0
 
 testOneRegi_region <- ""
 
+# Save whether model is locked before runs are started
+model_was_locked <- file.exists(".lock")
+
 # Restart REMIND in existing results folder (if required by user)
 if (any(c("--reprepare", "--restart") %in% argv)) {
   # choose results folder from list
@@ -344,6 +347,7 @@ if (any(c("--reprepare", "--restart") %in% argv)) {
     cfg$remind_folder <- getwd()                      # overwrite remind_folder: run to be restarted may have been moved from other repository
     cfg$results_folder <- paste0("output/",outputdir) # overwrite results_folder in cfg with name of the folder the user wants to restart, because user might have renamed the folder before restarting
     save(cfg,file=paste0("output/",outputdir,"/config.Rdata"))
+    startedRuns <- startedRuns + 1
     if (! '--test' %in% argv) {
       submit(cfg, restart = TRUE)
     } else {
@@ -433,9 +437,6 @@ if (any(c("--reprepare", "--restart") %in% argv)) {
       message("\nYour slurmConfig selection will overwrite the settings in your scenario_config file.")
     }
   }
-
-  # Save whether model is locked before runs are started
-  model_was_locked <- if (file.exists(".lock")) TRUE else FALSE
 
   # Modify and save cfg for all runs
   for (scen in rownames(scenarios)) {
--- a/start_coupled.R
+++ b/start_coupled.R
@@ -313,7 +313,7 @@ start_coupled <- function(path_remind, path_magpie, cfg_rem, cfg_mag, runname, m
         # for the sbatch command set the number of tasks per node
         if (subseq.env$cfg_rem$gms$optimization == "nash" && subseq.env$cfg_rem$gms$cm_nash_mode == "parallel") {
           # for nash: set the number of CPUs per node to number of regions + 1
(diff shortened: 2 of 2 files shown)


start.R
```diff
@@ -297,6 +297,9 @@ modeltestRunsUsed <- 0
 
 testOneRegi_region <- ""
 
+# Save whether model is locked before runs are started
+model_was_locked <- if (file.exists(".lock")) TRUE else FALSE
```
dklein-pik, line 301: With the upcoming PR pik-piam/gms#23 the existence of the lock file will not be meaningful anymore. Once the PR is merged we need to change it here.

dklein-pik (approved):

start.R
pfuehrlich-pik, line 301:
```suggestion
model_was_locked <- file.exists(".lock")
```

start.R
```diff
@@ -344,6 +347,7 @@ if (any(c("--reprepare", "--restart") %in% argv)) {
     cfg$remind_folder <- getwd()                      # overwrite remind_folder: run to be restarted may have been moved from other repository
     cfg$results_folder <- paste0("output/",outputdir) # overwrite results_folder in cfg with name of the folder the user wants to restart, because user might have renamed the folder before restarting
     save(cfg,file=paste0("output/",outputdir,"/config.Rdata"))
+    startedRuns <<- startedRuns + 1
```
pfuehrlich-pik, line 350: Do we really need `<<-` here? This operator violates the usual R scoping rules so it should be avoided if possible. This line is not in a function, so replacing `<<-` with the normal `<-` should just work.

pfuehrlich-pik (approved):

start.R
orichters: Thanks, I will discuss with Mika!

orichters: haha, yes!

start.R
orichters: yes, I tested it. There are two other situations below and also in start_bundle_coupled.R where a similar call is indeed in a function – if there is a better solution, let me know.

start.R
```diff
@@ -297,6 +297,9 @@ modeltestRunsUsed <- 0
 
 testOneRegi_region <- ""
 
+# Save whether model is locked before runs are started
+model_was_locked <- file.exists(".lock")
+
```
mikapfl, line 302: This breaks with the new locking.

orichters: @mikapfl: I introduced this already [in March](https://github.com/remindmodel/remind/commit/a7c1600c41c78c4d81efc502cf29640cf3c440ad). Once the new locking is in place, we can simply delete it. The problem is that killing a job does not remove it from the queue, so the `.lock` file remains there even if no run is present that will unlock it at some point. The new lock mechanism doesn't have this problem, so we can simply delete it once it is implemented.

mikapfl: I see, it was introduced yesterday, and only moved in this commit. Anyway, it will break. model_lock and model_unlock don't guarantee that the lock file will be present when the model is locked, or that the lock file will be removed when the model is unlocked.

start.R
pfuehrlich-pik: A function should never change values in the calling/outer scope, but it can use the return value: Then the calling scope should analyze the return value and change variables in the calling scope accordingly.

start.R
orichters: No, it was introduced in March before we even started discussing the lock problem, then moved yesterday, and moved again today. I think "break" is a bit too hard :) It may give you inaccurate information if the `.lock` file is there although the model is using a different locking mechanism and it may not be locked in reality. Yes, let us not forget it, but currently I decided to merge to fix the actual bug that breaks all coupled runs…

mikapfl: well, all functionalities provided by `model_was_locked` (the information given to the user) will be inaccurate, so it breaks this thing, but it doesn't break the model. :wink: 

Sure, sorry, my `git blame` is weak and I started this discussion in the wrong PR/issue. As you said, it is not catastrophic to have wrong information for the user and your plan sounds good. But it is a bit unfortunate that things were implemented which circumvent the API. This model_lock/model_unlock API is there as an abstraction to separate concerns, and proper engineering would have been to introduce a `gms::model_locked` function (or something along those lines) and use that in the model so that all things related to locks are in the same library (and can e.g. be tested there).

PR #860, opened 2022-06-13
Update output slurm exceptions and plotIterations

robinhasse:
I extended the removal of low memory slurm options from @orichters to a formulation that allows to filter the slurm options for various output scripts with a regex. This allows to only consider direct mode for `plotIterations` as this script is interactive and does not work with slurm. It finishes in a couple of seconds and is not regularly used so making this script slurm-compatible does not seem necessary to me.
I also updated `plotIterations` to make the aesthetic mapping interactive and set the defaults according to our current (and probably exclusive) application of monitoring the price targets implementation by @Renato-Rodrigues.


output.R
```diff
@@ -139,11 +139,15 @@ choose_mode <- function(title = "Please choose the output mode") {
...
-  slurm_options <- c(if (! high_mem) c("--qos=priority", "--qos=short", "--qos=standby"),
+choose_slurmConfig_priority_standby <- function(title = "Please enter the slurm mode, uses priority if empty",
+                                                slurmExceptions = NULL) {
+  slurm_options <- c("--qos=priority", "--qos=short", "--qos=standby",
                      "--qos=priority --mem=8000", "--qos=short --mem=8000",
                      "--qos=standby --mem=8000", "--qos=priority --mem=32000", "direct")
+  if (!is.null(slurmExceptions)) {
+    slurm_options <- grep(slurmExceptions, slurm_options, value = TRUE)
+  }
+  if (length(slurm_options) == 1) return(slurm_options[[1]])
```
orichters, line 150: I like how you catch the length 1 case, but I wonder if for safety reasons, one should make sure that `direct` is always offered. So replace Line 148 with
```
slurm_options <- unique(c("direct", grep(slurmExceptions, slurm_options, value = TRUE))
```
This way, also the second comment would be fine…

orichters (changes requested):
looks fine, apart from getting the `direct` option back for `reporting`

robinhasse:
> looks fine, apart from getting the `direct` option back for `reporting`

You are right, that got lost. If you think, direct should never be removed, I adopt your proposal. Otherwise I could add it to the regex for `reporting` with a `or` connection.

orichters:
> > looks fine, apart from getting the `direct` option back for `reporting`
> 
> You are right, that got lost. If you think, `direct` should never be removed, I adopt your proposal. Otherwise I could add it to the regex for `reporting` with an `or` connection.

Yes, I think I would always offer `direct`. At the moment at least, I don't see a reporting where you should not be able to run it locally.

output.R
```diff
@@ -280,12 +284,16 @@ if (comp == "Exit") {
 
   # define slurm class or direct execution
   if (! exists("source_include")) {
-    modules_using_high_mem <- c("reporting")
-    need_high_mem <- isTRUE(any(modules_using_high_mem %in% output))
+    # for selected output scripts, only slurm configurations matching these regex are available
+    slurmExceptions <- switch(output,
+      reporting      = "--mem=[0-9]*[0-9]{3}",
```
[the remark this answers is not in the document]
robinhasse: `direct` is now always available following your suggestion.

output.R
robinhasse: `direct` is now always available following your suggestion.

orichters:
Sorry, I did an error: The line I proposed lacks a `)` at the end…

orichters (approved):
Fine from my side regarding the slurm choice. Better wait for Pascal's approval concerning the rest before merging.

pfuehrlich-pik (approved):

PR #863, opened 2022-06-14
Bugfix: Fix condition in $IF statement to correctly switch on/off sca…

dklein-pik:
…ling of landuse emissions to historical data.

# Purpose of this PR

## Checklist:
* updated in code documentation
* adjusted reporting if needed


dklein-pik:
> The commit message sounds like this didn't work in the past? Perhaps explain this more clearly, if the case (what implications). Else all good

Indeed it didnt work in the past. Until now the scaling was always switched on, even setting c_scaleEmiHistorical to "off" didnt have an effect.

I cant change the commit message ex post. So, I'm afraid I can't make it more clear than responding here?

PR #870, opened 2022-06-17
initialize modelHistExclude for backward compatibility

orichters:
## Purpose of this PR

#846 added modelHistExclude as an option, but if the script is called without it, it breaks. Initialize it to an empty vector for backward compatibility, it is overwritten if it is specified

## Checklist:
* no update in code documentation
* no adjusted reporting


fbenke-pik (approved):
Looks good to me. Thank you for finding an elegant way to achieve backward compatibility. Forgot to look into this.

PR #873, opened 2022-06-21
Add capacity variable without technology differentiation and set historic bounds on gas-power capacities in Germany

fschreyer:
Setting upper bound of gas capacity to 30 GW and 34 GW in 2015 and 2020 respectively to be in line with historic developments. 

Still waiting for test run to complete...


robertpietzcker (approved):
very useful

PR #874, opened 2022-06-22
new CES parameter and gdx files for SDP, SDP_EI, SDP_RC, SSP1, SSP2EU…

LaviniaBaumstark:
…, SSP2EU_lowEn and SSP5 for input data 6.311

# Purpose of this PR
* improve PR template
* new input data and CES parameters
## Checklist:
* updated in code documentation
* adjusted reporting if needed

### DIFF
--- a/pull_request_template.md
+++ b/pull_request_template.md
@@ -6,18 +6,21 @@
 - [ ] Bug fix 
 - [ ] Refactoring
 - [ ] New feature 
-- [ ] Minor change
+- [ ] Minor change (default scenarios show only small differences)
 - [ ] Fundamental change
 - [ ] This change requires a documentation update
 
 
 ## Checklist:
 
-- [ ] My code follows the coding ettiquette
+- [ ] My code follows the coding etiquette
 - [ ] I have performed a self-review of my own code
 - [ ] Changes are commented, particularly in hard-to-understand areas
-- [ ] I have updated the in code documentation
+- [ ] I have updated the in-code documentation
 - [ ] I have adjusted reporting where it was needed
 - [ ] The model compiles and starts successfully
+
+## Further information (optional):
+- Test runs are here: 
 
 
--- a/pull_request_template.md
+++ b/pull_request_template.md
@@ -1,6 +1,23 @@
 # Purpose of this PR
 
+
+## Type of change
+
+- [ ] Bug fix 
+- [ ] Refactoring
+- [ ] New feature 
+- [ ] Minor change
+- [ ] Fundamental change
+- [ ] This change requires a documentation update
+
+
 ## Checklist:
-* updated in code documentation
-* adjusted reporting if needed
-* checked that REMIND still compiles
+
+- [ ] My code follows the coding ettiquette
+- [ ] I have performed a self-review of my own code
+- [ ] Changes are commented, particularly in hard-to-understand areas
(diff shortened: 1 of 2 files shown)


pull_request_template.md
```diff
@@ -1,6 +1,23 @@
...
+- [ ] Fundamental change
+- [ ] This change requires a documentation update
+
+
 ## Checklist:
-* updated in code documentation
-* adjusted reporting if needed
-* checked that REMIND still compiles
+
+- [ ] My code follows the coding ettiquette
```
orichters, line 16: I think the "spelling etiquette" is with one t in etiquette

pull_request_template.md
```diff
@@ -1,6 +1,23 @@
...
+- [ ] I have performed a self-review of my own code
+- [ ] Changes are commented, particularly in hard-to-understand areas
+- [ ] I have updated the in code documentation
```
orichters, line 19: I would write "in-code documentation" like [github does](https://coderefinery.github.io/documentation/in-code-documentation/)

pull_request_template.md
```diff
@@ -1,6 +1,23 @@
...
+- [ ] I have adjusted reporting where it was needed
+- [ ] The model compiles and starts successfully
```
orichters, line 21: I usually added the path to the run so reviewers can have a look at the log if they have a suspicion.

pull_request_template.md
```diff
@@ -1,6 +1,23 @@
 # Purpose of this PR
 
+
+## Type of change
+
+- [ ] Bug fix 
+- [ ] Refactoring
+- [ ] New feature 
+- [ ] Minor change
```
orichters, line 9: for me, it isn't entirely clear what a "minor change" or "fundamental change" is, in particular as "bug fix" of "refactoring" are alternative options.

orichters (approved):
I added some minor comments

pull_request_template.md
LaviniaBaumstark: sometimes, user make local tests and external users will not be able to provide a path for everyone. I will add the option for a path to test runs

pull_request_template.md
LaviniaBaumstark: :-)

pull_request_template.md
LaviniaBaumstark: for me a "minor change" comes with less danger for the whole system. A "fundamental change" results in different results/model behaviour, etc.

orichters: then maybe add a short explanation or write it explicitly "minor change: does not alter model behavior". Because a bug fix could also result in different model results, but as I understand you now, the idea is that you can and should tick multiple boxes here. And of course, if it changes model behavior, one should state what it changes :)

pull_request_template.md
dklein-pik: What is the "default" scenario? Starting REMIND with the untouched default.cfg? I am not sure (1) that all people know this (2) that people will do this test runs before. They will probably have a test run with some configuration where they check the effects of their changes. Should they always to a "default" run?

LaviniaBaumstark: true, "default" might also change over time....in future, I expect that we will have a predefined set of test-scenarios. For now I would hope that developer to test runs - at least for SSP2 - if they introduce new features or make fundamental changes

PR #875, opened 2022-06-22
Add quick test run (-q) and help (-h) to start options

giannou:
# Purpose of this PR

- Adds the possibility to perform a very quick but full REMIND run as described [here](https://github.com/pik-piam/discussions/discussions/25). This workflow should be the last step before committing changes to the model (as discussed [here](https://github.com/remindmodel/remind/pull/874#discussion_r903372582) in #874 ). To perform this `quickRun` simply type `Rscript start.R -q`
- Adds a help message describing the available command-line options of `start.R`. To see this help type `Rscript start.R -h`

## Type of change

- [x] New feature 

## Checklist:

* Test runs are here: `/home/giannou/REMIND/output/testOneRegi_bu`
* Comparison of results (what changes by this PR?): nothing


start.R
```diff
@@ -454,14 +462,14 @@ if (any(c("--reprepare", "--restart") %in% argv)) {
       cfg$force_replace    <- TRUE
       if (testOneRegi_region != "") cfg$gms$c_testOneRegi_region <- testOneRegi_region
     }
-
+    if ("--quickRun" %in% argv) system("sed -i 's/rtredg = 1.d-9;/rtredg = 1.d-4;/' config/conopt3.op9")
```
orichters, line 465: How do you make sure this setting is reverted after the run has finished

start.R
```diff
@@ -327,7 +335,7 @@ if (any(c("--reprepare", "--restart") %in% argv)) {
     } else {
       if (! is.null(cfg[["backup"]][["cm_nash_mode"]])) cfg$gms$cm_nash_mode <- cfg$backup$cm_nash_mode
     }
-    if ("--testOneRegi" %in% argv) {
+    if (any(c("--quickRun", "--testOneRegi") %in% argv)) {
       if (is.null(cfg[["backup"]][["optimization"]])) cfg$backup$optimization <- cfg$gms$optimization
```
orichters, line 339: I would add `cfg$gms$cm_iteration_max <- 1` but only for quickRun to make sure it is always quick independent of the settings in default.cfg or scenario config file.

start.R
```diff
@@ -282,6 +285,11 @@ if (!all(known)) {
   config.file <- argv[!known][[1]] 
 }
 
+if ("--help" %in% argv) {
+  message(helpText)
+  q()
```
orichters, line 290: Didn't know `q()` was something one could use in a .R file...

orichters (commented):
I really like it, that is very helpful. Did you test it with a scenario config as well? I have no doubt that it works, though.

I think in [tutorial 2](https://github.com/remindmodel/remind/blob/develop/tutorials/2_RunningREMIND.md) and [tutorial 3](https://github.com/remindmodel/remind/blob/develop/tutorials/3_RunningBundleOfRuns.md), I would replace testOneRegi by quickRun, maybe check other tutorials as well, except for the debugging one.

start.R
orichters: Not sure this works with this line, though:  https://github.com/remindmodel/remind/blob/develop/scripts/start/prepare_and_run.R#L488
The rest seems fine

start.R
```diff
@@ -23,6 +24,8 @@ require(stringr)
 #'
 #' Control the script's behavior by providing additional arguments:
 #'
+#' --help, -h: show this text and exit
+#'
 #' --debug, -d: start a debug run with cm_nash_mode = debug
 #'
 #' --interactive, -i: interactively select run(s) to be started
```
orichters, line 31: the new `-q` option seems to be missing here in the help text. And I would rename it to simply `quick`. What else should it start if not a `run`?

start.R
giannou: I added a new conopt3 options file to avoid these issues

giannou:
> Did you test it with a scenario config as well?

What is supposed to happen there, and how do I do it? And what is the use-case? :)

giannou:
Minor issue: with the new PR template an uncomplete tasks list is shown in the PR list of the model, even when someone has completed their tasks #876

tutorials/3_RunningBundleOfRuns.md
```diff
@@ -64,7 +64,7 @@ Rscript start.R --test config/scenario_config_XYZ.csv
 ```
 Running the complete chain of runs, but only for one region and one iteration (as long as `cm_iteration_max = 1` wasn't changed), can be started with:
```
orichters, line 65: you can delete "(as long as `cm_iteration_max = 1` wasn't changed)" because it is set to 1 always with `-q`

giannou: done, thanks!

giannou:
> Minor issue: with the new PR template an uncomplete tasks list is shown in the PR list of the model, even when someone has completed their tasks #876

Well, one can also remove the options not chosen and have the correct overview shown. So feel free to close the issue

orichters (approved):
Great work! I tested it, running a default run and the NGFS set of runs using the `quick` option. The NGFS scenarios went through without a problem. While `default` and `SSP2-Base` runs in 3 minutes (very good!), the other runs took something between 5 and 13 minutes gams_runtime. This is very useful if you want to test the scripts for subsequent runs etc., or have a quick check if your settings work.

Minor detail: [here in choose_slurmConfig.R](https://github.com/remindmodel/remind/blob/9f39662e85623c190293cc888d13a4cc21146f40/scripts/start/choose_slurmConfig.R#L31-L40) you could replace `reporting` which is not needed as `output.R` doesn't call this function anymore anyway by `quick` three times.

What is also not working I suspect is to coherently combine `--reprepare` with `--quick`: `--quick` and `--testOneRegi` don't differ in their behavior in this case, so you don't get faster convergence. But: If you start something with `--testOneRegi` and `--reprepare`, it saves all the settings and if you restart the same run without `--testOneRegi`, the initial settings are restored. If you wanted to implement that with `--quick` as well, it would be more work I guess.

There seem to be two option: Implement `--quick` together with `--reprepare`, but then you would need to add `--quick` also to [these lines](https://github.com/remindmodel/remind/blob/c33bb95392122b7eae02a2053ece877066e51ca6/start.R#L295-L297) as `--restart` and `--quick` are incompatible, you have to use `--reprepare` to copy the files etc.

An alternative would be to simply completely disallow combining `--quick` and `--restart`/`--reprepare` there, that would be perfectly fine for me. In a similar way as it is the case now, you could ask whether you want to use `--testOneRegi` instead.

Anyway, these are probably minor details overall, so I approve the PR.

giannou:
Thanks! I think I'll go ahead and merge as the start scripts will be reworked anyhow and hopefully then incompatibilities will be treated in a more systematic way.

PR #877, opened 2022-06-23
Reactivate global couple products for CHP technologies

merfort:
# Purpose of this PR
This PR reactivates the global fallback values for couple product (heat) conversion efficiencies of CHP technologies. Due to inconsistencies in the input data REMIND was infeasible for historic time steps, when some in regions CHP technologies produced heat as a secondary product, so as a quick fix the coupling products were disabled. [This bugfix](https://github.com/pik-piam/mrremind/pull/272) in mrremind corrects the inconsistencies. 

The new corrected input data also has the effect that now in all regions for all CHP technologies a region-specific couple product efficiency can be derived. Therefore reactivating the global fallback values (that will only become active if no region-specific value is available) will have no effect at the moment. However, since input data may change again, we nevertheless activate the global fallback values.

## Type of change

- [x] Bug fix 

## Checklist:

- [x] Changes are commented, particularly in hard-to-understand areas (not applicable)
- [x] I have updated the in-code documentation (not applicable)
- [x] I have adjusted reporting where it was needed (not applicable)
- [x] The model compiles and starts successfully

### DIFF
--- a/modules/04_PE_FE_parameters/iea2014/datainput.gms
+++ b/modules/04_PE_FE_parameters/iea2014/datainput.gms
@@ -236,11 +236,11 @@ loop(pc2te(enty,enty2,te,enty3),
 display pm_prodCouple;
 
 *** define global values for couple production that can be used if the regional IEA data are 0
-***p04_prodCoupleGlob("pecoal","seel","coalchp","sehe")        = 0.61;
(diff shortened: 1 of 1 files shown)


LaviniaBaumstark:
does it mean that the current input data already contain the regional data so that it already has no effect?

merfort:
Exactly

Issue #834, opened 2022-05-25
Ideas to improve start scripts for coupled runs


orichters:
These are ideas to improve or adapt the start scripts for coupled runs, collected while discussing PR #794.

- the `sbatch` command used to start subsequent runs is added to the logs. It can be used to restart a run, also by a different user if that is desired. But: Say user1 starts the whole bundle. After Base-rem-1 is finished, NDC-rem-1 is started automatically. You kill it and ask user2 to start it this way. Now, NDC-rem-1 finishes quicker than Base-rem-2. The script in NDC-rem-1 that starts subsequent runs realizes that Base-rem-2 isn't finished and doesn't start NDC-rem-2. Once Base-rem-2 is finished, it starts NDC-rem-2. But now, it is again on the account of user1. Maybe it is worth adapting the script such that `rem-1` until `rem-x` all remain on the same user account

- In parallel mode, subsequent runs could potentially be started before running MAgPIE, or the MAgPIE report of the Base run could also be fed into the subsequent run, or it could be left as is. Another option would be to change the naming such that `mag-2` is run before `rem-2` (and `mag-1` wouldn't exist). This would make it maybe a bit easier, because `mag-5` and `rem-5` would be matched, not `rem-5` and `mag-4` as of now.

- confusion and errors may still arise if `start_bundle_coupled.R` is run while runs are already running. You certainly should simply not do that. Alternative would be a check for existing runs (or something like a lock file in the output folder).

- for parallel mode, the log could be written to the actual output subfolder. But this is impossible as the output folder will be deleted later by `start_coupled.R`. Somehow, this had to be changed.

PR #878, opened 2022-06-23
adapt to new locking method

orichters:
## Purpose of this PR

- Adapt messages for users to new locking mechanism
- delete Sys.sleep(10) which was a safety mechanism against locking problems for subsequent runs

## Type of change

- adapt messages for users to new locking introduced with https://github.com/pik-piam/gms/pull/23
- to make sure there is backward compatibility, check whether [is_model_locked](https://github.com/pik-piam/gms/blob/master/R/is_model_locked.R) exists

## Checklist:

- [x] follows the R coding etiquette
- Changes are not commented
- I have not updated the in-code documentation
- I have not adjusted reporting
- [x] The model compiles and starts successfully: I did checks with the new locking and the old locking mechanism in the same folder.


start_coupled.R
```diff
@@ -327,7 +327,6 @@ start_coupled <- function(path_remind, path_magpie, cfg_rem, cfg_mag, runname, m
         message(subsequentcommand)
         if (length(needfulldatagdx) > 0) {
           system(subsequentcommand)
-          Sys.sleep(10)
```
mikapfl, line 330: I'm not quite sure if we should remove the sleep already. If people are running current REMIND develop with old snapshots with the old locking, this will make it more likely that things break, right?

mikapfl (approved):
Looks pretty good, just one question about timing and impact of the changes.

start_coupled.R
orichters: agreed. I undid it and added it our wishlist so I will not forget it: https://github.com/remindmodel/remind/issues/834

PR #881, opened 2022-06-24
overwrite mif file with new policyCosts, provide information whether mif/gdx files exist

orichters:
## Purpose of this PR

- instead of writing a new  `_adjustedPolicyCosts.mif`, overwrite the usual file, making post-processing easier. Also improve used feedback with coloring dependent on whether the mif/gdx files exist. Helpful if the script is called via command line…
- don't let output.R write "starting output generation" if there is no output to generate (testOneRegi, quick mode)

## Type of change

- [x] faciliate post-processing
- [x] better user communication

## Checklist:

- [x] I tested the script and it runs successfully

### DIFF
--- a/scripts/output/comparison/policyCosts.R
+++ b/scripts/output/comparison/policyCosts.R
@@ -120,9 +120,9 @@ write_new_reporting <- function(mif_path,
                                 scen_name,
                                 new_polCost_data) {
 
-  new_mif_path <- paste0(substr(mif_path,1,nchar(mif_path)-4),"_adjustedPolicyCosts.mif")
+  new_mif_path <- mif_path # paste0(substr(mif_path,1,nchar(mif_path)-4),"_adjustedPolicyCosts.mif")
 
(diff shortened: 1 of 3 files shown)


scripts/output/comparison/policyCosts.R
```diff
@@ -120,9 +120,9 @@ write_new_reporting <- function(mif_path,
                                 scen_name,
                                 new_polCost_data) {
 
-  new_mif_path <- paste0(substr(mif_path,1,nchar(mif_path)-4),"_adjustedPolicyCosts.mif")
+  new_mif_path <- mif_path # paste0(substr(mif_path,1,nchar(mif_path)-4),"_adjustedPolicyCosts.mif")
 
-  message("A mif file with the name ", crayon::green(paste0("REMIND_generic_",scen_name,"_adjustedPolicyCosts.mif"))," is being created in the ",scen_name," output folder.")
+  message("A mif file with the name ", crayon::green(new_mif_path), " is written in the ",scen_name," output folder.")
```
johanneskoch94, line 125: I guess the message could be changed to something like "The mif file will be overwritten...", but it's just cosmetics :)

johanneskoch94 (approved):
Looks good! See my one comment - if you feel like it

PR #885, opened 2022-06-29
Remove two superfluous inclusions of n2obio

dklein-pik:
# Purpose of this PR

N2O emissions from biomass were included three times. Remove two obsolete calculations. Closes remindmodel/development_issues#106

## Type of change

- [X] Bug fix 
- [ ] Refactoring
- [ ] New feature 
- [X] Minor change (default scenarios show only small differences)
- [ ] Fundamental change

## Checklist:

* Test runs are here: 


core/preloop.gms
```diff
@@ -137,6 +137,16 @@ p_agriEmiPhaseOut(t)$(t.val ge 2040) = 1;
...
   = (1-p_agriEmiPhaseOut(t)*c_BaselineAgriEmiRed)*pm_macBaseMagpie(t,regi,enty);
+  
+*** The N2O emissions generated during biomass production in agriculture (in MAgPIE)
+*** are represented in REMIND by applying the n2obio emission factor (zero in coupled runs)
+*** in q_macBase. In standaolne runs the resulting emissions need to be subtracted (see below)
+*** from the exogenous emission baseline read from MAgPIE, since the baseline already implicitly 
+*** includes the N2O emissions from biomass. In q_macBase in core/equations.gms the N2O 
+*** emissions resulting from the actual biomass demand in REMIND are then added again. 
+
+pm_macBaseMagpie(t,regi,"n2ofertin") = pm_macBaseMagpie(t,regi,"n2ofertin") - (p_efFossilFuelExtr(regi,"pebiolc","n2obio") * pm_pebiolc_demandmag(t,regi));
```
robertpietzcker, line 148: I am not sure I fully understand this (always only talking about the REMIND standalone version)

Is pm_pebiolc_demandmag updated over REMIND iterations? If yes, it seems that the Magpie baseline is updated each iteration with the result that total fertilizer emissions become independent of the biomass use - which to me seems wrong. 

Shouldn't only baseline biomass emissions be subtracted from the Magpie emissions, which should be fixed and not be adjusted over iterations?
so that in a standalone REMIND run with lots of biomass use, total fertilizer emissions actually do go up. 
(But maybe I misunderstand these lines)

core/preloop.gms
```diff
@@ -137,6 +137,16 @@ p_agriEmiPhaseOut(t)$(t.val ge 2040) = 1;
 *** Rescale non-co2 base line emissions from agriculture 
 pm_macBaseMagpie(t,regi,enty)$(emiMac2sector(enty,"agriculture","process","ch4") OR emiMac2sector(enty,"agriculture","process","n2o"))
   = (1-p_agriEmiPhaseOut(t)*c_BaselineAgriEmiRed)*pm_macBaseMagpie(t,regi,enty);
+  
+*** The N2O emissions generated during biomass production in agriculture (in MAgPIE)
+*** are represented in REMIND by applying the n2obio emission factor (zero in coupled runs)
+*** in q_macBase. In standaolne runs the resulting emissions need to be subtracted (see below)
+*** from the exogenous emission baseline read from MAgPIE, since the baseline already implicitly 
+*** includes the N2O emissions from biomass. In q_macBase in core/equations.gms the N2O 
+*** emissions resulting from the actual biomass demand in REMIND are then added again.
```
merfort, line 146: I tried to make this explanation a bit clearer. 
```suggestion
*** The N2O emissions generated during biomass production in agriculture (in MAgPIE)
*** are represented in REMIND by applying the n2obio emission factor (zero in coupled runs)
*** in q_macBase. In standalone runs the resulting emissions associated with baseline bioenergy 
*** production (pm_pebiolc_demandmag) need to be subtracted (see below) from the 
*** exogenous emission baseline read from MAgPIE, since the baseline already implicitly 
*** includes the N2O emissions from biomass. In q_macBase in core/equations.gms the N2O 
*** emissions resulting from the actual biomass demand in REMIND are then added again. 
```

core/preloop.gms
dklein-pik: 0. Yes, standalone only. 
2. It is in preloop, so not within the loop.
3. `pm_pebiolc_demandmag` is an exogenous parameter that is *not* dependent on model behaviour. According to the current REMIND scenario is selected from a lookup table providing different biomass demands for different SSPs and RPCs.

merfort:
I think it would be good, to also add some comments to the description of `q_macBase`. In particular, I think it would be good to explain that in case of coupled runs `pm_macBaseMagpie` contains all N2O emissions (directly from MAgPIE) and that `p_efFossilFuelExtr(regi,"pebiolc","n2obio")` is zero then. Otherwise this might lead to confusion in the future.

core/preloop.gms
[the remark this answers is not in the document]
dklein-pik: The emission factor is zero in coupled runs https://github.com/remindmodel/remind/blob/25c70e47d678b4bf69e1708bc7ad739cda109f3b/core/datainput.gms#L739 But we can keep your changes if you want.

core/input/generisdata_emi.prn
```diff
@@ -34,16 +34,16 @@ pecoal.seliqfos.coalftrec      26.1
 pecoal.seliqfos.coalftcrec      7.8      18.3
 pecoal.sesofos.coaltr          26.1             0.04415
 pecoal.segafos.coalgas         26.1             0.04415
-pebiolc.seel.bioigccc         -20.0      20.0   0.12614
-pebiolc.seel.bioigcc                            0.12614
+pebiolc.seel.bioigccc         -20.0      20.0
```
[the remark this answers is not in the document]
dklein-pik: Done

core/preloop.gms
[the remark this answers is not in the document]
merfort: Ah, right. Then let's keep it as it is and just add a comment above.

merfort (approved):
From my side, this looks good now, thanks David!

core/preloop.gms
robertpietzcker: thanks for the explanation!

PR #893, opened 2022-07-05
add --reset/-0 to start.R to reset main.gms to default.cfg

orichters:
## Purpose of this PR

As my [suggestions to update a tmp file instead of main.gms](https://github.com/remindmodel/remind/pull/740) do not seem to please and further progress related to start scripts stalled, add an option to reset `main.gms` manually with `./start.R --reset` or `./start.R -0`. Only the `Last modification (input data): Thu Jun 23 10:11:30 2022` string remains (which strangely enough does not show input data changes, but is rather always set to the current time when prepare_and_run.R is executed). If `main.gms` was manually edited to add a switch, this remains and is set to the value specified in `config/default.cfg`.

## Type of change

- [x] New minor feature, does not affect scenarios at all

## Checklist:

- [x] My code follows the coding etiquette
- [x] I have performed a self-review of my own code
- [x] Changes are commented, particularly in hard-to-understand areas
- [x] I have updated the in-code documentation

PR #895, opened 2022-07-05
add lock to --reset

orichters:
## Purpose of this PR

add locking to the `--reset` flag introduced in #893 such that it does not interfere with model runs.

## Type of change

- [x] Bug fix

## Checklist:


start.R
```diff
@@ -304,8 +304,12 @@ if ("--reset" %in% argv) {
   source("./config/default.cfg")
   cfg$gms$c_expname <- cfg$title
   cfg$gms$c_description <- substr(cfg$description, 1, 255)
+  lock_id <- gms::model_lock(timeout1 = 1)
```
mikapfl, line 307: Minor thing: shouldn't the timeout be shorter? `--reset` is an interactive operation, no?

mikapfl (approved):
Looks good.

orichters:
I reduced the timeout by 80% to 12 minutes. So if you start it while some runs are still waiting, that is ok, but it doesn't wait that long. Hope that is ok as a compromise.

PR #904, opened 2022-07-15
fix wrong ref scenario for NGFS integrated damage runs

orichters:
## Purpose of this PR

fix reference error in NGFS integrated damage scenarios

## Type of change

- Bug fix without any impact on the rest of REMIND


christophbertram (approved):
Indeed, a tricky question

PR #906, opened 2022-07-18
correct slurmConfig messaging for --quick

orichters:
## Purpose of this PR

- If combining `--quick` with a scenario config, it tells you it uses the slurmConfig from that file, while in fact it doesn't. This PR adapts the messaging to reality.
- reduce slurm time to 1 hour to get quicker access on non-priority qos

## Type of change

- [x] minor bug fix that does not affect any scenario


giannou (approved):
Nice, and thanks for the introduction of the time limit

PR #908, opened 2022-07-19
create joint REMIND and MAgPIE reporting for every coupling round

orichters:
## Purpose of this PR

- until now, `start_coupled.R` produces a joint reporting `./output/C_SSP2-Base.mif` after the final -rem- run (say: -rem-5).
- after this PR, a joint mif file is created after every coupling round (if a MAgPIE report exists), so for -rem-2 with -mag-1, for -rem-3 with -mag-2 etc. The model names are kept to `REMIND` and `MAgPIE` to make it distinguishable
- this file can then be used to plot MAgPIE variables also in compareScenarios2
- the `./output/C_SSP2-Base.mif` will still be written after the last iteration, containing a joint reporting with `REMIND-MAgPIE` as model name
- I moved it before the starting of subsequent runs to avoid that the mif file is rewritten while read by some other script (should not happen, though, as subsequent runs read only gdx files, as far as I know)
- I also adapted the code a bit in plot_compare_iterations because the mbind command took very long if one run had the MAgPIE part in the mif and the other not.
- you can have a look at `/p/tmp/oliverr/remind-smallfix/output/C_o_1p5c_bIT.mif`, also the corresponding -rem- folders, and `compScen-rem-1-5_C_o_1p5c_bIT.pdf`

## Type of change

- [x] New feature 
- [x] Minor change (no impact on scenarios)

## Checklist:

- [x] I tested the NGFS scenarios using [this quick setup](https://github.com/pik-piam/discussions/discussions/28) and it works well. Also cs2 worked.

### DIFF
--- a/scripts/output/comparison/plot_compare_iterations.R
+++ b/scripts/output/comparison/plot_compare_iterations.R
@@ -227,8 +227,10 @@ plot_iterations <- function(runname) {
   report_path <- report_path[!grepl("with|adj",report_path)]
 
   tmp <- NULL
-  for (r in report_path) tmp <- mbind(tmp, read.report(r,as.list=FALSE))
-  tmp <- tmp[,,"Price|Carbon (US$2005/t CO2)"]
+  for (r in report_path) {
+    tmp1 <- read.report(r, as.list=FALSE)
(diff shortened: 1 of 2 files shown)


merfort (approved):
Looks all good to me, thanks!

PR #928, opened 2022-08-01
Setting default of cm_abortOnConsecFail to 5.

PhilippVerpoort:
# Purpose of this PR
As discussed many times within the group, we would like to turn this switch on by default so that newbies don't constantly run into arbitrary REMIND errors due to a series of consecutive INFES early in the REMIND iteration.

## Type of change

- [x] New feature


orichters (approved):
thanks, as discussed.

Issue #929, opened 2022-08-02
consolidate R script functions


0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
Originated from [this discussion](https://mattermost.pik-potsdam.de/rd3/channels/rse/xwy948yap7dhmp8ae35txankih) initiated by @chroetz.

There seems to be a lot potential for consolidating utility functions in the R scripts used in the REMIND repository.  "Choose one item from a list" is already being discussed (see above), `/get-?[lL]ine/` is another candidate.

```
 ag --r --nogroup "<- function" | tr ':' ' ' | awk '{ print $3, $1}' | sort -k 1,1 | column -t
addyrs                               scripts/output/single/notebook_templates/EDGETransportMultiIterationAnalysis.Rmd
.agriculture_costs                   scripts/start/prepare_and_run.R
.agriculture_tradebal                scripts/start/prepare_and_run.R
.bioenergy_costs                     scripts/start/prepare_and_run.R
.bioenergy_price                     scripts/start/prepare_and_run.R
.bioenergy_production                scripts/start/prepare_and_run.R
choose_filename_prefix               output.R
choose_folder                        output.R
chooseFromList                       start.R
choose_mode                          output.R
choose_module                        output.R
choose_slurmConfig                   scripts/start/choose_slurmConfig.R
choose_slurmConfig_priority_standby  output.R
combine_slurmConfig                  scripts/start/choose_slurmConfig.R
compareScenTable                     scripts/output/comparison/multiComparison.R
configure_cfg                        start.R
.copy.fromlist                       scripts/start/prepare_and_run.R
.copy.fromlist                       scripts/start/submit.R
create_fixing_files                  scripts/start/prepare_and_run.R
create_standard_fixings              scripts/start/prepare_and_run.R
datapath                             scripts/iterative/EDGE_transport.R
debug_coupled                        start_coupled.R
didremindfinish                      start.R
.emissions_mac                       scripts/start/prepare_and_run.R
.escape                              scripts/insertheader/insertheader.R
fill_missing_scenarios               scripts/output/comparison/bioenergy_SSP.R
fillyears                            scripts/output/comparison/plot_compare_iterations.R
.findheader                          scripts/insertheader/insertheader.R
get_demand                           scripts/output/single/notebook_templates/EDGETransportMultiIterationAnalysis.Rmd
.getExtension                        scripts/insertheader/insertheader.R
[...]

PR #930, opened 2022-08-02
Use gms::getLine() instead of re-defining it everywhere

mikapfl:
# Purpose of this PR

Instead of defining `getLine` or variants of it in every script, use the version from `gms`. This is not a new dependency, we depend on `gms` anyway in REMIND. See discussion in #929.

## Type of change

- [x] Refactoring

## Checklist:

### DIFF
--- a/output.R
+++ b/output.R
@@ -41,19 +41,6 @@ if (file.exists("/iplex/01/landuse")) { # run is performed on the cluster
   latexpath <- NA
 }
 
-get_line <- function() {
-  # gets characters (line) from the terminal of from a connection
-  # and stores it in the return object
-  if (interactive()) {
-    s <- readline()
-  } else {
-    con <- file("stdin")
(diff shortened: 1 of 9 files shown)


output.R
```diff
@@ -41,19 +41,6 @@ if (file.exists("/iplex/01/landuse")) { # run is performed on the cluster
...
-  if (interactive()) {
-    s <- readline()
-  } else {
-    con <- file("stdin")
-    s <- readLines(con, 1, warn = FALSE)
-    on.exit(close(con))
-  }
-  return(s)
-}
-
```
0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q, line 56: Does `gms::getLine()` behave the same as this code in both interactive and batch mode?

mikapfl: Yes.

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q (approved):

PR #935, opened 2022-08-10
adapt output.R, start.R and cs2 to new gms::chooseFromList()

orichters:
## Purpose of this PR

- `chooseFromList` was moved to `gms` in this PR: https://github.com/pik-piam/gms/pull/32
- various similar functions are replaced by a call to `gms::chooseFromList()`
- part of efforts to [consolidate R script functions](https://github.com/remindmodel/remind/issues/929)

## Type of change

- [x] Refactoring

## Checklist:

### DIFF
--- a/output.R
+++ b/output.R
@@ -41,95 +41,7 @@ if (file.exists("/iplex/01/landuse")) { # run is performed on the cluster
   latexpath <- NA
 }
 
-choose_folder <- function(folder, title = "Please choose a folder") {
-  dirs <- NULL
-
-  # Detect all output folders containing fulldata.gdx
-  # For coupled runs please use the outcommented text block below
-
-  dirs <- basename(dirname(Sys.glob(file.path(folder, "*", "fulldata.gdx"))))
-
-  # DK: The following outcommented lines are specially made for listing results of coupled runs
-  # runs <- findCoupledruns(folder)
-  # dirs <- findIterations(runs,modelpath=folder,latest=TRUE)
-  # dirs <- sub("./output/","",dirs)
-
-  dirs <- c("all", dirs)
-  cat("\n\n", title, ":\n\n")
-  cat(paste(seq_along(dirs), dirs, sep = ": "), sep = "\n")
-  cat(paste(length(dirs) + 1, "Search by the pattern.\n", sep = ": "))
-  cat("\nNumber: ")
-  identifier <- getLine()
-  identifier <- strsplit(identifier, ",")[[1]]
-  tmp <- NULL
-  for (i in seq_along(identifier)) {
-    if (length(strsplit(identifier, ":")[[i]]) > 1) {
-      tmp <- c(tmp, as.numeric(strsplit(identifier, ":")[[i]])[1]:as.numeric(strsplit(identifier, ":")[[i]])[2])
-    } else {
-      tmp <- c(tmp, as.numeric(identifier[i]))
-    }
-  }
-  identifier <- tmp
-  # PATTERN
-  if (length(identifier) == 1 && identifier == (length(dirs) + 1)) {
-    cat("\nInsert the search pattern or the regular expression: ")
-    pattern <- getLine()
-    id <- grep(pattern = pattern, dirs[-1])
-    # lists all chosen directories and ask for the confirmation of the made choice
-    cat("\n\nYou have chosen the following directories:\n")
-    cat(paste(seq_along(id), dirs[id + 1], sep = ": "), sep = "\n")
-    cat("\nAre you sure these are the right directories?(y/n): ")
-    answer <- getLine()
-    if (answer == "y") {
-      return(dirs[id + 1])
-    } else {
-      choose_folder(folder, title)
-    }
-  } else if (any(dirs[identifier] == "all")) {
-    identifier <- 2:length(dirs)
-    return(dirs[identifier])
-  } else {
-    return(dirs[identifier])
-  }
-}
-
-
-choose_module <- function(Rfolder, title = "Please choose an outputmodule") {
-  module <- gsub("\\.R$", "", grep("\\.R$", list.files(Rfolder), value = TRUE))
-  cat("\n\n", title, ":\n\n")
-  cat(paste(seq_along(module), module, sep = ": "), sep = "\n")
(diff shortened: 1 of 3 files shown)


output.R
```diff
@@ -41,113 +41,17 @@ if (file.exists("/iplex/01/landuse")) { # run is performed on the cluster
...
-  } else {
-    stop("This mode is invalid. Please choose a valid mode.")
-  }
-  return(comp)
-}
-
-choose_slurmConfig_priority_standby <- function(title = "Please enter the slurm mode, uses the first option if empty",
-                                                slurmExceptions = NULL) {
+choose_slurmConfig_output <- function(slurmExceptions = NULL) {
   slurm_options <- c("--qos=priority", "--qos=short", "--qos=standby",
```
orichters, line 45: - I kept choose_slurmConfig because it also defines the list etc.

output.R
```diff
@@ -161,36 +65,37 @@ choose_filename_prefix <- function(modules, title = "") {
 }
 
 if (exists("source_include")) {
-  comp <- FALSE
-} else if (!exists("comp")) {
-  comp <- choose_mode("Please choose the output mode")
+  comp <- "single"
```
orichters, line 68: `choose_mode` was completely replaced by some pre- and postprocessing and a call to `chooseFromList`

output.R
```diff
@@ -161,36 +65,37 @@ choose_filename_prefix <- function(modules, title = "") {
...
-if (comp == "Exit") {
-  q()
-} else if (comp == "comparison" | comp == "export") {
-  print("comparison")
-  # Select output modules if not defined by readArgs
-  if (!exists("output")) {
-    output <- choose_module(paste0("./scripts/output/", comp),
-                            "Please choose the output module to be used for output generation")
-  }
+if (! exists("output")) {
```
orichters, line 77: all `single`, `comparison` and `exit` share one call to select module now.

output.R
```diff
@@ -161,36 +65,37 @@ choose_filename_prefix <- function(modules, title = "") {
...
-      outputdirs <- temp
-      for (i in seq_along(temp)) {
-        outputdirs[i] <- file.path("output", temp[i])
-      }
-    } else {
-      temp <- choose_folder(remind_dir, "Please choose the runs to be used for output generation, separate with comma")
-      outputdirs <- temp
+    dir_folder <- if (exists("remind_dir")) remind_dir else "./output"
+    dirs <- basename(dirname(Sys.glob(file.path(dir_folder, "*", "fulldata.gdx"))))
+    # DK: The following outcommented lines are specially made for listing results of coupled runs
```
orichters, line 91: I moved that from the function call, maybe @dklein-pik can tell whether this is still in use.

scripts/output/comparison/compareScenarios2.R
```diff
@@ -116,10 +86,15 @@ profiles <- remind2::getCs2Profiles()
 
 # Let user choose cs2 profile(s).
 profileNamesDefault <- determineDefaultProfiles(outputdirs[1])
-profileNames <- chooseFromSequence(
-  names(profiles),
-  "Choose profiles for cs2.",
-  profileNamesDefault)
+profileNames <- gms::chooseFromList(
+  ifelse(names(profiles) %in% profileNamesDefault, crayon::cyan(names(profiles)), names(profiles)),
```
chroetz, line 90: With this code the resulting `profileNames` may contain the crayon markup. Despite that it works on the cluster, as  `profileNames` is used to create a slurm command that is then executed via `system()`. This seems to "clear" the crayon markup. But for local use it probably will crash. So we need to use `returnBoolean` here, I guess

chroetz (changes requested):
Thanks! I tested the cs2 choices. It works, but it still needs a small fix, see comment in code.

scripts/output/comparison/compareScenarios2.R
orichters: You are right, I did correct that.

output.R
mikapfl: If it is still in use, it should be moved into a proper `if` with a command-line switch or something. Any commented-out code which is in use is a pretty serious bug, imo. But that is of course out of scope for this PR.

output.R
```diff
@@ -161,36 +65,37 @@ choose_filename_prefix <- function(modules, title = "") {
...
+    # runs <- findCoupledruns(folder)
+    # dirs <- findIterations(runs,modelpath=folder,latest=TRUE)
+    # dirs <- sub("./output/","",dirs)
+    temp <- chooseFromList(dirs, type = "runs to be used for output generation", returnBoolean = FALSE,
+                                      multiple = TRUE)
+    outputdirs <- file.path("output", temp)
```
mikapfl, line 97: Please use a better name than `temp`. I know you didn't invent it, but it gets even harder to understand what `temp` is now that the function call is a generic `chooseFromList` instead of `choose_folder`. Maybe `folders` would be a good name instead of `temp`?

output.R
```diff
@@ -242,24 +147,19 @@ if (comp == "Exit") {
...
```
mikapfl, line 161: duplicated code - maybe move to a separate function after all?

mikapfl (approved):
Thank you for this cleanup! Looks good, I'd maybe suggest de-duplicating a rather small part of code into a shared function and better variable naming. But it is a net-positive as is, I'm also fine with merging straight away.

output.R
mikapfl: or `selected_folders`

output.R
orichters: I checked what this code does: It only selects coupled runs. I replaced it with a different solution: the `rem-[1-9]` part is extracted and used as a group index. So you can easily select all `rem-5` runs with one number.

output.R
orichters: I used selectedDirs

output.R
orichters: wasn't even necessary. Was a completely redundant structure:
```
if () {
  thecode
  ...
} else {
  thecode
  ...
}
```
So simply moved it before the if.

output.R
mikapfl: ah, right, it seems I will never use camelCase on first try, always get caught by the linter. :rofl:

PR #937, opened 2022-08-15
Fix output.R: object 'outputUsingDirect' not found

chroetz:
# Purpose of this PR

output.R failed when sourced from another script with error message `object 'outputUsingDirect' not found`

This is solved by putting the definition of `outputUsingDirect` where it can also be seen when output.R is called directly and when it is sourced by another script.

## Type of change

- [X] Bug fix

## Checklist:


mikapfl (approved):
Looks good to me.

PR #941, opened 2022-08-16
adds some documentation

chroetz:
# Purpose of this PR

Add documentation about Units of certain variables, in particular related to DAC.

## Type of change

- [x] Refactoring

## Checklist:

- [X] I have updated the in-code documentation

### DIFF
--- a/core/input/generisdata_tech.prn
+++ b/core/input/generisdata_tech.prn
@@ -1,7 +1,8 @@
 *** SOF ./core/input/generisdata_tech.prn
 $ontext
 IMPORTANT: all costs are now given in $2015! As long as the model runs in $2005, the values need to be converted in datainput.gms
-inco0       Initial investment costs given in $(2015)/kW(output) capacity
+tech_stat   technology status: how close a technology is to market readiness. Scale: 0-3, with 0 'I can go out and build a GW plant today' to 3 'Still some research necessary'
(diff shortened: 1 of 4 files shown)


fschreyer:
Thanks!

I would replace ``for dac: $(2015)/tC(removed)`` by ``for dac: $(2015)/tC(captured)``. Removed could be interpreted as "stored".

chroetz:
> I would replace for dac: $(2015)/tC(removed) by for dac: $(2015)/tC(captured). Removed could be interpreted as "stored".

True! I changed it

chroetz:
I noticed that sometimes `yr` is used for the unit *year* and sometimes `a`...

core/input/generisdata_tech.prn
```diff
@@ -182,7 +184,7 @@ inco0               0.01       21600
 mix0                   0           0
 eta                 1.00        1.00
 omf                 0.84       0.025
-omv                    0           0
+omv                    0
```
LaviniaBaumstark, line 187: are you sure that here are all row correct? in .prn files we need to be careful with tabs

chroetz: hm, pretty sure. `Rscript start.R -q` works. No value is treated as 0. We decided to remove the 0 as it is not used for dac.

LaviniaBaumstark (approved):

PR #949, opened 2022-08-24
report run statistics also for failed runs

giannou:
# Purpose of this PR

If the GAMS part of a REMIND run fails, run statistics (like e.g. model runtime) were not reported. This PR fixes that.

## Type of change

- [x] Bug fix 

## Checklist:

### DIFF
--- a/scripts/start/prepare_and_run.R
+++ b/scripts/start/prepare_and_run.R
@@ -1062,11 +1062,6 @@ run <- function(start_subsequent_runs = TRUE) {
   }
   message("")
 
-  if (stoprun) {
-    stop("GAMS did not complete its run, so stopping here:\n       No output is generated, no subsequent runs are started.\n",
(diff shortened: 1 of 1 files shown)


0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q (approved):

orichters:
@giannou: If you tested it and it works, that is fine! Before I changed it a while ago, the runtime reporting always [tried to read from fulldata.gdx](https://github.com/remindmodel/remind/commit/d596fef72f1878d790c720a3f4c57a96e28c0530#diff-c328a7ad81b0c7c5467604c39f07ba8f2cf20e165d7a72df04abf175ddd58e8aL1053) and this always failed if this file did not exists. As a result, runstatistics were not reported, and the user was left with a a cryptic error message. After this PR, if no gdx file exists, `numeric(0)` seems to be passed as modelstat, so if `runstatistics` can live with that, then everything is fine and thanks for changing it.

PR #954, opened 2022-08-26
move non_optimal.gdx and abort.gdx for calib and restarts, cleanup

orichters:
## Purpose of this PR

- if a run writes a non_optimal.gdx or abort.gdx, this file remains there also for the next calibration iteration or if the run is restarted, confusing not only a variety of scripts from rs2 to the model summary in log.txt, but also potentially the user. With this PR, the files are moved once the model is restarted to `non_optimal_beforeRestart.gdx` etc., and once a calibration iteration (say: 5) is over, to `non_optimal_05.gdx` etc.. This way, they are clearly marked as outdated.
- <s>I don't particularly like the `system(mv…)` call, but there seems to be no native `file.move` in R. Corrections welcome.</s> `file.rename` is it…
- I haven't tested the part in the calibration iteration in a full run as I don't do calibrations, but I guess it should be safe.
- fix minor comment confusion issue created by this bugfix, namely now having `****` which looks like an error in `full.lst`: https://github.com/pik-piam/magclass/pull/125

## Type of change

- [x] Bug fix 
- [x] Minor change

## Checklist:

### DIFF
--- a/scripts/start/prepare_and_run.R
+++ b/scripts/start/prepare_and_run.R
@@ -959,6 +959,9 @@ run <- function(start_subsequent_runs = TRUE) {
         file.copy("fulldata.gdx", paste0(cfg$gms$cm_CES_configuration,".gdx"), overwrite = TRUE)
         file.copy("fulldata.gdx", sprintf("input_%02i.gdx", cal_itr),
                   overwrite = TRUE)
+        if (file.exists("non_optimal.gdx")) {
+          file.rename("non_optimal.gdx", sprintf("non_optimal_%02i.gdx", cal_itr))
+        }
 
         # Update file modification time
         fulldata_m_time <- file.info("fulldata.gdx")$mtime
--- a/scripts/start/prepare_and_run.R
+++ b/scripts/start/prepare_and_run.R
@@ -953,19 +953,20 @@ run <- function(start_subsequent_runs = TRUE) {
         getLoadFile()
 
         # Store all the interesting output
-        file.copy("full.lst", sprintf("full_%02i.lst", cal_itr), overwrite = TRUE)
-        file.copy("full.log", sprintf("full_%02i.log", cal_itr), overwrite = TRUE)
+        interestingOutput <- c("full.lst", "full.log", "fulldata.gdx", "non_optimal.gdx", "abort.gdx")
+        file.copy(from = interestingOutput,
+                  to = sub("^(.*)(\\.[^\\.]+)$", sprintf("\\1_%02i\\2", cal_itr), interestingOutput), overwrite = TRUE)
         file.copy("fulldata.gdx", "input.gdx", overwrite = TRUE)
-        file.copy("fulldata.gdx", paste0(cfg$gms$cm_CES_configuration,".gdx"), overwrite = TRUE)
-        file.copy("fulldata.gdx", sprintf("input_%02i.gdx", cal_itr),
-                  overwrite = TRUE)
-        if (file.exists("non_optimal.gdx")) {
-          file.rename("non_optimal.gdx", sprintf("non_optimal_%02i.gdx", cal_itr))
+        if (cal_itr < cfg$gms$c_CES_calibration_iterations) {
+          unlink(c("abort.gdx", "non_optimal.gdx"))
+        } else { # calibration was successful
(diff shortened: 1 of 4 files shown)


scripts/start/prepare_and_run.R
```diff
@@ -959,6 +959,9 @@ run <- function(start_subsequent_runs = TRUE) {
         file.copy("fulldata.gdx", paste0(cfg$gms$cm_CES_configuration,".gdx"), overwrite = TRUE)
         file.copy("fulldata.gdx", sprintf("input_%02i.gdx", cal_itr),
                   overwrite = TRUE)
+        if (file.exists("non_optimal.gdx")) {
+          file.rename("non_optimal.gdx", sprintf("non_optimal_%02i.gdx", cal_itr))
+        }
```
0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q, line 964: How about consolidating this a bit?
```
interesting_output <- c('full.lst', 
                        'full.log', 
                        'fulldata.gdx', 
                        'non_optimal.gdx', 
                        'abort.gdx')

file.copy(from = interesting_output,
          to = sub('^(.*)(\\.[^\\.]+)$', 
                   sprintf('\\1_%02i\\2', cal_itr), 
                   interesting_output),
          overwrite = TRUE)
```
`file.copy()` doesn't care about missing files, just returns `FALSE` for those.

The 
```
file.copy("fulldata.gdx", paste0(cfg$gms$cm_CES_configuration,".gdx"), overwrite = TRUE)
```
part I would prefer to have somewhere outside the loop.  We should only provide this file if the calibration was successful (i.e. did all the calibration iterations set).  See below.

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q: Hmm, can't comment on lines you didn't touch.
So after the closing brace in line 972, before
```
  } else {
    stop("unknown realisation of 29_CES_parameters")
  }
```
add
```
# if the last CES calibration iteration was successful, write out calibration
# results
if (file.info("fulldata.gdx")$mtime > fulldata_m_time) {
    file.copy("fulldata.gdx", paste0(cfg$gms$cm_CES_configuration, '.gdx'))
    
    file.copy(from = paste0(cfg$gms$cm_CES_configuration, '_ITERATION_',
                            cfg$gms$c_CES_calibration_iterations, '.inc'),
              to = paste0(cfg$gms$cm_CES_configuration, '.inc'))
}
```

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q (approved):
Two suggestions for consolidating the code a bit and making it easier to pick up calibration results via script (something I plan on implementing).  Not really necessary for this merge request, so go ahead without it if you don't feel like it.

scripts/start/prepare_and_run.R
LaviniaBaumstark: > The
> 
> ```
> file.copy("fulldata.gdx", paste0(cfg$gms$cm_CES_configuration,".gdx"), overwrite = TRUE)
> ```
> 
> part I would prefer to have somewhere outside the loop. We should only provide this file if the calibration was successful (i.e. did all the calibration iterations set)

We already store a gdx named with the calibration-scenario which is later used for loading the right gdx-file, if the calibration was successful. Why do we need another gdx file?

LaviniaBaumstark (approved):

scripts/start/prepare_and_run.R
0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q: I suggested to create this file only when the calibration finishes successfully, not after each calibration iteration that had one or more feasible Nash iterations.  We don't want that file available for use when the calibration crashed half way.

orichters: Your suggestion with the list to copy is not what I intend. I think `abort.gdx` and `non_optimal.gdx` should be renamed, not copied. Concerning the rest, I'm not really into calibration stuff and so it would be great if you could implement and test these changes yourself. So I would prefer to merge my stuff and you can improve it. But: should `non_optimal.gdx` and `abort.gdx` be renamed also for the last calibration run, or should they stay (or be copied) in this case such that the scripts correctly identify that there were problems?

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q (approved):

LaviniaBaumstark (approved):

PR #961, opened 2022-08-31
add DESCRIPTION, make renv install modelstats

pfuehrlich-pik:
# Purpose of this PR
Add a DESCRIPTION file to document dependencies and other metadata in a standard way. Renv would not realize `modelstats` was a dependency (for rs2), so it is now listed explicitly as a dep in DESCRIPTION.

## Type of change

- [x] Bug fix 

## Checklist:

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,4 +1,3 @@
-Package: remindmodel
 Title: REMIND - REgional Model of INvestments and Development
 Version: 0.1
 Date: 2022-08-31
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,0 +1,75 @@
+Package: remindmodel
+Title: REMIND - REgional Model of INvestments and Development
+Version: 0.1
+Date: 2022-08-31
+Description: This is not a package yet, the purpose of this file is to
+    document metadata, notably dependencies.
+License: file LICENSE
+URL: https://github.com/remindmodel/remind,
+    https://doi.org/10.5281/zenodo.6794920
+BugReports: https://github.com/remindmodel/remind/issues
+Depends:
+    R (>= 4.0.0)
+Imports:
+    callr,
+    colorspace,
+    crayon,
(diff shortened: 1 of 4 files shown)


DESCRIPTION
```diff
@@ -0,0 +1,75 @@
+Package: remindmodel
```
0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q, line 1: REMIND is not a package.

pfuehrlich-pik: That's true. I got the idea to create a DESCRIPTION anyways from [Hadley Wickham](https://r-pkgs.org/Metadata.html), all the way at the bottom he describes this off-label use of a DESCRIPTION. Tools like renv and devtools can work with this DESCRIPTION even if the whole thing is not a package. Apparently the way you set a project's name in a DESCRIPTION is via the `Package` field, in the `Description` field it is clearly state that this is not a package though.

giannou: I'm fine with adding the DESCRIPTION file. Will it then complement the automatic installation of needed packages and dependencies by `renv`?

pfuehrlich-pik: Yes, `renv::dependencies` will check the DESCRIPTION in addition to scanning all R and Rmd files to determine dependencies, and this is what's installed when running R with an empty renv.

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q: > I got the idea to create a DESCRIPTION anyways from [Hadley Wickham](https://r-pkgs.org/Metadata.html)

… from Hadley Wickham's book on writing R packages, which REMIND isn't.

`renc::dependencies()` also parses DESCRIPTION files without a `Package` field.

pfuehrlich-pik: I removed the `Package: remindmodel` line

giannou (approved):

PR #965, opened 2022-09-07
update renv/activate.R to 0.15.5

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:

### DIFF
--- a/renv/activate.R
+++ b/renv/activate.R
@@ -2,7 +2,7 @@
 local({
 
   # the requested version of renv
-  version <- "0.15.4"
+  version <- "0.15.5"
 
   # the project directory
   project <- getwd()
@@ -54,7 +54,7 @@ local({
   # mask 'utils' packages, will come first on the search path
   library(utils, lib.loc = .Library)
 
-  # unload renv if it's already been laoded
+  # unload renv if it's already been loaded
   if ("renv" %in% loadedNamespaces())
     unloadNamespace("renv")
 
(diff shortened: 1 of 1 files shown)


pfuehrlich-pik:
When checking out this PR I get
```
renv 0.15.4 was loaded from project library, but this project is configured to use renv 0.15.5.
Use `renv::record("renv@0.15.4")` to record renv 0.15.4 in the lockfile.
Use `renv::restore(packages = "renv")` to install renv 0.15.5 into the project library.
```
This could irritate some people and leave them unsure what to do, so I'd update renv only when necessary. Is there an important feature/bugfix in renv 0.15.5?

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
> outdated (defined as "an update is available") packages

[[↑]](https://github.com/remindmodel/remind/pull/808#issuecomment-1128530734)

> Is there an important feature/bugfix in renv 0.15.5?

[Lots](https://github.com/remindmodel/remind/pull/808#issuecomment-1128530734), none of which I know to be relevant for REMIND.

The irritating message will show up when library and lockfile version diverge, whenever that may be.  Users also get irritating messages when they update their package library more regularly than the cluster library is updated.

pfuehrlich-pik:
The updateRenv script, which is recommended for updates in the tutorial, will create a snapshot right after updating, so there won't be such a message. The message is also different when updating renv vs. any other package, although renv quite nicely tells you what to do, so it's probably not a big deal.
That being said, I'd be the last one to oppose frequent updates. It's just not what some (most?) REMIND users want usually. What's your take on this @LaviniaBaumstark ?

LaviniaBaumstark:
For now I would not update. Would be different if we would know of a fix which is necessary for us. Let's try to not confuse people starting to work with renv

PR #966, opened 2022-09-07
use custom renv snapshot filter to include all of, but only those packages required by REMIND in the lock file

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
otherwise, REMIND breaks on renv being unable to install packages for which it can't determine upstream package sources (development packages, packages on repositories requiring access tokens)

- [x] Bug fix 

### DIFF
--- a/renv/settings.dcf
+++ b/renv/settings.dcf
@@ -3,7 +3,7 @@ external.libraries:
 ignored.packages:
 package.dependency.fields: Imports, Depends, LinkingTo
 r.version:
-snapshot.type: all
+snapshot.type: implicit
 use.cache: TRUE
 vcs.ignore.cellar: TRUE
 vcs.ignore.library: TRUE


0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
[renv_fail_log.txt](https://github.com/remindmodel/remind/files/9524904/renv_fail_log.txt)

pfuehrlich-pik:
I was hoping we could solve this problem with `ar5data` and `sr15data` by not using these two packages anymore. `ar5data` is never declared as a dependency and does not seem to be used anywhere (it does not show up when running `reposearch ar5data` on cluster), can you outline in which cases it is used?
For sr15data: it is an optional dependency of remind2 (but never used there) and remind (the deprecated R package). My plan was to remove it from remind2, and ignore the remind R package as it is deprecated. Do you see a problem with that? If yes, when is sr15data needed?

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
The point is that `renv` as currently configured will try to install [all packages in the library](https://rstudio.github.io/renv/reference/snapshot.html#snapshot-type) regardless of whether they are required or not, and fail if it can't figure out a valid upstream source.  This is the case for couple of packages that basically make data accessible, but which we can't publish due to license restrictions (`ar5data`, `ar6data`, `ieaeb2014`, `ieaeb2022`) and also all packages people develop locally.
This makes REMIND unusable for development and testing on local machines.

pfuehrlich-pik:
Are you sure about that? As I understand renv analyzes *.R, *.Rmd, and the DESCRIPTION for dependencies and installs these into the renv library [here](https://github.com/remindmodel/remind/blob/3b9f369c4b5cd32d01e9c3e541b5dc4f2c9d0e4f/.Rprofile#L11). The renv snapshot type only controls which packages that are present in the renv library are written into renv.lock, but it does not control which packages are installed into the renv library.
Ah but I think I see the problem now, you manually installed e.g. ar5data into the renv library, and renv then fails to snapshot that, right? I'd say it's a feature not a bug: when renv cannot restore the package environment there should be an error. For these cases renv provides a [cellar](https://rstudio.github.io/renv/articles/cellar.html), but I've not used that before.
Also, it seems like your local remind has a dependency on one of the problematic packages, otherwise you'd not need the package in your remind renv. The online repo/develop branch of remind does not have that dependency though, so I'd say that's not a thing that should be fixed here/for everyone. Instead I'd suggest you look into the cellar feature and/or put the code from this PR into you `~/.Rprofile`:
```
# set custom renv snapshot filter to include all of, but only those packages
# required by REMIND in the lock file.
options(renv.snapshot.filter = function(project) {
    unique(unlist(tools::package_dependencies(
        renv::dependencies(path = project, progress = FALSE)$Package,
        recursive = TRUE), use.names = FALSE))
})
```
You can set the snapshot type to custom [here](https://github.com/remindmodel/remind/blob/3b9f369c4b5cd32d01e9c3e541b5dc4f2c9d0e4f/renv/settings.dcf#L6)

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
Skipping over the parts indicating that you didn't read the problem description …

> You can set the snapshot type to custom [here](https://github.com/remindmodel/remind/blob/3b9f369c4b5cd32d01e9c3e541b5dc4f2c9d0e4f/renv/settings.dcf#L6)

made me test that.  From the `renv` reference
> **"implicit"**
> Only capture packages which appear to be used in your project in the lockfile. The intersection of packages installed in your R libraries, alongside those used in your R code as inferred by [renv::dependencies()](https://rstudio.github.io/renv/reference/dependencies.html), will enter the lockfile.

It wasn't clear (to me) that this includes recursive dependencies, which was the whole point of the custom filter, as `renv::dependencies()` does not return those.

So: 
https://github.com/remindmodel/remind/blob/3b9f369c4b5cd32d01e9c3e541b5dc4f2c9d0e4f/renv/settings.dcf#L6
is a bug, because it will cause all packages installed in the user library, whether REMIND depends on them or not, to be included in the lock file, which fill fail and break REMIND if any random package has no source available.
```
snapshot.type: implicit
```
[is correct](https://github.com/0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q/remind/blob/39d0e0c479b70b2537dd6f6be0ce0f98d4efb138/renv/settings.dcf#L6), as it only includes actually required packages in the lock file.

pfuehrlich-pik:
Good point, with the DESCRIPTION we can explicitly list dependencies that renv would otherwise not detect (like `kableExtra`) so now implicit snapshots are probably the best fit. One small downside is that in a run folder `renv::status()` now reports a bunch of unused packages (e.g. `kableExtra`). Might be irritating, but does not really change anything, so I think we can leave it. The alternative would be to copy the DESCRIPTION to every run folder.

pfuehrlich-pik (approved):

PR #970, opened 2022-09-09
fix various calibration issues

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
* use exp(log(a) * b) = a ** b for the production function and calibration, because the latter formulation is not accurate in GAMS for small values
* scale `en`, `kap`, or both prices should the value of `en`, `kap`, or the sum of them exceed `inco`, to prevent &xi; < 0 errors
* prevent `p80_repyLastOptim` from being evaluated in the first iteration, as it would only hold non-sensical data, or no data at all if based on a `testOneRegi` run
* exclude `p37_energy_limit_slope` calculation from calibration runs to enable use of `testOneRegi` gdxes

- [x] Bug fix 
- [x] New feature 

## Checklist:


modules/29_CES_parameters/calibrate/preloop.gms
```diff
@@ -716,6 +716,103 @@ pm_cesdata(t,regi_dyn29,"kap","eff")
 
 display "after change cap eff consistency", pm_cesdata, pm_cesdata_putty;
 
+*** If the value (quantity x price) of either en or kap, or the sum of both,
+*** exceed the quantity of inco (all of which would result in negative labour
+*** prices), scale these prices down accordingly, and warn this is happening.
```
LaviniaBaumstark, line 721: I assume, you tested it. I am always a bit reluctant scaling manually some values. I am afraid that we will have to adjust it too often. Does the auto-scaling not do its job?

LaviniaBaumstark (commented):
thanks, nice

modules/29_CES_parameters/calibrate/preloop.gms
0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q: Yes, I tested it.  This is no manual scaling, the scaling factors are calculated endogenously, there are no values to adjust.  There was no automatic scaling prior to this patch.

LaviniaBaumstark (approved):

PR #974, opened 2022-09-16
 make parallel mode the default for REMIND-MAgPIE coupling, use lucode::readArgs for flags

orichters:
## Purpose of this PR
- `main.gms`: adapt default settings to default.cfg (has been done automatically by the start scripts)
- `output.R`: add `--test` mode where nothing is executing for better debugging. Maybe @pfuehrlich-pik can have a look at that.
- `start.R`: use the [`lucode2::readArgs`](https://github.com/pik-piam/lucode2/blob/master/R/readArgs.R) function that was enhanced in [lucode PR 113](https://github.com/pik-piam/lucode2/pull/113) to treat flags such as `--test` instead of defining this functionality multiple times. Use `flags` instead of `argv` – the latter also contains the csv file and further command line arguments. Maybe @pfuehrlich-pik can have a look at that.
- `start_bundle_coupled.R`: Also use `flags` and `readArgs`. Maybe @pfuehrlich-pik can have a look at that.
- `start_bundle_coupled.R` and `start_coupled.R`: remove the old mode for REMIND-MAgPIE coupling. The parallel mode (that was introduced in https://github.com/remindmodel/remind/pull/794) is now the only mode. This reduces confusion, streamlines the tutorials, reduces total runtime, generates consistent sets of -rem-x scenarios for all x (not only for x = max_iterations), and allows using priority slots for coupled runs. I see no downsides. I essentially went through the code and looked for all instances of `parallel` and then assumed this is `TRUE`. Delete the warning of not using `C_` for coupled runs. I replaced all hardcoded instances of `C_` already a while ago. See successful test runs in `tmp/oliverr/remind-smallfix/output/test_*` using `prefix_runname <- "test"`. This is something for @dklein-pik.
- delete the Sleep(10) command that was introduced before @mikapfl fixed the locking
- @LaviniaBaumstark: FYI.

## Type of change

- [x] Refactoring: coupled start scripts
- [x] New feature: --test mode for `output.R`

## Checklist:

- [x] The model compiles and runs successfully (`Rscript start.R -q`): `tmp/oliverr/remind-smallfix/output/testOneRegi`


start_bundle_coupled.R
```diff
@@ -5,8 +5,30 @@
...
+require(magclass)
+require(gms)
+require(remind2)
+require(gtools) # required for mixedsort()
+require(dplyr) # for filter, secelt, %>%
+require(stringr) # for str_sub
+
+helpText <- "
+#' Usage:
+#' Before running this script, edit this file and provide user settings
```
pfuehrlich-pik, line 18: What is "this file" here?

pfuehrlich-pik (commented):
Looks good, thanks for cleaning up here!
What is actually tested with "--test"? I have the feeling "--dry-run" or similar could be a better name for what the flag is doing?

start_bundle_coupled.R
orichters: well, `start_bundle_coupled.R` itself

orichters: I agree `--test` is not entirely appropriate, but: We had a `test` mode in `start_bundle_coupled.R` since Lavinia [moved the code to github](https://github.com/remindmodel/remind/blob/6dc75ecdfc61da2e4c439fc2f3e8ce0c6d3d81f4/start_bundle_coupled.R#L63), and it simply [did not submit](https://github.com/remindmodel/remind/blob/6dc75ecdfc61da2e4c439fc2f3e8ce0c6d3d81f4/start_bundle_coupled.R#L321-L322) the runs to the cluster.

I liked that and [added such a mode to start.R in March](https://github.com/remindmodel/remind/pull/701). So I guess this naming is already well established so I would rather not dare to change it.

dklein-pik (approved):
Thank you! Very useful!

start_bundle_coupled.R
pfuehrlich-pik: Alright, then --test is the better choice, I agree. I was thinking if another script for some reason runs `source("start_bundle_coupled.R")` it might be confusing to have "this file". You could consider just writing out "start_bundle_coupled.R", but I approve either way :slightly_smiling_face:

orichters: I fixed the `--help` text. For `output.R`, one may think in the future to pass the `--test` option to the individual output scripts (in `direct` mode) and implement it such that they don't write stuff, but rather test whether everything is fine. But this would be quite some work, I guess.

PR #977, opened 2022-09-21
add CES calibration tutorial

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
- [x] New feature 

@gabriel-abrahao Here's the tutorial, give it a read-through once you have time.  My questions:
- Does it make sense to you, do you get an idea how the calibration works?
- Can you set up a calibration run?  (Kind of trivial I think.)
- I can give you a finished calibration run in case you want to check through the output part.

@LaviniaBaumstark, you're here to sign off on my OCD renaming the tutorials such that they show up in order in the listing.


tutorials/12_CES_Calibration.md
```diff
@@ -0,0 +1,225 @@
...
+   energy service demands, labour, capital) and the output (`inco`, GDP), and
+2. the previous iterations' `ppf` prices.
+
+Trajectories under (1) come from the input files `./core/input/f_gdp.cs3r`,
+`./core/input/f_pop.cs3r`,
+`./modules/29_CES_parameters/calibrate/input/f29_capitalQuantity.cs4r`, and
+`./core/input/f_fedemand.cs4r` which are generated automatically by `mrremind`
+and always present.
+
+Prices under (2) are calculated using the `input.gdx` provided to the
```
LaviniaBaumstark, line 61: should we mention here that usually the 'input.gdx' file is a copy of a gdx file from former runs which comes with the CES-parameters and has the name "cm_CES_configuration"

tutorials/12_CES_Calibration.md
```diff
@@ -0,0 +1,225 @@
...
+`./modules/29_CES_parameters/load/input/` and the .gdx file to the directory
+`./config/gdx-files/`.  At PIK, this is done automatically using the RSE support
+scripts.  See [this wiki
+page](https://redmine.pik-potsdam.de/projects/remind-r/wiki/GDX_and_CES_parameter_Handling)
+for details.
+
+As for diagnostic output, there are the `full.log` and `full.lst` files for each
+calibration iteration (`full_01.log` …), the file `CES_calibration.csv`
+containing all the relevant calibration parameters (inputs and outputs) for all
+iterations for automated analysis and a `CES_calibration_report` .pdf file with
```
LaviniaBaumstark, line 122: add info that the `CES_calibration_report` also can be generates by "Rscript output.R" up to any existing CES-iteration afterwards

LaviniaBaumstark (approved):
thanks, nice, I like the renamed tutorials to have them in the correct order!
Maybe a better name for the new tutorial would be "12_Running_CES_Calibration.md" or "12_Calibrating_CES_Parameters.md" in line with the other tutorial names

tutorials/12_CES_Calibration.md
0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q: People do know that the `input.gdx` comes from a previous REMIND run when they made it this far in the tutorials, I would argue.  I added a note on `cm_CES_configuration` below in the Results section.

LaviniaBaumstark (approved):

gabriel-abrahao (approved):
Thanks for doing this, it will be very helpful! I went through it and all steps seem clear. I managed to start a calibration and all the file names seem to check out. 

I also added a short introduction saying what the calibration is for and when you might need one. I felt like this was needed since I took a long time to understand this myself, so explaining it before going into the details might be good for a newcomer. Please take a quick look at it, and if I said something wrong there, that's just a sign of how badly we needed this tutorial!

Also added some pointers on how to actually start a run and so on, which I feel might also be useful for newcomers to follow.

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
Thank @gabriel-abrahao!  I combined your introduction with mine, as they shared a lot of material and read quite repetitive.  I also took the liberty to adjust your other comments a little.  Newcomers shouldn't calibrate REMIND, and everybody reading this tutorial should have read those before and be able to start runs, so there is no need to rehash that here.  (Also, the `default.cfg` part is going to change in the foreseeable future (#986), this way we don't have to adjust that here.

PR #979, opened 2022-09-26
clean up and improve documentation of switches I introduced or dealt with

fschreyer:
# Purpose of this PR

This cleans up a bunch of switches in the default.cfg. Deprecated switches were removed, some switches were extended/relabeled and numerous switches were described in more detail. 

## Type of change

- [x] Refactoring

## Checklist:

- [x] check whether switches work after reformulation

### DIFF
--- a/modules/47_regipol/regiCarbonPrice/datainput.gms
+++ b/modules/47_regipol/regiCarbonPrice/datainput.gms
@@ -109,9 +109,8 @@ loop(ext_regi$altFeEmiFac_regi(ext_regi),
 $endif.altFeEmiFac
 
 
-*** VRE capacity factor adjustments for Germany in line with ARIADNE assumptions
-$ifthen.GerVRECapFac not "%cm_ariadne_VRECapFac_adj%" == "off" 
-loop(te$sameas(te,"wind"),
+*** VRE capacity factor adjustments for Germany in line with results from detailed models in ARIADNE project
+ loop(te$sameas(te,"wind"),
   loop(regi$sameas(regi,"DEU"),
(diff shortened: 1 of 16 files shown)


modules/47_regipol/regiCarbonPrice/datainput.gms
```diff
@@ -109,9 +109,8 @@ loop(ext_regi$altFeEmiFac_regi(ext_regi),
 $endif.altFeEmiFac
 
 
-*** VRE capacity factor adjustments for Germany in line with ARIADNE assumptions
-$ifthen.GerVRECapFac not "%cm_ariadne_VRECapFac_adj%" == "off" 
-loop(te$sameas(te,"wind"),
+*** VRE capacity factor adjustments for Germany in line with results from detailed models in ARIADNE project
```
Renato-Rodrigues, line 112: I would suggest to remove this code from remind and instead add this change directly to the "calcCapacityFactor" function in mrremind, as the "f_cf.cs3r" input file exists to provide exactly that input to remind. 

If you open the function you will see that this is already done there for Germany coal capacity factors for example.

That way you also avoid the need to have this code repeated in the none realization, and its results will be also considered in a H12 run.

Renato-Rodrigues (changes requested):

modules/47_regipol/regiCarbonPrice/datainput.gms
fschreyer: I agree. Let's include that in some future clean-up when we scan the datainput files for stuff that can be moved to mrremind.

LaviniaBaumstark (approved):

PR #986, opened 2022-09-29
Start scripts

LaviniaBaumstark:
# Purpose of this PR
This PR is the first step toward the new structure of our start scripts. We will get rid of duplications in main.gms and default.cfg. With this PR:
* all documentation is moved from default.cfg to main.gms
* all GAMS-related switches are moved from default.cfg to main.gms
* the new function readSettings() is used to read in settings from main.gms
* default.cfg now only contains settings which are used during preparation phase

ATTENTION: Currently, main.gms contains the default settings of REMIND. Please be careful when changing this file. We are working on a solution which will not adjust any file when setting up a REMIND run. But for now, please make sure that all changes to main.gms are intended! Usage of scenario_config_* - files is unchanged.  

## Type of change

- [x] Refactoring

## Checklist:

* Test runs are here: /p/tmp/lavinia/MO/start_scripts/fork_Lav_start_scripts/remind and /p/tmp/lavinia/MO/start_scripts/nachMerge_2022_09_29/remind


scripts/start/prepare_and_run.R
```diff
@@ -245,11 +245,11 @@ prepare <- function() {
   setwd(cfg$remind_folder)
 
   # Check configuration for consistency
-  cfg <- check_config(cfg, reference_file="config/default.cfg",
-                      settings_config = "config/settings_config.csv",
-                      extras = c("backup", "remind_folder", "pathToMagpieReport", "cm_nash_autoconverge_lastrun",
-                                 "gms$c_expname", "restart_subsequent_runs", "gms$c_GDPpcScen",
-                                 "gms$cm_CES_configuration", "gms$c_description"))
+#  cfg <- check_config(cfg, reference_file="config/default.cfg",
```
dklein-pik, line 248: This is outcommented because there is no default.cfg anymore against which the cfg object could be compared, right? How do we deal with this in future? What is the purpose of the consistency check? Do we need a replacement for this in future?

dklein-pik (approved):

scripts/start/prepare_and_run.R
LaviniaBaumstark: right, the new default.cfg does not include all settings and therefore the check would not work. We will have to introduce a similar check for our new starting script structure in the end

main.gms
```diff
@@ -230,548 +217,1220 @@ $setGlobal c_description  REMIND run with default settings
...
 ***-----------------------------------------------------------------------------
-***                     SWITCHES and FLAGS
+***                     SWITCHES 
 ***-----------------------------------------------------------------------------
 ***--------------- declaration of parameters for switches ----------------------
-parameters
+parameter
   cm_iteration_max          "number of iterations, if optimization is set to negishi or testOneRegi; used in nash mode only with cm_nash_autoconvergence = 0"
+;
+  cm_iteration_max       = 1;     !! def = 1
```
mikapfl, line 410:
```suggestion
parameter
  cm_iteration_max          "number of iterations, if optimization is set to negishi or testOneRegi; used in nash mode only with cm_nash_autoconvergence = 0"
         / 1 /    !! def = 1
;
```
Remind me again what the reason was to do the definition and the starting value in two different statements. Is ist because `gms` and/or `lucode2` can't deal with the combined definition?

main.gms
```diff
@@ -230,548 +217,1220 @@ $setGlobal c_description  REMIND run with default settings
...
+$setGlobal c_scaleEmiHistorical  on  !! def = on
+$setGlobal cm_nash_mode  parallel      !! def = parallel
+$SetGlobal cm_quick_mode  off          !! def = off
+$setGLobal cm_debug_preloop  off !! def = off
+$setGlobal c_EARLYRETIRE  on         !! def = on
+*** set conopt version. Warning: conopt4 is in beta
+$setGlobal cm_conoptv  conopt3    !! def = conopt3
+$setGlobal cm_ccsfosall  off        !! def = off
+$setGlobal cm_APscen  SSP2          !! def = SSP2
+$setglobal cm_CES_configuration  indu_subsectors-buil_simple-tran_edge_esm-POP_pop_SSP2EU-GDP_gdp_SSP2EU-En_gdp_SSP2EU-Kap_debt_limit-Reg_62eff8f7   !! this will be changed by start_run()
```
mikapfl, line 1421: Just for aesthetics: Could we write `$setGlobal` everywhere, with consistent capitalization? I know that GAMS allows whatever, but if we use the same capitalization here that we also would use in R, it is one thing less to think about. If wanted, I can write a `sed` command to standardize all `$setGlobal` statements.

start.R
```diff
@@ -331,6 +331,8 @@ if (any(c("--reprepare", "--restart") %in% argv)) {
     
     # state if columns are unknown and probably will be ignored, and stop for some outdated parameters.
     source("config/default.cfg")
+    # read in GAMS related switches from main.gms
+    cfg$gms <- as.list(readSettings("main.gms"))
```
mikapfl, line 335: Do we have to note somewhere that `gms` in Version `0.18.0` or higher is now needed to start REMIND?

mikapfl (approved):

main.gms
```diff
@@ -230,548 +217,1220 @@ $setGlobal c_description  REMIND run with default settings
 ***------------------------------------------------------------------------------
 
 ***---------------------    01_macro    -----------------------------------------
+*** (singleSectorGr):
```
giannou, line 220: Here I would either remove the `:` or write a description

main.gms
```diff
@@ -230,548 +217,1220 @@ $setGlobal c_description  REMIND run with default settings
 ***------------------------------------------------------------------------------
 
 ***---------------------    01_macro    -----------------------------------------
+*** (singleSectorGr): 
 $setGlobal macro  singleSectorGr  !! def = singleSectorGr
 ***---------------------    02_welfare    ---------------------------------------
+*** (utilitarian):
```
giannou, line 223: Here I would either remove the `:` or write a description

main.gms
```diff
@@ -230,548 +217,1220 @@ $setGlobal c_description  REMIND run with default settings
...
+*** (utilitarian): 
 $setGlobal welfare  utilitarian  !! def = utilitarian
 ***---------------------    04_PE_FE_parameters    ------------------------------
+*** (iea2014): new PE-FE parameters based on IEA 2014
 $setGlobal PE_FE_parameters  iea2014  !! def = iea2014
 ***---------------------    05_initialCap    ------------------------------------
+*** (on):      load existing CES parameters matching model configuration
 $setGlobal initialCap  on             !! def = on
 ***---------------------    11_aerosols    --------------------------------------
+*** (exoGAINS)   :
```
giannou, line 232: formatting: remove unneeded spaces and treat the `:` as I mentioned above

main.gms
```diff
@@ -230,548 +217,1220 @@ $setGlobal c_description  REMIND run with default settings
...
+parameter
   cm_deuCDRmax                 "switch to limit maximum annual CDR amount in Germany in MtCO2 per y"
-  cm_EnSecScen             "switch for running an ARIADNE energy security scenario, introducing a tax on PE fossil energy in Germany"
-  cm_Ger_Pol               "switch for selecting different policies for Germany used in the ARIADNE scenarios"
-  cm_process_based_steel      "switch to turn on process-based steel implementation"
+*** switch to cap annual DEU CDR amount by value assigned to switch, or no cap if -1, in MtCO2
 ;
-
-*** --------------------------------------------------------------------------------------------------------------------------------------------------------------------
-***                           YOU ARE IN THE WARNING ZONE (DON'T DO CHANGES HERE)
```
giannou, line 440: This has to be removed, right?

main.gms
```diff
@@ -230,548 +217,1220 @@ $setGlobal c_description  REMIND run with default settings
...
 $setGlobal optimization  nash         !! def = nash
 ***---------------------    81_codePerformance    -------------------------------
+*** (off): nothing happens
+*** (on):  test code performance: noumerous (30) succesive runs performed in a triangle, tax0, tax30, tax150, all growing exponentially,
+***                    therefore use carbonprice|exponential, c_emiscen|9, and cm_co2_tax_2020|0.
 $setGlobal codePerformance  off       !! def = off
 
 ***-----------------------------------------------------------------------------
-***                     SWITCHES and FLAGS
+***                     SWITCHES
```
giannou, line 404: The following still contains flags, so why this change? Also, adding an explanation of what a switch does and how it is different than what a flag does, might be useful

giannou (approved):

giannou:
An update of the tutorials is also needed, right? Also of this file: https://github.com/LaviniaBaumstark/remind/blob/start_scripts/scripts/start/readme.txt

LaviniaBaumstark:
thanks for looking into codeCheck. For now I would delete setGlogals from not_used.txt files. That's how it is handled for all other globals. Probably, I some of the entries in not_used are only because the globals were also declared as parameters which I cleaned with this PR. In the long-term we anyway need to find a way to document ALL settings in the documentation.

main.gms
LaviniaBaumstark: tradition? I also wanted to change to your suggestion but wanted to keep the current version for now. We anyway need to come up with a template how to organize parameter description/documentation in main.gms.

main.gms
LaviniaBaumstark: I agree, but let's open a separate PR after we have merged this one for that

main.gms
LaviniaBaumstark: I would keep the `:` to remind REMIND developers to add a description :-)

main.gms
LaviniaBaumstark: it is isn't it?

main.gms
LaviniaBaumstark: flags are now starting with:
***-----------------------------------------------------------------------------
***                     FLAGS 
***-----------------------------------------------------------------------------

main.gms
giannou: sorry, I meant it has to be removed also from other places where it is still there

main.gms
mikapfl: Note that a PR which does en-masse transformation will always have the potential for lots of merge conflicts. That's why it would be easier to do this now, when merge conflicts have to be dealt with anyway and no additional merge conflicts will be generated by changing the format en-masse.

main.gms
LaviniaBaumstark: I cannot find it anywhere in main.gms

giannou:
The help text at the top of `start.R` needs an update, it currently states ` Without additional arguments this starts a single REMIND run using the settings from config/default.cfg.`

Renato-Rodrigues (approved):

Issue #988, opened 2022-10-04
after rerunning REMIND reporting, MAgPIE results are missing


orichters:
After each REMIND iteration, `start_coupled.R` [writes the MAgPIE data into the REMIND mif file](https://github.com/remindmodel/remind/blob/727dd53fef47ca4b98f09d4ce5f342ddb1393f13/start_coupled.R#L200-L224). But if you rerun the REMIND reporting, you lose this information, and `compareScenarios2`, the export with [`xlsx_IIASA`](https://github.com/remindmodel/remind/blob/develop/scripts/output/export/xlsx_IIASA.R) and every other scripts fail to find the data.

An alternative option I could imagine is for the reporting to check whether the file in `cfg$pathToMagpieReport` exists and always add that content directly to the `.mif` file. Further ideas?

dklein-pik:
Good idea. If `cfg$pathToMagpieReport` exists, REMIND has been started using MAgPIE data from this report, so why not always append the corresponding MAgPIE report. However, the code in `start_coupled.R` you cited above and that currently appends the MAgPIE report to the remind report should then be removed.

merfort:
Sounds like a neat solution. Moving the concatenation of reportings to the reporting procedure itself seems to be very natural.

PR #992, opened 2022-10-04
append MAgPIE results in reporting.R, not in start_coupled.R

orichters:
## Purpose of this PR
- fixes https://github.com/remindmodel/remind/issues/988
- moves concatenating MAgPIE results to REMIND mif from `start_coupled.R` to `reporting.R`
- improve error message if `magpie4::reportEmissions()` failed for some reason

## Type of change

- [x] Bug fix 
- [x] Refactoring

## Checklist:

- [x] The model compiles and runs successfully: see `/p/tmp/oliverr/remind-smallfix/output/test_*` folders with coupled runs, also starting `reporting` via `output.R` worked for both coupled and non-coupled run.

### DIFF
--- a/scripts/output/single/reporting.R
+++ b/scripts/output/single/reporting.R
@@ -25,7 +25,7 @@ if(!exists("source_include")) {
 
 gdx      <- file.path(outputdir,gdx_name)
 gdx_ref  <- file.path(outputdir,gdx_ref_name)
-if(!file.exists(gdx_ref)) { gdx_ref <- NULL }
+if (!file.exists(gdx_ref)) { gdx_ref <- NULL }
 scenario <- getScenNames(outputdir)
 ###############################################################################
 # paths of the reporting files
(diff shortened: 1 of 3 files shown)


scripts/output/single/reporting.R
```diff
@@ -77,12 +77,31 @@ message("start generation of EDGE-T reporting")
...
 
-  write.mif(EDGET_output, remind_reporting_file, append=T)
-  deletePlus(remind_reporting_file, writemif=T)
+  write.mif(EDGET_output, remind_reporting_file, append = TRUE)
+  deletePlus(remind_reporting_file, writemif = TRUE)
 
 message("end generation of EDGE-T reporting")
 }
 
+configfile <- file.path(outputdir, "config.Rdata")
```
dklein-pik, line 86: I realize only now that this has the disadvantage of making reporting.R dependent on the existence of the config file. There is, however, no other way, unless we write the path to the magpie report into the gdx, which is a bit too much for now, I think. So, long story short: I'm fine with this solution :-)

dklein-pik (approved):

scripts/output/single/reporting.R
orichters: `reporting.R` depends on `config.Rdata` anyhow: It calls [`lucode2::getScenNames()`](https://github.com/pik-piam/lucode2/blob/master/R/getScenNames.R) which reads from it.

dklein-pik: Ah, missed that. Thanks!

PR #997, opened 2022-10-10
Introduce technology category tech_stat 4 for electrolysis and DAC, adapt their initial parameters for learning and document it more clearly

fschreyer:
# Purpose of this PR

This introduces the technology category tech_stat = 4 that comprises technologies in a currently very early development stage. It prevents buildings these technologies before 2025. Moreover, learning parameters of these technologies are initialized using the cumulated capacity (ccap0) and initial investment cost (inco0) in generisdata_tech.prn based on what we expect for 2025. Learning only starts in 2025 for these technologies because the learning formulation would induce too fast learning when starting at too low initial capacities of 2020 (as it was before). In addition, I added some documentation in the generisdata file on how and why I set the learning parameters as they are. 

## Type of change

## Checklist:

Maybe interesting to @katarkow and @aodenweller as we talked about related issues at some point. 

* Test runs are here: 

/p/tmp/schreyer/Modeling/remind/Current/output/Base_CCap0Up_2022-10-08_20.20.16

/p/tmp/schreyer/Modeling/remind/Current/output/NDC_CCap0Up_2022-10-09_01.58.38

/p/tmp/schreyer/Modeling/remind/Current/output/PkBudg500_CCap0Up_2022-10-09_09.07.47


LaviniaBaumstark (approved):

aodenweller (commented):
Hi @fschreyer, there's a typo here as it should say 1200 EUR/kW(el) not GW(el). More importantly, if 1200 EUR/kW is the cost in 2025, that's quite high I think. Do you have a source? Also, I wouldn't use a floor cost at all. This could also be overly restrictive. If learning only starts in 2025, a floor cost shouldn't be necessary. Maybe @FalkoUeckerdt also has something to add.

aodenweller:
On the other hand, 20 GW globally in 2025 is already quite optimistic. The interquartile range of our analysis is 5-8 GW by 2025 globally. Even under emergency growth it might be hard to get to 20 GW by 2025.

fschreyer:
Thanks for chiming in, Adrian and Falko. Sry, I was only looking at Figure 1d of your paper, Adrian, and probably misunderstood what is shown. The CAPEX were based on the Figure of Falko's paper Fig. S3. I am a bit surprised about the low floor cost. I will update the numbers in a different PR.

robertpietzcker:
as long as the learning rate stays at 15%, having 50 or 100 $/kW should be fine. The higher it goes the more critical the floor cost value becomes

fschreyer:
> as long as the learning rate stays at 15%, having 50 or 100 $/kW should be fine. The higher it goes the more critical the floor cost value becomes

Well, that's only true if the initial cumulative capacity (ccap0) isn't too low which was what this change was mainly about. With 15% you half the difference to the floor cost for every 8-fold increase of cum. capacity.

PR #998, opened 2022-10-11
renv_init.log

pfuehrlich-pik:
# Purpose of this PR

renv was cluttering the log during initialization of each run renv. Now this is written into the run folder in renv_init.log instead.

## Type of change

## Checklist:


orichters (approved):
Great, thanks. I may prefer log-renv.txt so it is more easily found, but no strong opinion on that.

By the way: is this if clause correct or should it be the other way round? 

https://github.com/remindmodel/remind/blob/415b1be6ab18f39c85c559c6764d9314f2a09539/scripts/start/submit.R#L60

orichters (approved):
Great, thanks. I may prefer log-renv.txt so it is more easily found, but no strong opinion on that.

By the way: is this if clause correct or should it be the other way round? 

https://github.com/remindmodel/remind/blob/415b1be6ab18f39c85c559c6764d9314f2a09539/scripts/start/submit.R#L60

pfuehrlich-pik:
Yes, the if-clause is correct. If the autoRenvUpdates option is set to TRUE we run the update script. If it is unset we use the default value FALSE, and just run the check which packages could be updated.

PR #999, opened 2022-10-11
checkSetup.R: Do not depend on non-existing R package

mikapfl:
# Purpose of this PR

In some circumstances (notably, when using renv on the cluster), `checkSetup.R` detects `"R"` as a package dependency. But "R" is not an R package, and therefore it ends with the nonsensical recommendation to run `renv::install("R")`. With this PR, this is explicitly filtered out.

## Type of change

- [x] Bug fix 
- [x] Minor change

## Checklist:

No change in model, only in the util R script, so I didn't check any modelruns.

### DIFF
--- a/scripts/utils/checkSetup.R
+++ b/scripts/utils/checkSetup.R
@@ -8,7 +8,10 @@ stopifnot(`pdflatex not found, check your LaTeX installation` = Sys.which("pdfla
 message("checking for pdflatex executable on your PATH - ok")
 
 message("checking if required R packages are installed")
-missingDeps <- Filter(function(x) !requireNamespace(x, quietly = TRUE),
+filter_func <- function(x) {
+  return(!requireNamespace(x, quietly = TRUE) && x != "R")
+}
+missingDeps <- Filter(filter_func,
                       renv::dependencies(dev = TRUE)[, "Package"])
 if (length(missingDeps) > 0) {
   stop("Some required R packages are missing, install them with `renv::install(",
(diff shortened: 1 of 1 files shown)


scripts/utils/checkSetup.R
```diff
@@ -8,7 +8,10 @@ stopifnot(`pdflatex not found, check your LaTeX installation` = Sys.which("pdfla
 message("checking for pdflatex executable on your PATH - ok")
 
 message("checking if required R packages are installed")
-missingDeps <- Filter(function(x) !requireNamespace(x, quietly = TRUE),
+filter_func <- function(x) {
+  return(!requireNamespace(x, quietly = TRUE) && x != "R")
+}
+missingDeps <- Filter(filter_func,
                       renv::dependencies(dev = TRUE)[, "Package"])
```
pfuehrlich-pik, line 15:
```suggestion
missingDeps <- Filter(function(x) !requireNamespace(x, quietly = TRUE),
                      setdiff(renv::dependencies(dev = TRUE)[, "Package"], "R"))
```

pfuehrlich-pik: `setdiff` also applies `unique`, so no package is checked twice. There is no linter in remind (yet), but I'd prefer if we could stick to camel case anyway, so I'd suggest renaming `filter_func` to `filterFunc` if you want to stick to that solution.

mikapfl: `setdiff` here works nicely as well. For me, `x != "R"` is a bit more readable, but that's probably just because I am *always* confused when seemingly scalar values in R are treated as sets or vectors. How can "R" be a set, it is just a single string! :wink:

pfuehrlich-pik (approved):

Issue #918, opened 2022-07-22
exoGAINS overwrites fulldata.gdx in the middle of an iteration


orichters:
Problem:
- if the run is killed because of time-out or whatever: You would expect that the fulldata.gdx is from the last finished iteration. It isn't, as [11_aerosols/exoGAINS/postsolve.gms](https://github.com/remindmodel/remind/blob/b5eafdb64a6b1db54790d6241fda7054dfa710c8/modules/11_aerosols/exoGAINS/postsolve.gms#L15) overwrites it in the middle of an iteration. That does not seem to be a very sensible idea to me, although this is the case since this repo here exists.
- we may want to save the data to a different file and [load it from there](https://github.com/orichters/remind/blob/develop/scripts/input/exoGAINSAirpollutants.R#L33). Note also that exoGAINS writes a non_optimal.gdx, but never ever loads it.

Tagging @rauner and @schreyer who last worked on this and @LaviniaBaumstark of course.

PR #1005, opened 2022-10-17
avoid that 11_exoGAINS overwrites fulldata.gdx, improve restart message

orichters:
## Purpose of this PR
- Fixes two issues with 11_exoGAINS: first, that it overwrites `fulldata.gdx` in the middle of an iteration, causing some mess on restarts sometimes. #918. Second, that after non-optimal solve in early iterations, no fulldata.gdx is found and it crashes. Solution is to write an own gdx file that is only used by exoGAINS and later deleted
- improve some logging, preserve `log.txt` as `log_beforeRestart.txt` if runs are restarted and tell the user about it

## Type of change

- [x] Bug fix 
- [x] No change in scenario results

## Checklist:

- [x] The model compiles and runs successfully (see `/p/tmp/oliverr/remind/output/default_2022-10-06_17.35.04`)

### DIFF
--- a/modules/11_aerosols/exoGAINS/postsolve.gms
+++ b/modules/11_aerosols/exoGAINS/postsolve.gms
@@ -13,8 +13,10 @@
 *** run exoGAINS from iteration 2 onwards to avoid incomplete GDX files when running it in the first iteration
 if (iteration.val ge 2,
 
-*** write data to file
-Execute_Unload 'fulldata_exoGAINS';
+*** write data to file if an optimal solution was found
+if((o_modelstat le 2),
+    Execute_Unload 'fulldata_exoGAINS';
+);
 
 *** Calculate AP emissions
 Execute "Rscript exoGAINSAirpollutants.R";
--- a/modules/11_aerosols/exoGAINS/postsolve.gms
+++ b/modules/11_aerosols/exoGAINS/postsolve.gms
@@ -12,16 +12,11 @@
 
 *** run exoGAINS from iteration 2 onwards to avoid incomplete GDX files when running it in the first iteration
 if (iteration.val ge 2,
-*** write the fulldata.gdx file after each optimal iteration
-***AJS* in Nash status 7 is considered optimal in that respect (see definition of o_modelstat in solve.gms)
-if((o_modelstat le 2),
-    Execute_Unload 'fulldata';
-  );  
-if((o_modelstat gt 2),
-    Execute_Unload 'non_optimal';
-  );
 
-*** Calcualte AP emissions
+*** write data to file
+Execute_Unload 'fulldata_exoGAINS';
+
+*** Calculate AP emissions
 Execute "Rscript exoGAINSAirpollutants.R";
 
 *** Read input ref results for tall with following dimensions: pm_emiAPexsolve(tall,all_regi,all_sectorEmi,emiRCP)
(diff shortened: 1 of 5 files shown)


merfort:
Thanks Oli! 
Regarding the problem that overwriting `fulldata.gdx` this fix should work.

I am just a bit confused regarding the part of the missing `fulldata.gdx` in case of an early non-optimal solve. As far as I understand it, the way it **worked before** was, that REMIND writes `non_optimal.gdx` in case of a non-optimal solution and `exoGAINSAirpollutants.R` in that case falls back to `input.gdx` instead (if there is no `fulldata.gdx` from a previous iteration). So the problem is not that no `fulldata.gdx` was found, but rather that there actually is a `fulldata.gdx` that somehow contains some values that break the reporting, probably due to the fact that the iteration was non-optimal. At least in my case (`/p/tmp/merfort/REMIND_development/remind/output/SSP2EU-PkBudg500_sustTax_1p5_EFTtax_20_glob_v2_2022-10-13_17.47.17`) the problem is that using `fulldata.gdx` breaks the reporting within `exoGAINSAirpollutants.R`, while using `input.gdx` would work (I checked that). I just don't know, why the `fulldata.gdx` that was written after the second iteration breaks the reporting, since the second iteration was actually feasible...

To cut a long story short, in theory the original implementation (only reporting when optimal) should work, and I would actually rather not write a `fulldata_exoGAINS.gdx` after a non-optimal solution, since this might actually break the reporting, while using `input.gdx` doesn't.
Or am I getting this completely wrong?

A minor point: I think it would be nice if exoGAINSAirpollutants.R wrote something like "No fulldata_exoGAINS.gdx found, using input.gdx instead" in that case, that would help with debugging in the future.

merfort:
Btw, I pulled from your branch and reran my scenario, but the bug persists.

orichters:
I have to admit, @merfort, then I don't understand the problem at all. Because exoGAINS was never run on a gdx file from an non_optimal iterations. Because if it did not converge, it was writing to `non_optimal.gdx` but always loading from `fulldata.gdx` or `input.gdx`. So I don't know at all why your runs fail.

merfort:
Ok, I did some further tests and I think, I can clarify this. There are two unrelated issues:
1) The problem with my runs is not related to the missing of a `fulldata.gdx`. The cause here is most likely that I used some (too) old NDC runs as reference (for pre-2025 values), which somehow breaks the reporting. It just already emerges at REMIND runtime, because `exoGAINSAirpollutants.R` calls some reporting functions (in particular `reportFE`). That is a bit inconvenient, since you usually don't want a broken reporting to also break REMIND itself, but I would not know, how to prevent that, since `exoGAINSAirpollutants.R` apparently relies on some reporting variables. Sorry for the confusion that I introduced with my runs!
2) With your updates in `modules/11_aerosols/exoGAINS/postsolve.gms` REMIND will now always write a gdx file that can be used for `exoGAINSAirpollutants.R` , also in the case of a non-optimal solution. I think that may potentially lead to errors within the reporting functions called within `exoGAINSAirpollutants.R`. So, my suggestion would be to write `fulldata_exoGAINS.gdx` only in the case of an optimal solution, i.e. keep the condition `if((o_modelstat le 2), ...`. Consequently REMIND will fallback to `input.gdx` if it there are non-optimal solutions in the first iteration(s) and to the last successfully written `fulldata_exoGAINS.gdx` in case of non-optimal solutions in later iterations.

merfort (approved):
Thanks! Looks all good now.

PR #1012, opened 2022-10-17
Improve description of early retirement switches

stephenbi:
## Type of change

- [x] Documentation


stephenbi:
For c_tech_early_reti_rate, how about: 
"maximum percentage of technology-specific capital stock that can be early retired in one year in each region, overriding c_regi_earlyreti_rate. Use this switch to enable stringent phase-out policy enforcement."

cchrisgong:
Hmm, I wouldn't say "Use this switch to enable stringent phase-out policy enforcement." because I usually use it to enable more relaxed policies (for pc for example, because pc is never economical in REMIND)..

cchrisgong:
perfect :) as long as they are clear, they can be a bit long i think

cchrisgong (approved):

PR #1016, opened 2022-10-17
use piamInterfaces to generate mappings

orichters:
## Purpose of this PR
- refactor `xlsx_IIASA.R`, called via `output.R` -> `export`
- if no mapping file is provided, generate one using `piamInterfaces`
- improves the use of `iiasatemplate` such that also units are checked and possibly corrected
- write more to the log file about what has been done
- @pfuehrlich-pik: do I have to add piamInterfaces elsewhere to the dependencies? I added a line to the renv tutorial as well that I found useful
- requires https://github.com/pik-piam/piamInterfaces/pull/14

## Type of change

- [x] Refactoring
- [x] no impact on default scenarios

## Checklist:

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -48,7 +48,7 @@ Imports:
     ncdf4,
     nleqslv,
     optparse,
-    piamInterfaces,
+    piamInterfaces (>= 0.0.17),
     plotly,
     purrr,
     quitte,
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -48,6 +48,7 @@ Imports:
     ncdf4,
     nleqslv,
     optparse,
+    piamInterfaces,
     plotly,
     purrr,
     quitte,
--- a/scripts/output/export/xlsx_IIASA.R
+++ b/scripts/output/export/xlsx_IIASA.R
@@ -18,22 +18,17 @@ library(stringr) # str_sub, str_split
 library(tibble)
 library(tidyr)
 
-meslog <- function(text, LOGFILE) {
-  message(text)
-  write(text, file = LOGFILE, append = TRUE)
-}
-
 options(warn = 1)
 
-MODEL <- "REMIND 3.0"                            # modelname in final file
-REMOVE_FROM_SCEN <- ""                           # you can use regex such as: "_diff|_expoLinear"
-ADD_TO_SCEN <- NULL                              # is added at the beginning
+model <- "REMIND 3.0"                            # modelname in final file
+removeFromScen <- ""                           # you can use regex such as: "_diff|_expoLinear"
+addToScen <- NULL                              # is added at the beginning
 
 # filenames relative to REMIND main directory (or use absolute path) 
-MAPPING <- NULL                                  # file obtained from piamInterfaces, or AR6/SHAPE/NAVIGATE or NULL to get asked
+mapping <- NULL                                  # file obtained from piamInterfaces, or AR6/SHAPE/NAVIGATE or NULL to get asked
 iiasatemplate <- "yaml_or_xlsx_file_from_IIASA"  # provided for each project, can be yaml or xlsx with a column 'Variable'
 
-# note: you can also pass all these options to output.R, so 'Rscript output.R LOGFILE=mylogfile.txt' works.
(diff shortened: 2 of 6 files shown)


fbenke-pik (approved):

scripts/output/export/xlsx_IIASA.R
```diff
@@ -9,117 +9,134 @@ require(data.table)
...
 library(magpie4)
-library(stringr) # str_sub
+library(piamInterfaces)
+library(stringr) # str_sub, str_split
+library(tibble)
+library(tidyr)
+
+meslog <- function(text, LOGFILE) {
+  message(text)
+  write(text, file = LOGFILE, append = TRUE)
```
pfuehrlich-pik, line 23: There's also `sink` which can achieve a similar thing: Run `sink(LOGFILE, append = TRUE, type = "message", split = TRUE)`, then all calls to `message` will write to terminal (actually `stderr`) and the logfile. You must remember to call `sink()` at the end of the script/when you want to stop redirecting output to the logfile.

pfuehrlich-pik (approved):

DESCRIPTION
```diff
@@ -48,6 +48,7 @@ Imports:
     ncdf4,
     nleqslv,
     optparse,
+    piamInterfaces (>= 0.0.14),
```
pfuehrlich-pik, line 51: This looks good.

scripts/output/export/xlsx_IIASA.R
orichters: `split = TRUE` only splits R output (via Rvprintf) and the default output from writeLines: it does not split all output that might be sent to stdout. (https://stat.ethz.ch/R-manual/R-devel/library/base/html/sink.html)

So I don't think it works here…

pfuehrlich-pik: Could be, although in my experience sink usually works as expected. I'm a little afraid that we will see `meslog` or similar functions defined over and over in different scripts, in that case I'd much prefer pulling meslog up into e.g. the gms package.

orichters: please try `sink(file("test.log", "w"), type = "message", split = TRUE)` (without the `file` part, it also crashes):
```
Error in `sink()`:
! cannot split the message connection
```
Happy to add `meslog` or something similar to `gms`.

pfuehrlich-pik: Okay I'm convinced `sink` - does not like messages. I just remembered another way:
```r
withCallingHandlers({
  # all the code that should message to console and log
}, message = function(x) {
  cat(x$message, file = LOGFILE, append = TRUE)
  message(x$message) # not sure, maybe need to use cat here
})
```
This is better than the sink approach anyway, because you cannot forget to run `sink()` at the end. This is also similar to what [madrat does](https://github.com/pik-piam/madrat/blob/master/R/withMadratLogging.R).

scripts/output/export/xlsx_IIASA.R
```diff
@@ -9,234 +9,221 @@ require(data.table)
...
 
-data %>%
-  mutate(value = ifelse(!is.finite(value) | paste(value) == "", 0, value)) %>%
-  pivot_wider(names_from="period", values_from="value") -> writetoexcel
+  message("\n### mif files: ", sum(existingfiles), " used, ", sum(! existingfiles), " not found, see above.")
+  message("\n### Output files written:\n- ", OUTPUT_mif, "\n- ", OUTPUT_xlsx)
 
-writexl::write_xlsx(list("data" = writetoexcel), OUTPUT_xlsx)
+}, message = function(x) {
+  cat(x$message, file = logFile, append = TRUE)
```
pfuehrlich-pik, line 226: Will this also still print to console or only to logFile?

orichters: it prints also to console. I don't know exactly, why, to be fair, but what you suggested produced two prints to the console.

pfuehrlich-pik: Okay good. I'd guess there's a chain of handlers, and after your custom handler it will just continue with the default handler which prints to console.

PR #1023, opened 2022-10-19
delete outdated options for nucscen, keep only number 2 (default), 5 and 6 

LaviniaBaumstark:
# Purpose of this PR
cleaning options of nucscen

## Type of change

- [x] Refactoring

## Checklist:

* Test runs are here: 

### DIFF
--- a/modules/40_techpol/NDC/bounds.gms
+++ b/modules/40_techpol/NDC/bounds.gms
@@ -13,11 +13,6 @@ vm_cap.lo(t,regi,"tnrs","1")$(t.val ge 2025) = p40_TechBound(t,regi,"tnrs")*0.00
 vm_cap.lo(t,regi,"hydro","1")$(t.val ge 2025) = p40_TechBound(t,regi,"hydro")*0.001;
 
 
-* FS: in case of a nuclear phase-out scenario (nucscen 7), nuclear lower bound from p40_techBound only up to 2025
-if(cm_nucscen eq 7,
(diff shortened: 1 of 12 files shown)


modules/40_techpol/NDC/bounds.gms
```diff
@@ -31,8 +26,6 @@ if(cm_nucscen eq 5,
   vm_cap.lo(t,regi_nucscen,"tnrs","1")$(t.val gt 2020) = 0;
 );
```
mikapfl, line 27: why do we have this only in NDC and not in the other realizations of 40_techpol? And isn't this redundant with the bound in bounds.gms which applies starting 2015?
https://github.com/LaviniaBaumstark/remind/blob/69cb5f61eafc3a5bab0f25dfe5e4afb27934a7af/core/bounds.gms#L222

LaviniaBaumstark: I assume that this is specific for some national plans and the other realizations are not as up to date? 
In core/bounds.gms it is only active for nucscen=5

mikapfl: here in NDC/bounds it is also only active for nucscen=5, right?

LaviniaBaumstark: hm, right, I will also delete this part...

mikapfl (approved):

PR #1028, opened 2022-10-21
Check Dependencies

pfuehrlich-pik:
# Purpose of this PR
With this PR, dependency version requirements from DESCRIPTION are checked at the beginning of each R session via `lucode2::checkDeps()` in `.Rprofile`. If an installed package version is older than the required version there is a warning and instructions to update.

Sidenote: I re-added the `Package: remindmodel` line to DESCRIPTION, because some tools (e.g. `pak`) require this line to actually use the DESCRIPTION.

## Type of change

- [ ] Bug fix 
- [ ] Refactoring
- [x] New feature 
- [ ] Fundamental change

## Checklist:


0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q (approved):

mikapfl:
Hi,

thanks, I have a couple of questions.
1. This doesn't actually stop people, just emit a warning. In a test run with a too-old gms, this warning can be easily missed, the slurm config question proceeds as usual and then the run breaks due to missing options in gms. Should we instead check the dependencies in the start scripts before submitting and just stop if the dependencies are not fulfilled?
2. What would be the one-stop solution to get to a "good" state? In my repo, I always get the warning `  goxygen is required, but not installed - please install it.`, but I don't seem to find a command to just automatically install everything that is required. It is annoying to have to `renv::install("goxygen")` by hand.

Cheers

Mika

pfuehrlich-pik:
1. Could you ask that question in the next remind meeting?
2. There's probably multiple ways, each with their own pitfalls. One way would be [renv::install()](https://rstudio.github.io/renv/reference/install.html#project-description-files) without arguments.

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
> Should we instead check the dependencies in the start scripts before submitting 

Yes.  No need to ask in the REMIND meeting, that will just devolve into 15 minutes of pointless discussion.

> and just stop if the dependencies are not fulfilled?

I think `lucode2::checkDeps()` and `lucode2::checkRequirement()` could be improved in some regards.  I would suggest to 
- have `checkRequirement()` just return the state of the requirement, either `TRUE` if it is met, or what goes currently into the warning ("package not installed" or "requirement not met")
- have `checkDeps()` take an argument on what to do:
  - `"warn"`: basically what it does now, but without the call signature from somewhere inside `Map()` which is likely to confuse people
  - `"stop"`: throw an error; what @mikapfl is getting at
  - `"pass"`: invisible return the state of all packages (for use in other functions)

I can implement that, but didn't want to just start fiddling with it.  Is there any code dependent on `checkDeps()` besides this yet? @pfuehrlich-pik

pfuehrlich-pik:
Sounds good, feel free to go ahead and implement that. `checkDeps()` is not used anywhere yet.

orichters:
Sounds good. Maybe add `ask` as option 4, asking the user whether he wants to run the necessary update now but also allows him to just say `no`.

mikapfl:
> Sounds good, feel free to go ahead and implement that. `checkDeps()` is not used anywhere yet.

Cool, will do that.

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
> Sounds good. Maybe add `ask` as option 4, asking the user whether he wants to run the necessary update now but also allows him to just say `no`.

Can you put that into a [feature request](https://github.com/pik-piam/lucode2/issues/new)?  This will require ~~us~~ Pascal to figure out how to install specific package versions automatically.

mikapfl:
We'll do this via start scripts instead, see https://github.com/remindmodel/remind/pull/1030

PR #1037, opened 2022-10-25
prevent updating renv package

pfuehrlich-pik:
# Purpose of this PR
`make update-all-renv` would also update the renv package itself, resulting in changes to e.g. `renv/activate.R`. We discussed [here](https://github.com/remindmodel/remind/pull/965) that the renv package should not be updated unless there is a good reason, so this PR prevents updating renv via `make update-all-renv`. New dependencies will now no longer be installed with that command (because it's tricky to prevent updating renv in the process), but https://github.com/remindmodel/remind/pull/1030 will add checks and probably a convenient way to install new dependencies anyway.

## Type of change

- [x] Bug fix 

## Checklist:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,8 +15,7 @@ update-renv:    ## Upgrade all pik-piam packages in your renv to the respective
 
 update-all-renv: ## Upgrade all packages (including CRAN packages) in your renv
                  ## to the respective latest release, make new snapshot
-	Rscript -e 'renv::update()'
-	Rscript -e 'renv::install()'
+	Rscript -e 'renv::update(exclude = "renv")'
 	Rscript -e 'renv::snapshot()'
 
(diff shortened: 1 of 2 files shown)


cchrisgong:
thanks! 

does make update-renv also need to be disabled?

pfuehrlich-pik:
Good catch! `renv::install()` is indeed a problem, because it will not only install missing deps, but instead it will install and also update all deps from DESCRIPTION. We don't want to install updates for CRAN packages in updateRenv, so I removed it. `make update-renv` just runs the updateRenv.R script, so it is fine now.

mikapfl (approved):
R dependency management is hard. Sorry, apparently I didn't test `renv::install` properly, I thought I explicitly tested if it only installs new packages, but probably my test renv was fully updated…

cchrisgong (approved):

PR #1038, opened 2022-10-26
fix wrong calibration results being written out after the second calibration iteration

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
fix #954 where wrong calibration results where written out after the second calibration iteration; clean up some bits and pieces


orichters (approved):
thanks!

PR #1040, opened 2022-10-26
fix bug in combine_slurmConfig

orichters:
## Purpose of this PR

``` R
slurmOld <- "--qos=standby --nodes=1 --tasks-per-node=12"
slurmNew <- "--qos=medium --nodes=5 --mem=8000"
combine_slurmConfig(slurmOld, slurmNew)
```
currently yields:
``` R
"--qos=medium --nodes=5 --mem=8000 --nodes=1 --tasks-per-node=12"
```
I don't think this is intended. After this PR, it returns:
``` R
"--qos=medium --nodes=5 --mem=8000 --tasks-per-node=12"
```
This is also how I understand the comment above the code.

Tagging @dklein-pik who [introduced this function](https://github.com/remindmodel/remind/commit/b69f943c55bd0011288fb40994a2c0f1e0fb3b2a)

### DIFF
--- a/scripts/start/choose_slurmConfig.R
+++ b/scripts/start/choose_slurmConfig.R
@@ -107,7 +107,7 @@ combine_slurmConfig <- function (original, update_with) {
   names(v_original) <- gsub("--(.*)=.*","\\1",unlist(strsplit(original,split=" ")))
 
   # remove elements from "original" that are existing in "update_with"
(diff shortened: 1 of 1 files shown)


mikapfl (approved):
Looks good! If only we had unit tests in remind, then we could make sure things work as expected…

PR #1044, opened 2022-11-01
poll environment variable REMIND_sync_renv on whether to synv renv or not

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
- [x] New feature 

As per https://github.com/remindmodel/remind/pull/808#issuecomment-1128674710

## Checklist:

- [ ] My code follows the coding etiquette

### DIFF
--- a/scripts/start/submit.R
+++ b/scripts/start/submit.R
@@ -48,17 +48,28 @@ submit <- function(cfg, restart = FALSE, stopOnFolderCreateError = TRUE) {
       # detected like this
       firstRunInCascade <- normalizePath(renv::project()) == normalizePath(".")
       if (firstRunInCascade) {
-        capture.output({ # suppress printing status
-          renvStatus <- renv::status()
-        })
-        if (!renvStatus$synchronized) {
-          message("The new run will use the package environment defined in renv.lock, ",
-                  "but it is out of sync, probably because you installed packages/updates manually. ",
-                  "Write current package environment into renv.lock first? (Y/n)", appendLF = FALSE)
-          if (tolower(gms::getLine()) %in% c("y", "")) {
-            renv::snapshot(prompt = FALSE)
+          capture.output({ # suppress printing status
+              renvStatus <- renv::status()
+          })
+
+          if (!renvStatus$synchronized) {
+              # check environment variable on whether to syncronise renv
+              Rrs <- Sys.getenv('REMIND_sync_renv')
+
+              # if neither 'always' nor 'never', ask the user
+              if (!Rrs %in% c('always', 'never')) {
+                  message("The new run will use the package environment defined in renv.lock, ",
+                          "but it is out of sync, probably because you installed packages/updates manually. ",
+                          "Write current package environment into renv.lock first? (Y/n)", appendLF = FALSE)
+                  if (tolower(gms::getLine()) %in% c("y", "")) {
+                      Rrs <- 'always'
+                  }
+              }
+              # if always or user answered 'yes', do it
+              if ('always' == Rrs) {
+                  renv::snapshot(prompt = FALSE)
+              }
           }
-        }
 
         if (getOption("autoRenvUpdates", FALSE)) {
           source("scripts/utils/updateRenv.R")
@@ -94,15 +105,15 @@ submit <- function(cfg, restart = FALSE, stopOnFolderCreateError = TRUE) {
                stdout = renvLogPath, stderr = "2>&1")
     }
 
-    # Save the cfg (with the updated name of the result folder) into the results folder. 
-    # Do not save the new name of the results folder to the .RData file in REMINDs main folder, because it 
+    # Save the cfg (with the updated name of the result folder) into the results folder.
+    # Do not save the new name of the results folder to the .RData file in REMINDs main folder, because it
     # might be needed to restart subsequent runs manually and should not contain the time stamp in this case.
     filename <- file.path(cfg$results_folder, "config.Rdata")
     cat("   Writing cfg to file", filename, "\n")
     # remember main folder
     cfg$remind_folder <- normalizePath(".")
     save(cfg, file = filename)
-    
+
     # Copy files required to configure and start a run
     filelist <- c("prepare_and_run.R" = "scripts/start/prepare_and_run.R",
                   ".Rprofile" = ".Rprofile")
@@ -114,8 +125,8 @@ submit <- function(cfg, restart = FALSE, stopOnFolderCreateError = TRUE) {
   on.exit(setwd(cfg$remind_folder))
   # Change to run folder
   setwd(cfg$results_folder)
-  
-  # send prepare_and_run.R to cluster 
+
+  # send prepare_and_run.R to cluster
   cat("   Executing prepare_and_run.R for",cfg$results_folder,"\n")
   if (grepl("^direct", cfg$slurmConfig)) {
     log <- format(Sys.time(), paste0(cfg$title,"-%Y-%H-%M-%S-%OS3.log"))
@@ -124,6 +135,6 @@ submit <- function(cfg, restart = FALSE, stopOnFolderCreateError = TRUE) {
     system(paste0("sbatch --job-name=",cfg$title," --output=log.txt --mail-type=END --comment=REMIND --wrap=\"Rscript prepare_and_run.R \" ",cfg$slurmConfig))
     Sys.sleep(1)
   }
-    
+
   return(cfg$results_folder)
 }


mikapfl:
Hm, I'm struggling to see how the choices presented here are useful.

Generally, independent of `REMIND_sync_renv`, there will be no upgrades or new packages installed unless you explicitly do it via `renv::install` or one of the `make` commands or accepting an automatic upgrade triggered by a dependency in the DESCRIPTION or any other number of package management commands.

Depending on the way you do any of the package manipulations, a `renv::snapshot` is called or not
* if you use `make upgrade-renv` or `make upgrade-all-renv`, `renv.lock` is updated via `renv::snapshot`
* if you accept an automatic upgrade triggered by a dependency in the DESCRIPTION, `renv.lock` is updated via `renv::snapshot`
* if you run `renv::install` or similar command manually, and *do not* call `renv::snapshot`, then the renv and the `renv.lock` is out of date and the sync-question will arise and `REMIND_sync_renv` gives you a choice.

Did I understand that correctly? In normal operation, this should be an edge case which almost never happens. Maybe you see the need for this because these syncs are somehow triggered more often than they should be?

scripts/start/submit.R
```diff
@@ -48,17 +48,28 @@ submit <- function(cfg, restart = FALSE, stopOnFolderCreateError = TRUE) {
...
-        capture.output({ # suppress printing status
-          renvStatus <- renv::status()
-        })
-        if (!renvStatus$synchronized) {
-          message("The new run will use the package environment defined in renv.lock, ",
-                  "but it is out of sync, probably because you installed packages/updates manually. ",
-                  "Write current package environment into renv.lock first? (Y/n)", appendLF = FALSE)
-          if (tolower(gms::getLine()) %in% c("y", "")) {
-            renv::snapshot(prompt = FALSE)
+          capture.output({ # suppress printing status
```
pfuehrlich-pik, line 51: Seems like you were indenting with 4 spaces, while the rest of the file is indented with 2. We should decide for one or the other and stick to that consistently.

scripts/start/submit.R
```diff
@@ -48,17 +48,28 @@ submit <- function(cfg, restart = FALSE, stopOnFolderCreateError = TRUE) {
...
+              renvStatus <- renv::status()
+          })
+
+          if (!renvStatus$synchronized) {
+              # check environment variable on whether to syncronise renv
+              Rrs <- Sys.getenv('REMIND_sync_renv')
```
pfuehrlich-pik, line 57: Also for consistency, I'd suggest we use an option for this, just like for the autoRenvUpdates:
```suggestion
              Rrs <- getOption("remindSyncRenv", "ask")
```

scripts/start/submit.R
```diff
@@ -48,17 +48,28 @@ submit <- function(cfg, restart = FALSE, stopOnFolderCreateError = TRUE) {
...
+              if (!Rrs %in% c('always', 'never')) {
+                  message("The new run will use the package environment defined in renv.lock, ",
+                          "but it is out of sync, probably because you installed packages/updates manually. ",
+                          "Write current package environment into renv.lock first? (Y/n)", appendLF = FALSE)
+                  if (tolower(gms::getLine()) %in% c("y", "")) {
+                      Rrs <- 'always'
+                  }
+              }
+              # if always or user answered 'yes', do it
+              if ('always' == Rrs) {
```
pfuehrlich-pik, line 69: I'd suggest a more verbose variable name:
```suggestion
              remindSyncRenv <- getOption("remindSyncRenv", "ask")

              # if neither 'always' nor 'never', ask the user
              if (!remindSyncRenv %in% c('always', 'never')) {
                  message("The new run will use the package environment defined in renv.lock, ",
                          "but it is out of sync, probably because you installed packages/updates manually. ",
                          "Write current package environment into renv.lock first? (Y/n)", appendLF = FALSE)
                  if (tolower(gms::getLine()) %in% c("y", "")) {
                      remindSyncRenv <- 'always'
                  }
              }
              # if always or user answered 'yes', do it
              if ('always' == remindSyncRenv) {
```
If you want to stick to the three letter name, please rename to `Rsr` to match the envvar/option name.

pfuehrlich-pik (commented):
Seems like a useful feature to me, mostly for people using renv functions directly instead of the provided scripts. I'd still suggest adding a note to the remind renv tutorial, so everyone is aware of this feature.

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
> Did I understand that correctly?

No.  Every single time I start REMIND – and I start REMIND a lot – this code
https://github.com/remindmodel/remind/blob/a9197932d889d6e37ae33f58214157d8dd1a095d/scripts/start/submit.R#L51-L53
is run, ~~reporting~~ checking version differences of packages in the renv lock file and my user library, e.g.
> ```
> > renv::status()
> The following package(s) are out of sync:
> 
>        Package   Lockfile Version   Library Version
>  edgeTransport             0.21.1            0.22.1
>        mrwater             1.3.28            1.3.30
> 
> Use `renv::snapshot()` to save the state of your library to the lockfile.
> Use `renv::restore()` to restore your library from the lockfile.
> ```
which usually are there, because I update my R packages regularly, and then nagging me about whether I want to update the lock file
https://github.com/remindmodel/remind/blob/a9197932d889d6e37ae33f58214157d8dd1a095d/scripts/start/submit.R#L54-L61
which I never want to do.

So, pursuant to https://github.com/remindmodel/remind/pull/808#issuecomment-1128674710, I want a way to opt out of that.
If I want to update the renv, I'll do so myself, thankyouverymuch.

mikapfl:
> because I update my R packages regularly

But that means you update the R package in the remind main renv, right? So, running `update.packages` in the remind main folder. That will for example have surprising interactions with the dependency check in the start scripts. If you accept an update there, it will snapshot and thereby automatically sync the remind main renv and all run renvs. In general, what is the use case for having an up-to-date main renv (which will be used to run e.g. start.R), but stable run renvs?

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
> But that means you update the R package in the remind main renv, right?

No. 
> version differences of packages in the renv lock file and my user library
   
Consider this example:
- the current state
  ```
  ~/PIK/Remind (develop)$ Rscript -e 'renv::status()'
  The following package(s) are out of sync:
      Package   Lockfile Version   Library Version
     mrwater             1.3.28            1.3.30
   Use `renv::snapshot()` to save the state of your library to the lockfile.
  Use `renv::restore()` to restore your library from the lockfile.
  ```
- circumventing the local `.Rprofile` and thus `renv`, I update a package.  Or all packages.  Somewhere else.  Like in RStudio.
  ```
  ~/PIK/Remind (develop)$ R_PROFILE_USER=~/.Rprofile Rscript -e 'install.packages("edgeTransport")'
  Installing package into ‘/home/pehl/R/x86_64-pc-linux-gnu-library/4.0’
  (as ‘lib’ is unspecified)
  trying URL 'https://rse.pik-potsdam.de/r/packages/src/contrib/edgeTransport_0.22.1.tar.gz'
  Content type 'application/x-gzip' length 296109 bytes (289 KB)
  ==================================================
  downloaded 289 KB
  
  Warning in untar2(tarfile, files, list, exdir, restore_times) :
    using pax extended headers
  * installing *source* package ‘edgeTransport’ ...
  ** using staged installation
  ** R
  ** inst
  ** byte-compile and prepare package for lazy loading
  ** help
  *** installing help indices
  ** building package indices
  ** installing vignettes
  ** testing if installed package can be loaded from temporary location
  ** testing if installed package can be loaded from final location
  ** testing if installed package keeps a record of temporary installation path
  * DONE (edgeTransport)
  
  The downloaded source packages are in
  	‘/tmp/Rtmp0zlEB5/downloaded_packages’
  ```
  Note that they went into my user library `/home/pehl/R/x86_64-pc-linux-gnu-library/4.0`, totally independent of this REMIND directory and its `renv`.
- and now:
  ```
  ~/PIK/Remind (develop)$ Rscript -e 'renv::status()'The following package(s) are out of sync:

         Package   Lockfile Version   Library Version
   edgeTransport             0.21.1            0.22.1
         mrwater             1.3.28            1.3.30
     
  Use `renv::snapshot()` to save the state of your library to the lockfile.
  Use `renv::restore()` to restore your library from the lockfile.
  ```
  the `renv` is "more" out of sync.

But I don't care about those updates.  I want a stable package environment unitl I decided otherwise.  And I don't want to be pestered about this every 15 minutes during development.

mikapfl:
Okay, that looks like a bug to me. Why is your personal library considered in the renv? @pfuehrlich-pik is that intended?

At least we would have to be careful about `renv::snapshot` calls then, e.g. in the start scripts (currently a result of checkDeps) because they will transfer stuff from the personal library into the renv.

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
> Okay, that looks like a bug to me. Why is your personal library considered in the renv?

Confluence of two undesirable behaviours:
- I configured my user library in `~/.Rprofile`.
- The REMIND `./.Rprofile` sources `~/.Rprofile`, which can do all kinds of stuff breaking `renv`.

I moved on to `~/.Renviron`.  Maybe we should take stock if and what people need from their `~/.Rprofile` and maybe not source it.

Anyways, the merge request is moot.

Thanks @mikapfl for remote-sanity-checking my R setup ;)

pfuehrlich-pik:
I also think sourcing `~/.Rprofile` in the local `.Rprofile` is not a good idea, so I'd welcome if we 'd change that.

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
> I also think sourcing `~/.Rprofile` in the local `.Rprofile` is not a good idea, so I'd welcome if we 'd change that.

Drop it and see if it breaks anything?  Or ask people beforehand?

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
All I [find in the MO Wiki](https://redmine.pik-potsdam.de/projects/mo/search?utf8=%E2%9C%93&scope=subprojects&wiki_pages=1&q=Rprofile) relates to the package repository (which is included), AppResults (does anybody start that from a REMIND directory?), and snapshots (obsolete).

PR #1046, opened 2022-11-03
Do not source the user .Rprofile, update docs, add tutorial 02a_RunningREMINDLocally.md

mikapfl:
# Purpose of this PR

So far, in REMIND's Rprofile, the user Rprofile (`~/.Rprofile`) was sourced. Also, a comment which predates the transition to git states that this is "very important". However, in my view it is harmful because it breaks reproducibility. If your REMIND only runs with some info from a user Rprofile, it will never run again. Let's remove it.

Does anybody know why it might be "very important"?

## Type of change

- [x] Bug fix 

## Checklist:


0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q (approved):
> Also, a comment which predates the transition to git states that this is "very important".

Only predates the transition to Github, actually: https://gitlab.pik-potsdam.de/REMIND/REMIND/-/commit/921146c8ce96f7ab3ca7cd09068ee3be9efe1b36

I would venture the educated guess that it was "very important" to include the RSE/MO package repository https://redmine.pik-potsdam.de/projects/mo/wiki/Configuration_of_repositories_-_Add_the_PIK_repository/25

The point about
> very important to load user specific settings

is detrimental to reproducibility, of course.

Kill it, kill it with fire!

pfuehrlich-pik:
After thinking about this some more, there are some problems with not sourcing the home Rprofile. One reason is to not take away global configuration like `options(warnPartialMatchAttr = TRUE)`, the repos option, or custom aliases whenever working in renv. The repos option is of course the most important, and that is not a problem on the cluster, because we load that with the global `.Rprofile`, but locally almost all users will have the repos set in their home `.Rprofile`. Also, this change will be confusing to some users, just because we had the current behavior of sourcing the home Rprofile in remind for such a long time.

pfuehrlich-pik:
I discussed this with Jan recently, did I forget one of your arguments @tscheypidi ?

mikapfl:
> After thinking about this some more, there are some problems with not sourcing the home Rprofile.
Yes, there are tradeoffs.

> One reason is to not take away global configuration like options(warnPartialMatchAttr = TRUE), the repos option, or custom aliases whenever working in renv. The repos option is of course the most important, and that is not a problem on the cluster, because we load that with the global .Rprofile, but locally almost all users will have the repos set in their home .Rprofile.

I'll take those in turn:
* not taking things like `options(warnPartialMatchAttr = TRUE)` is a feature in my book. If it is good for everybody running the model, put it into the model's .Rprofile. If it has tradeoffs, make it an option.
* not taking the repos option: We should somehow make this transparent as well. Ideas: Make it an option which is set in the configuration. Or add an `.Rprofile.user` in the model folder, which is git-ignored, but which is copied to the output folder to make sure the run is reproducible.

> Also, this change will be confusing to some users, just because we had the current behavior of sourcing the home Rprofile in remind for such a long time.

Yes, but the current state of affairs is confusing and dangerous to some users as well, and breaks reproducibility. I think we can't accept that, at least not in production runs.

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
> - not taking things like `options(warnPartialMatchAttr = TRUE)` is a feature in my book. If it is good for everybody running the model, put it into the model's .Rprofile. If it has tradeoffs, make it an option.

I second that.  REMIND behaviour _should not_ depend on whether this option happens to be set or not.  Even if we were to keep sourcing  `~/.Rprofile`, we should specifically exclude/overwrite stuff like this.

> The repos option is of course the most important, and that is not a problem on the cluster, because we load that with the global `.Rprofile`

What global `.Rprofile`?
```
$ ls $R_HOME/etc/Rprofile.site
ls: cannot access '/p/system/packages/R/4.1.2/gcc-mkl/lib64/R/etc/Rprofile.site': No such file or directory
```
> but locally almost all users will have the repos set in their home `.Rprofile`.

I'll venture two wagers:
1. No REMIND user sets any repository other than `options(repos = c(CRAN = "@CRAN@", pik = "https://rse.pik-potsdam.de/r/packages"))` in their `~/.Rprofile`.

> Also, this change will be confusing to some users, just because we had the current behavior of sourcing the home Rprofile in remind for such a long time.

2. No REMIND user runs R from within a REMIND directory interactively, and would therefore be in a position to even notice the difference.

mikapfl:
> What global `.Rprofile`?

If you have loaded the `piam` module, you get our lovely piam `.Rprofile`:
```bash
$ env|grep R_PROFILE
R_PROFILE=/p/projects/rd3mod/R/.Rprofile_1.24
```
it contains many things, unfortunately. Good things, but also scary things because the cluster is a scary place.

> No REMIND user sets any repository other than `options(repos = c(CRAN = "@CRAN@", pik = "https://rse.pik-potsdam.de/r/packages"))` in their `~/.Rprofile`.

For starters, many probably would also set `options(remind_repos)` if running REMIND on their laptop. Also, `gdxrrw::igdx("/opt/gams/39.1")` or something similar is usually needed.

> No REMIND user runs R from within a REMIND directory interactively, and would therefore be in a position to even notice the difference.

I have hoped so in the past, but running R from within a REMIND directory interactively is surprisingly popular. Some people are *sourcing* `start.R` to start REMIND.

pfuehrlich-pik:
> not taking the repos option: We should somehow make this transparent as well. Ideas: Make it an option which is set in the configuration. Or add an .Rprofile.user in the model folder, which is git-ignored, but which is copied to the output folder to make sure the run is reproducible.

Something like this could be the solution. The thing is, some smart user might just say `source(~/.Rprofile)` in their `.Rprofile.user` :see_no_evil: Maybe we should just have an option to enable `~/.Rprofile` which is disabled by default. And if the user sets it, we should copy `~/.Rprofile` into the run folder?

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
> If you have loaded the `piam` module, you get our lovely piam `.Rprofile`:

Yeay, another undocumented dependency!

> > No REMIND user sets any repository other than `options(repos = c(CRAN = "@CRAN@", pik = "https://rse.pik-potsdam.de/r/packages"))` in their `~/.Rprofile`.
> 
> For starters, many probably would also set options(remind_repos) if running REMIND on their laptop.

Yes, they do.  I do.  But that does not relate to the `repos` option. :point_up:  
But that's the one thing that would break: getting new input data locally.  I'd say use an environment variable for that.

> Also, `gdxrrw::igdx("/opt/gams/39.1")` or something similar is usually needed.

No.  Because https://github.com/pik-piam/gdx/blob/master/R/onLoad.R.

> I have hoped so in the past, but running R from within a REMIND directory interactively is surprisingly popular. Some people are _sourcing_ `start.R` to start REMIND.

The kind of nonsense up with which we shall not put.

I stand fast:
> Kill it, kill it with fire!

pfuehrlich-pik (approved):
All in all, I agree that reproducibility is probably the more important feature. There might be some problems, but I'm confident now that we can find good solutions when they come up. Thanks for the discussion guys!

.Rprofile
```diff
@@ -12,3 +12,58 @@ if (isTRUE(rownames(installed.packages(priority = "NA")) == "renv")) {
...
+invisible(lapply(REMIND_repos_vars,
+                 function(x) { assign(x, Sys.getenv(x), pos = 1) }))
+use_cluster_defaults <- TRUE
+
+# add local directories, if any
+if ('' != get(REMIND_repos_vars[1])) {
+  foo <- unlist(
+    strsplit(REMIND_repos_dirs,
+             ifelse('Windows' == getElement(Sys.info(), 'sysname'),
+                    ';', ':'),
```
pfuehrlich-pik, line 35:
```suggestion
             if ('Windows' == getElement(Sys.info(), 'sysname') ';' else ':',
```
> Further note that if(test) yes else no is much more efficient and often much preferable to ifelse(test, yes, no) whenever test is a simple true/false result, i.e., when length(test) == 1. 

https://rdrr.io/r/base/ifelse.html

.Rprofile
```diff
@@ -12,3 +12,58 @@ if (isTRUE(rownames(installed.packages(priority = "NA")) == "renv")) {
...
+if (  use_cluster_defaults
+      & all(file.exists(c('/p/projects/rd3mod/inputdata/output',
+                          '/p/projects/remind/inputdata/CESparametersAndGDX')))) {
+  options(remind_repos = list(
+    '/p/projects/rd3mod/inputdata/output' = NULL,
+    '/p/projects/remind/inputdata/CESparametersAndGDX' = NULL))
+}
+
+# clean up
+rm(list = c(REMIND_repos_vars, 'REMIND_repos_vars', 'use_cluster_defaults'))
```
pfuehrlich-pik, line 69: I'd strongly recommend wrapping everything in `base::local({...})`, no need to `rm` anywhere, and you cannot forget individual variables.

.Rprofile
```diff
@@ -12,3 +12,58 @@ if (isTRUE(rownames(installed.packages(priority = "NA")) == "renv")) {
...
+  attr(bar, 'names') <- foo
+  options(remind_repos = c(options('remind_repos')[[1]], bar))
+  rm(list = c('foo', 'bar'))
+  use_cluster_defaults <- FALSE
+}
+
+# default to cluster directories
+if (  use_cluster_defaults
+      & all(file.exists(c('/p/projects/rd3mod/inputdata/output',
+                          '/p/projects/remind/inputdata/CESparametersAndGDX')))) {
```
pfuehrlich-pik, line 62:
```suggestion
if (use_cluster_defaults
      && all(file.exists(c('/p/projects/rd3mod/inputdata/output',
                           '/p/projects/remind/inputdata/CESparametersAndGDX')))) {
```

.Rprofile
mikapfl: doesn't work:
```
Error: 50:65: unexpected string constant
49:     strsplit(remindReposDirs,
50:              if ('Windows' == getElement(Sys.info(), 'sysname') ';'
```

mikapfl:
Closes: #1058

mikapfl:
@LaviniaBaumstark do you have more input on this? Otherwise I'd like to go ahead with the merge once I am able to actually test everything (at the moment, I'm locked out of the cluster for trying too often with wrong settings).

.Rprofile
pfuehrlich-pik: There's a brace missing, it should be
```r
if ('Windows' == getElement(Sys.info(), 'sysname')) ';'
```

.Rprofile
```diff
@@ -13,8 +13,71 @@ if (isTRUE(rownames(installed.packages(priority = "NA")) == "renv")) {
   message("Finished installing R package dependencies.")
 }
 
-# source global .Rprofile (very important to load user specific settings)
-# DO NOT EDIT THIS LINE!
-if (file.exists("~/.Rprofile")) {
-  source("~/.Rprofile")
+base::local({
```
pfuehrlich-pik, line 16:
```suggestion
local({
```
Prefixing with `base` looks weird, I just wrote it with `base::` to emphasize that it is just available anyway :slightly_smiling_face:

.Rprofile
```diff
@@ -13,8 +13,71 @@ if (isTRUE(rownames(installed.packages(priority = "NA")) == "renv")) {
...
+
+# unless specified otherwise, use cluster defaults
+use_cluster_defaults <- TRUE
+
+# add local directories, if any
+if ("" != remindReposDirs) {
+  directories <- unlist(
+    strsplit(remindReposDirs,
+             ifelse("Windows" == getElement(Sys.info(), "sysname"),
+                    ";", ":"),
```
pfuehrlich-pik, line 52:
```suggestion
             if ("Windows" == getElement(Sys.info(), "sysname")) ";" else ":",
```

.Rprofile
```diff
@@ -13,8 +13,71 @@ if (isTRUE(rownames(installed.packages(priority = "NA")) == "renv")) {
...
+             fixed = TRUE))
+  directoriesList <- rep(list(NULL), length(directories))
+  names(directoriesList) <- directories
+
+  options(remind_repos = c(options("remind_repos")[[1]], directoriesList))
+  rm(directories, directoriesList)
```
pfuehrlich-pik, line 58:
```suggestion
```
`local` will do the cleanup

.Rprofile
```diff
@@ -13,8 +13,71 @@ if (isTRUE(rownames(installed.packages(priority = "NA")) == "renv")) {
...
+
+# add remote directories, if any remote directory and username and SSH key are set
+if ("" != remindReposSCP && "" != remindReposSCPUser && "" != remindReposSCPKey) {
+  SCPUrls <- unlist(strsplit(remindReposSCP, ";", fixed = TRUE))
+  config <- list(list(username = remindReposSCPUser, ssh_private_keyfile = remindReposSCPKey))
+  for (SCPUrl in SCPUrls) {
+    names(config) <- SCPUrl
+    options(remind_repos = c(options("remind_repos")[[1]], config))
+  }
+  rm(SCPUrls, SCPUrl, config)
```
pfuehrlich-pik, line 70:
```suggestion
```

pfuehrlich-pik (approved):
Looks good, just remove the unnecessary `rm` calls.

pfuehrlich-pik:
There is also a batch accept suggested changes feature somewhere :wink:

.Rprofile
mikapfl: doh, I thought it was simply not possible to have an expression there. That's what I get for not learning the formal R grammar properly…

LaviniaBaumstark (approved):
please, also explain it in the next RMEIND meeting

mikapfl:
yeh, already added it to the minutes. I plan to merge this only *after* the next meeting, so everybody has a heads-up what to do so REMIND continues working locally.

tutorials/06_Advanced_ChangeInputs.md
```diff
@@ -24,30 +23,57 @@ the name of the needed input data is constructed. It is checked whether those in
...
 
-Local Input Data
-==================
-If you like to get the input data on your local machine you need to copy your ssh key (`/home/[your_cluster_user_name]/.ssh/id_dsa`) to your computer and adjust your local **.Rprofile** by adding the following line:
-``` r
-options(remind_repos=list("scp://cluster.pik-potsdam.de/p/projects/rd3mod/inputdata/output"=list(username="[your_cluster_user_name]",ssh_private_keyfile="[path_to_your_local_copy_of_your_key]")))
+## Download Input Data Automatically
+
+If you like to automatically download the input data when running REMIND on your local machine, you need to configure this via environment variables.
+On Linux, add to your `~/.bashrc` file (found in your home directory):
```
0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q, line 29: Why not `~/.Renviron`?  That is not affected by the `.Rprofile` in the REMIND repository, the syntax is identical on all systems, … and historic performance of REMIND team members modifying Windows environment variables has been less then stellar.

tutorials/06_Advanced_ChangeInputs.md
```diff
@@ -24,30 +23,57 @@ the name of the needed input data is constructed. It is checked whether those in
 
 The prepared input data is a compressed tar archive file "`.tgz`", which can be opened with software such as [7-Zip](https://www.7-zip.org/), or in terminal by `tar` and `untar` commands. 
 
-Local Input Data
-==================
-If you like to get the input data on your local machine you need to copy your ssh key (`/home/[your_cluster_user_name]/.ssh/id_dsa`) to your computer and adjust your local **.Rprofile** by adding the following line:
-``` r
-options(remind_repos=list("scp://cluster.pik-potsdam.de/p/projects/rd3mod/inputdata/output"=list(username="[your_cluster_user_name]",ssh_private_keyfile="[path_to_your_local_copy_of_your_key]")))
+## Download Input Data Automatically
```
0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q, line 26: This only touches on REMIND input data behind `inputRevision`, but not `CESandGDXversion`.  Probably because this tutorial was copied from MAgPIE which does/did not use gdxes.  But REMIND needs both, and both need to correspond (see remindmodel/development_issues#146), which is the reason multiple directories/connections can be used.

tutorials/06_Advanced_ChangeInputs.md
mikapfl: `~/.Renviron` sounds much better, yes, will update the tutorial accordingly.

tutorials/06_Advanced_ChangeInputs.md
mikapfl: But that should be fixed with this PR, right? It tells people to do:
```
export REMIND_repos_scp="scp://cluster.pik-potsdam.de/p/projects/rd3mod/inputdata/output;scp://cluster.pik-potsdam.de/p/projects/remind/inputdata/CESparametersAndGDX"
```
which includes also the CESparametersAndGDX.

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q: But doesn't tell them why, or which data to download to local directories for use (further below).

tutorials/02a_RunningREMINDLocally.md
```diff
@@ -0,0 +1,37 @@
+# Running REMIND locally
+Mika Pflüger (mika.pflueger@pik-potsdam.de)
+
+*Note:* This tutorial is not yet finished, not all steps needed are described, yet.
+
+## Download Input Data Automatically
+
+If you have access to the PIK cluster and like to automatically download the input data when running REMIND on your local machine, you need to configure this via environment variables.
```
LaviniaBaumstark, line 8: maybe we should add here a comment that only PIK members have access to the PIK cluster. Otherwise, even more external people ask for access to the PIK cluster

tutorials/06_Advanced_ChangeInputs.md
```diff
@@ -1,68 +1,77 @@
...
-``` bash
+(default setting: `regionmapping <- "config/regionmappingH12.csv"`).
+Based on the regional resolution and the input data revision 
+```R
 cfg$inputRevision
 ```
-the name of the needed input data is constructed. It is checked whether those input data are already available. If not they are automatically downloaded from `/p/projects/rd3mod/inputdata/output/` and distributed.
-
-The prepared input data is a compressed tar archive file "`.tgz`", which can be opened with software such as [7-Zip](https://www.7-zip.org/), or in terminal by `tar` and `untar` commands. 
+the name of the needed input data is constructed. It is checked whether those input data are already available. If not they are automatically downloaded and distributed.
```
LaviniaBaumstark, line 34: ...are available for PIK members.

tutorials/06_Advanced_ChangeInputs.md
```diff
@@ -1,68 +1,77 @@
...
-options(remind_repos=list("scp://cluster.pik-potsdam.de/p/projects/rd3mod/inputdata/output"=list(username="[your_cluster_user_name]",ssh_private_keyfile="[path_to_your_local_copy_of_your_key]")))
-```
-If your key is password-protected you have to remove this by using an empty passphrase:  
-``` cmd 
-ssh-keygen -p
--> path to id_dsa
--> empty passphrase
-```
-It is recommended to first copy the existing key (e.g. id_dsa into id_dsa-local) and remove the passphrase of this copy and copy it to your machine.
+The prepared input data is a compressed tar archive file `.tgz` and can be found on the cluster at `/p/projects/rd3mod/inputdata/output`.
```
LaviniaBaumstark, line 36: PIK-cluster

tutorials/06_Advanced_ChangeInputs.md
```diff
@@ -1,68 +1,77 @@
...
+If you want to peek inside the archive to debug something or out of curiosity you can use the software [7-Zip](https://www.7-zip.org/), or the `tar` command in the terminal.
 
-If you plan to run REMIND not with the default regional resolution you have to take care that REMIND starts from a gdx with the correct regional resolution. Either you can use one from an older run with the corresponding regional resolution or you can construct a new gdx with the correct regional resolution from a gdx in a different regional resolution by using the function `gdx_rename` from the package (gdx) (e.g. `gdx_rename("input.gdx",set_name="all_regi",c(REF="RUS",CAZ="ROW",...,MEA="MEA",USA="USA"))`).
+### How to Update Input Data
 
+1. Check for the current input data revision number in this cluster folder: `/p/projects/rd3mod/inputdata/cache`. Alternatively, run the helper tool `lastrev` (`/p/projects/rd3mod/tools/lastrev`) to get a list of the last five revX.XXX*_remind.tgz items in the default madrat output directory.
```
LaviniaBaumstark, line 41: that is not correct any more, you need to check for the latest revision in the output folder (/p/projects/rd3mod/inputdata/output)

tutorials/06_Advanced_ChangeInputs.md
```diff
@@ -1,68 +1,77 @@
...
+2. Clone the [remind-preprocessing repo](https://github.com/remindmodel/pre-processing) to your tmp folder on the cluster and edit its `start.R` file by inserting the next revision number. Use at least 4 decimal places for development/testing. If an old revision number is used, the input data will not be recalculated. Input data for a new regional resolution will be recalculated based on the existing cache information.
```
LaviniaBaumstark, line 43: to be exact: based on the existing puc-file

tutorials/06_Advanced_ChangeInputs.md
```diff
@@ -1,68 +1,77 @@
...
 
-1. Check for the current input data revision number in this cluster folder: `/p/projects/rd3mod/inputdata/cache`. Alternatively, run the helper tool `lastRev` (`/p/projects/rd3mod/tools/lastrev`) to get a list of the last five revX.XXX*_remind.tgz items in the default moinput output directory.
-2. Clone the repo [`https://gitlab.pik-potsdam.de/REMIND/preprocessing-remind`](https://gitlab.pik-potsdam.de/REMIND/preprocessing-remind) to your tmp folder on the cluster and edit its `start.R` file by inserting the next revision number. Use at least 4 decimal places for development/testing. If an old revision number is used, the input data will not be recalculated. Input data for a new regional resolution will be recalculated based on the existing cache information.
+## Initial Parameter Values
 
-3. Start the script with
+The initial values for all optimization parameters are taken from a previous run and supplied via an input `gdx` file.
+In a bundle of runs, a previous run from the same bundle is usually configured via the `path_gdx` configuration parameter.
```
LaviniaBaumstark, line 58: we also use the input.gdx file in a bundle of runs. We only use information from former scenarios for time steps before startyear. I would delete this sentence here

tutorials/06_Advanced_ChangeInputs.md
LaviniaBaumstark: jep

LaviniaBaumstark (approved):

PR #1047, opened 2022-11-04
delete switches that are defined twice from main.gms

orichters:
## Purpose of this PR
cleanup.
- `cfg$converted_input_folder` seems to be some legacy, reposearch shows it is never used somewhere
- cm_CES_configuration, cm_dispatchSetyDown and cm_dispatchSeelDown were defined twice in main.gms, one occurence is deleted.

## Type of change

- [x] cleanup


LaviniaBaumstark (approved):
thanks

PR #1050, opened 2022-11-04
Makefile: add check-fix target to run codeCheck in interactive mode.

mikapfl:
# Purpose of this PR

Typing `Rscript -e "gms::codeCheck(strict = TRUE, interactive = TRUE)"` manually is a mouthful and hard to remember, so add a make target so you can simply type `make check-fix` before comitting and have forgotten `not_used` entries and similar checked automatically.

## Type of change

- [x] New feature 
- [x] Minor change

## Checklist:


orichters (approved):
great. I wouldn't have care if `check` would simply be in interactive mode, but that is fine, too! Thanks for adapting the tutorial as well.

PR #1053, opened 2022-11-07
use lucode2::showUpdates, Tutorial: exclude renv from update

pfuehrlich-pik:
# Purpose of this PR
To remove code duplication lucode2::showUpdates was added. The renv tutorial showed `renv::update()`, added `exclude = "renv"` there.

## Type of change

- [x] Refactoring

## Checklist:


scripts/start/submit.R
```diff
@@ -62,26 +62,15 @@ submit <- function(cfg, restart = FALSE, stopOnFolderCreateError = TRUE) {
...
-          outdatedPackages <- utils::old.packages(instPkgs = installed[installed[, "Package"] %in% pikPackages, ])
-          if (!is.null(outdatedPackages)) {
-            message("The following PIK packages can be updated:\n",
-                    paste("-", outdatedPackages[, "Package"], ":",
-                          outdatedPackages[, "Installed"], "->", outdatedPackages[, "ReposVer"],
-                          collapse = "\n"),
-                    "\nConsider updating with `Rscript scripts/utils/updateRenv.R`.")
-          }
+        } else if (!is.null(lucode2::showUpdates())) {
+          message("Consider updating with `Rscript scripts/utils/updateRenv.R`.")
```
pfuehrlich-pik, line 66: Should we tell users to use make now in these cases @mikapfl ?
```suggestion
          message("Consider updating with `make update-renv`.")
```

orichters (approved):

scripts/start/submit.R
mikapfl: For linux users, `make` will be easier. People running on Windows probably want `Rscript`. I'd say, leave it as it is.

mikapfl (changes requested):
The lucode2 version needs to be updated also here: https://github.com/remindmodel/remind/blob/64f0d719855e612cb7118028b18924d895dbc4af/start.R#L216

That's not nice, but no way around, since we use lucode2 to check the versions in the DESCRIPTION so there is a circular dependency if just adding it to the DESCRIPTION.

pfuehrlich-pik:
I don't think so, but maybe I'm missing something. We need lucode2 >= 0.34.0 so we have access to `lucode2::checkDeps`. Once we have that all further required lucode2 updates are enforced/suggested by checkDeps, no?

mikapfl (approved):

PR #1054, opened 2022-11-07
Clean-up of several bioenergy-related switches

merfort:
# Purpose of this PR

This PR cleans up some switches related to bioenergy. Except for the bugfix regarding `cm_bioenergymaxscen` (see below) the changes in this PR are mostly refactoring work and don't affect the scenario results.

**Main changes:**

- `cm_bioenergymaxscen` :
   - Flexibilization of the switch, exchanging the fixed scenarios (`0`,`1`,`2`,`3`,`4`) by a numeric value that represents the maximum amount of bionergy in EJ/yr that can be produced. The switch is converted from a parameter to a flag, changing the default option from `0` to `off`. This was changed accordingly in all scenario configs, where the switch is present.
   - Fix calculation of bound for purpose grown biomass. There was a bug that led to a too high bound on purpose grown biomass in early years. 
   - Clean-up of the bioenergy bound calculation. Historically, there was an exogenously given bound on bioenergy for years less or equal 2025. This very generous bound is not binding anyways, so it is removed. It involves removing `s30_max_pebiolc` and `p30_max200_path(tall)`, since they are not needed anymore.
   - Change of the description of `cm_bioenergymaxscen`. Before it was **wrongly** stating that the bound affects _pebiolc production excluding residues_, now it **correctly** states that the bound affects _pebiolc production including residues but excluding traditionally used biomass_.
   - Renaming to `cm_maxProdBiolc` in order to comply with naming conventions and to reflect the new behavior.
- `cm_tradecost_bio`:
   - Flexibilization of the switch, exchanging the fixed scenarios (`1`,`2`) by a numeric value that can be read in. Scenario `2`, which was the default option, is now converted to the numeric value `1`, keeping trade costs at the default level. Scenario `1`, which was the scenario for _low trade costs_, is now converted to the numeric value `0.5`, halving the trade costs. This was changed accordingly in all scenario configs, where the switch is present.
   - Renaming to `cm_tradecostBio` in order to comply with naming conventions.
- `cm_biolc_tech_phaseout`
   - Ban 2nd gen. bioenergy trade in the bioenergy technology phaseout scenario (`cm_biolc_tech_phaseout eq 1`). In that case, trade is anyways just happening in early years at very low levels. It may, however, lead to problems with bioenergy market clearance, so it is disabled completely.
   - Renaming to `cm_phaseoutBiolc` in order to comply with naming conventions.
- Refactoring

## Type of change

- [x] Bug fix 
- [x] Refactoring

## Checklist:

- [x] I have adjusted reporting where it was needed (NA)

Test runs are here: 
`/p/tmp/merfort/REMIND_cleanup/remind`

a) Changes in `cm_bioenergymaxscen` / `cm_maxProdBiolc`: 
  * The 100 EJ bound that was originally set via `cm_bioenergymaxscen = 1`  is now (via `cm_maxProdBiolc = 100`) with the new (corrected) implementation a bit lower (green vs. orange line).
  * The new switch now allows for even lower bounds such as `cm_maxProdBiolc = 75` or `cm_maxProdBiolc = 60`.
image

b) Forbidding biomass trade with `cm_biolc_tech_phaseout`/`cm_phaseoutBiolc` does what it should (but does not significantly impact the results):
image


robinhasse (approved):
I cannot comment much on bio energy-related topics but the faster phase out of traditional biomass looks good to me. Thanks Leon!

bs538:
On traditional biomass phaseout: Already now (without faster) phaseout, IND, SSA, and OAS have a big break with historical trends. Why would you make it even faster?

Note on the phase-out: This is hard-coded in two places, and the numbers are rather ad-hoc. If we are already touching it, can we find a better way of modelling it? @robinhasse do you see a possibility to link this more explicitly to EDGE-Buildings output?

robinhasse:
> Why would you make it even faster?
> do you see a possibility to link this more explicitly to EDGE-Buildings output?

We maybe could as EDGE-B reports traditional biomass. But we don't make use of this information as everything is summed to the final energy carrier solids. But the traditional biomass data is not very reliable in any case, right? I now at least that we use a simple split of solid biomass into traditional and modern by GDPpC with the IEA Balances that don't have this split. Don't know if the historic numers in REMIND have a better basis.

bs538:
I now looked at the code changes as well. Regardless of whether we opt for the "fast" setting as default or not, the SDP specific modifications (fastest possible phaseout) would need to stay in. I would argue the same for SSP1 and SSP5 modifications (Btw, I'm just preparing a fix to that, as SDP_XX were missing there).

merfort:
Well, the way it is implemented in this PR is that we now opt for the "fasted possible phaseout" in scenarios. So you want to keep a slower phaseout in non-SDP scenarios?

bs538:
Sorry, misread the code (can I blame parental sleep deprivation?). You're right that after your change there is zero new capacity in all scenarios, which is the fasted possible phaseout (except if one wants to modify technology lifetimes I think).

But yes, I would keep a differentiation between scenarios here - unless perhaps the exploration by @robinhasse shows that already the fastest possible phaseout is justified in SSP2.

dklein-pik (approved):
Thank you for this big effort!

merfort:
Ok, I decided to remove all the changes related to traditional biomass phaseout, since this needs further discussion and should be handled in a separate PR.

bs538:
sounds good, thanks!

dklein-pik (approved):
Looks good!

mikapfl:
This breaks the default config in REMIND:
> Reading config file config/scenario_config.csv 
> 
> Automated checks did not find counterparts in main.gms and default.cfg for these columns in scenario_config.csv:
>   cm_bioenergymaxscen, cm_tradecost_bio
> The start script might simply ignore them. Please check if these switches are not deprecated.
> This check was added Jan. 2022. If you find false positives, add them to knownColumnNames in start/scripts/readCheckScenarioConfig.R.
> 
> Column name cm_bioenergymaxscen in remind settings is outdated. Rename to cm_maxProdBiolc, see https://github.com/remindmodel/remind/pull/1054
> Column name cm_tradecost_bio in remind settings is outdated. Rename to cm_tradecostBio, see https://github.com/remindmodel/remind/pull/1054

Please fix this.

PR #1065, opened 2022-11-10
Automatically assign development version

mikapfl:
# Purpose of this PR

@Renato-Rodrigues suggested having development versions which are automatically assigned to distinguish the REMIND development version more easily. Here, I implemented this as a github action. The effect is that after a PR is merged in remindmodel/remind develop, a version is automatically built in the format `major.minor.patch.dev123`, e.g. `3.0.0.dev839`. The major/minor/patch are taken from the last official release, and the dev-version (in the example 839) is the number of commits since the last official release, so will always be strictly increasing, but is reset when we make a release.

The automatically generated version is then used to update the `CITATION.cff` file, which is the only place where we had a version number so far. What else would you envision, @Renato-Rodrigues ?

Note that using this method, the version number is only assigned when a pull request is merged, so in forks and feature branches the version will just show which development version the fork/branch is based on. To also assign version numbers for forks and feature branches, the version would need to be really verbose like `3.0.0.dev839+mikapfl/remind_auto-dev-version_5` or so, which is not terribly useful, I think. There, a commit hash is probably just easier because it exists already and can be viewed directly with `git show`, etc.

## Type of change

- [x] New feature 

## Checklist:


Renato-Rodrigues:
Thank you very much!!!

> major.minor.patch.dev123

Was the patch field ever used, or is it intended to be used?
Is there any reason why not use the third level (patch) to add this version number instead of having an extra field?
Is there any reason why not to simply use the number without the keyword `dev` in the field? I am not familiar with best practices on versioning, but it seams cleaner to me to have only the number there, and it is also slightly easier to parse the version information.

> The automatically generated version is then used to update the CITATION.cff file, which is the only place where we had a version number so far. What else would you envision, @Renato-Rodrigues ?

For now I would be more than happy with this being only included at the citation file, as it is a human and machine readable file. Once we discuss further uses for the version number we can see if there is need for any additional automatic updates.

mikapfl:
> Was the patch field ever used, or is it intended to be used?

I think people are discussing a 3.0.1 bugfix release at the moment.

> Is there any reason why not to simply use the number without the keyword `dev` in the field?

I think we want to make sure people understand that these are development releases, which are *not* validated etc.

Renato-Rodrigues:
> I think people are discussing a 3.0.1 bugfix release at the moment.

Maybe we should ask this in the next remind meeting just to make sure we really need this extra field.

> I think we want to make sure people understand that these are development releases, which are not validated etc.

I would say that as the versioning is after the minor releases field, it already implies this fact without the need for extra strings.  

___

I have no further comments in the PR. Thank you very much for implementing this!

I think maybe you could present this in the next remind meeting before it is merged to settle on these points and see if anybody else has any further comments.

mikapfl:
I think the three-part version number derives from [semantic versioning](https://semver.org/), which recently gained traction. So, most people will think that any version with three components like 3.0.1 will be an official release.

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
> I think the three-part version number derives from [semantic versioning](https://semver.org/), which recently gained traction. So, most people will think that any version with three components like 3.0.1 will be an official release.

Can't help but to point out that under the scheme
> Given a version number MAJOR.MINOR.PATCH, increment the:
> 
> MAJOR version when you make incompatible API changes
> MINOR version when you add functionality in a backwards compatible manner
> PATCH version when you make backwards compatible bug fixes

we would have only major releases because backwards compatibility does not apply to mathematical models, except for the rare cases where we simply extend the reporting by writing more information to the .gdx files.

Renato-Rodrigues:
In my opinion, it would make sense in our case to have:

MAJOR -> for substantial changes in the REMIND model, e.g. change on demand sector formulations or added support for additional regional disaggregation. Incremental to this version would always require a validation round just before the release.
MINOR  -> for in between releases that have meaningful added features and changes to the model, but not necessarily include a comprehensive validation process or that do not justify a major version release.
PATCH or in our case DEV -> the automatic incremental REMIND version added by this PR.

LaviniaBaumstark:
> > Is there any reason why not to simply use the number without the keyword `dev` in the field?
> 
> I think we want to make sure people understand that these are development releases, which are _not_ validated etc.

I totally agree that we need to keep `dev` to clearly identify our development versions which are no releases and not validated.

LaviniaBaumstark:
> I think maybe you could present this in the next remind meeting before it is merged to settle on these points and see if anybody else has any further comments.

Please, let's not discuss this at the REMIND meeting. We are too many people for such a very specific topic where the preference is more based on taste than arguments. Could we come to a conclusion here?

mikapfl:
I propose to use `X.Y.Z.devABCD` as already done in this PR because:
1. Historically, we already had a series of 2.1.0, 2.1.1, etc. releases which were validated, and keeping the version numbering scheme consistent is arguably more important than any advantage I can see of having another, maybe better, versioning scheme.
2. Having `dev` in the last component of a development version makes it visible this is not a "traditional" version.

Unless there is strong opposition, I think we can go ahead as-is with this PR adding the development versions. If there is appetite to change the versioning scheme of official remind releases, I think that is an orthogonal problem to be discussed elsewhere.

Renato-Rodrigues:
I would agree as long as we include in the same PR a clear definition of the reason behind each version component, and make it clear when we should/will update each of them.
 
I provided a suggestion in my previous message on how to define the REMIND version using only three components (e.g. 3.1.dev556).

As long as we can agree in a valid proposal for using the four items option instead, whith each component having an unique and valid purpose in the REMIND versioning, I would see no problems on using the extra term if such versioning rules description is also included to the PR.

LaviniaBaumstark:
I think, for four items it would be `MAJOR.MINOR.PATCH.devABCD` and I think, we can merge it

LaviniaBaumstark (approved):

Renato-Rodrigues:
> I think, for four items it would be `MAJOR.MINOR.PATCH.devABCD` and I think, we can merge it

As I and @0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q already mentioned before, we don't see in any way how to define any commit to a GAMs model as not being a patch per definition. Could you explain better what you have in mind that differentiates this level of the version then from any incremental level of the dev part?

LaviniaBaumstark:
`MAJOR.MINOR,PATCH` would be reserved for `master` versions, while `devABC` would count for the `develop` branch. This enables ordered version across two branches. `PATCH` would be used if we have a bug-fix (not for a compilation error, but e.g. a wrong number in an important variable) which we want to have in the `master` but without all other developments from the meantime in the `develop` branch.

Renato-Rodrigues:
So you want to maintain parallel forks of remind with only selected changes? 
As far as I understand, this is what the patch description you just made implies.
I thought the policy was to be against polluting our main git hub with such cases.

mikapfl:
Hi,

the discussion about the meaning of version numbers is off-topic here, since the PR does not change the version numbering scheme for official releases. I moved the discussion about numbering official releases to #1138 .

Development releases will be another component after the official releases, so if the discussion in #1138 comes to a conclusion, I will adjust the development version numbering accordingly. But there is no reason not to have development versions using the current poorly-defined three-component scheme for official versions now.

Mika

mikapfl:
Implementation as-is doesn't work because the github actions bot can't push to `develop` directly because of branch protection. I haven't found a simple solution, so I revert for now and will look into it.

Renato-Rodrigues:
I definitely don't agree with the code change proponent deciding to merge a PR without having the ok from reviewers that requested for changes and/or requested for clarification points before the PR is merged.
This completely defeats the purposes of peer review. 
It is like submitting a paper and not accepting unilaterally some reviewer comments and having the paper published anyway because the author said so. It makes no sense at all. 

I also don't agree that the version format discussion is off-topic here.
If this PR were not reverted we would have a versioning system in place that could potentially conflict with a future format just because the merge is being rushed. I am the one that asked for the feature in the first place, and I really appreciate the help on implementing it. But, if the person that asked for the feature, and that is actually properly reviewing the topic, has no saying on this, what is the point anyway to have peer review in place?

PR #1069, opened 2022-11-11
fix debugging tutorial, allow not to select a scenario config in --interactive mode of start.R

orichters:
## Purpose of this PR

- fix debugging tutorial, claiming that you should rerun the scenario to speed-up debugging.
- refactor `start.R` so that not selecting a config file in `--interactive` mode starts a default run instead of failing

## Type of change

- [x] fix

## Checklist:

### DIFF
--- a/core/loop.gms
+++ b/core/loop.gms
@@ -160,13 +160,13 @@ if (o_modelstat le 2,
   execute_unload 'fulldata';
   !! retain gdxes of intermediate iterations by copying them using shell
   !! commands
-  if (c_keep_iteration_gdxes ge 1,
+  if (c_keep_iteration_gdxes eq 1,
     put_utility logfile, "shell" / 
       "cp fulldata.gdx fulldata_" iteration.val:0:0 ".gdx";
   );
 else
   execute_unload 'non_optimal';
-  if (c_keep_iteration_gdxes ge 1,
+  if (c_keep_iteration_gdxes eq 1,
     put_utility logfile, "shell" / 
       "cp non_optimal.gdx non_optimal_" iteration.val:0:0 ".gdx";
   );
--- a/core/loop.gms
+++ b/core/loop.gms
@@ -160,13 +160,13 @@ if (o_modelstat le 2,
   execute_unload 'fulldata';
   !! retain gdxes of intermediate iterations by copying them using shell
   !! commands
-  if (c_keep_iteration_gdxes eq 1,
+  if (c_keep_iteration_gdxes ge 1,
     put_utility logfile, "shell" / 
       "cp fulldata.gdx fulldata_" iteration.val:0:0 ".gdx";
   );
 else
   execute_unload 'non_optimal';
-  if (c_keep_iteration_gdxes eq 1,
+  if (c_keep_iteration_gdxes ge 1,
     put_utility logfile, "shell" / 
       "cp non_optimal.gdx non_optimal_" iteration.val:0:0 ".gdx";
   );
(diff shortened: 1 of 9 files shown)


0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
> - if `c_keep_iteration_gdxes = 2` is set, the `non_optimal_*.gdx` and `fulldata_*.gdx` files are written after each iteration as in `c_keep_iteration_gdxes = 1` mode, but if everything is fine, modelstat = 2 at the end and no other failure, they are deleted.
> 
> - helps debugging if something goes wrong but doesn't annoy you with numerous files if everything went fine at the end.

Honestly, I'm opposed to idea of dumping one to three GB of data on the cluster "just in case".  Either you debug a run and could use that information, then turn it on, or you don't, then don't.

> - I have some suspicion this could potentially interact with calibration runs where the same filenames are used for the calibration iterations (does `c_keep_iteration_gdxes` somehow interfere with calibration runs?)

Hasn't come up yet.

orichters:
I see. My starting point was the [debugging tutorial](https://github.com/orichters/remind/blob/develop/tutorials/10_DebuggingREMIND.md):
> If the error was not in the first iteration, a faster and more convenient way might be to start the run again in parallel mode with `c_keep_iteration_gdxes = 1` specified in `main.gms` or the `scenario_config_XYZ.csv`. And then start a debug run that uses the `fulldata_07.gdx` (or the first iteration in which this occurs) of this first run as `input.gdx` specified in `path_gdx` of `scenario_config_XYZ.csv`. That way, you should see where the infeasibility is in the first iteration of the debug run already and not have to wait until the 7th iteration.

So I seem to have to rerun the run, which also puts some stress on cluster resources. But maybe it is simply wrong what is in the debugging tutorial: The `fulldata.gdx` that exists in the output folder should actually be the required file, right? Then maybe we just have to adapt the tutorial?

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
> I see. My starting point was the [debugging tutorial](https://github.com/orichters/remind/blob/develop/tutorials/10_DebuggingREMIND.md):
> 
> > If the error was not in the first iteration, a faster and more convenient way might be to start the run again in parallel mode with `c_keep_iteration_gdxes = 1` specified in `main.gms` or the `scenario_config_XYZ.csv`. And then start a debug run that uses the `fulldata_07.gdx` (or the first iteration in which this occurs) of this first run as `input.gdx` specified in `path_gdx` of `scenario_config_XYZ.csv`. That way, you should see where the infeasibility is in the first iteration of the debug run already and not have to wait until the 7th iteration.

The tutorial is wrong.  If the error is not in the first Nash iterations, the `testOneRegi` run will not reproduce it at all in general, since the results of the different iterations will differ. This is not a matter of convenience or speed, but simply the only way to get the "input" .gdx (or as close as we are like to get) to the Nash iteration the produces an error.

> So I seem to have to rerun the run, which also puts some stress on cluster resources. 

If you are debugging a run, you will have to rerun it, yes.

orichters:
> If you are debugging a run, you will have to rerun it, yes.

That is not what I mean. The tutorial suggests (based probably on something somebody sent me) that you should rerun the run in nash mode with `c_keep_iteration_gdxes = 1` to get a GDX to start the debug run from (it is not talking about testOneRegi here). But if my runs is infes in iteration 8 (so produces a non_optimal.gdx), then I have the fulldata.gdx which represents the results after iteration 7, so I can start debugging from there, right?

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
Probably.

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q (approved):

PR #1072, opened 2022-11-14
showUpdates in start.R

pfuehrlich-pik:
# Purpose of this PR
The info which piam packages could be updated comes unnecessarily late. With this PR the update check is already done in start.R. The update check is additionally also done in the `submit` function, because there might be cases where submit is called directly, without sourcing/Rscripting `start.R` first (if this is not actually true we could just remove the check in submit).

## Type of change

- [x] Refactoring

## Checklist:


orichters (approved):
definitely better and I think the right place. Still, it is certainly easy to miss it, before R asks what slurmConfig you like. But I don't know what to do about it? `Sys.sleep(1)`? This at leasts add a small stop, so if you actually look at the log, you can see it.

PR #1076, opened 2022-11-14
output.R: Remove outdated path setting

mikapfl:
# Purpose of this PR

There was code in output.R which is most certainly outdated (it tried to access `/iplex` on the cluster, should be gone since at least 2015).

I didn't test anything, but `latexpath` and `pythonpath` are not used in the whole remind repo, so I think it is safe to not set them. But how would I smoke test `output.R`?

## Type of change

- [x] Refactoring

## Checklist:

### DIFF
--- a/output.R
+++ b/output.R
@@ -46,14 +46,6 @@ if (!exists("source_include")) {
   flags <- NULL
 }
 
-# Setting relevant paths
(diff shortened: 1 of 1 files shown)


orichters (approved):
yeah, looks very outdated :)

PR #1090, opened 2022-11-16
include pm_BAU_reg_emi_wo_LU_bunkers.cs4r only if it exists, to enable gamscompile testing

### LINKED ISSUE #1089: "All automated model tests compile successfully" unsatisfyable
https://github.com/remindmodel/remind/blob/1380bc1b6e8e29a1f736bbb802bb1e29f0e0c022/pull_request_template.md?plain=1#L23

Currently always fails for NDC runs.

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
- in NDC runs, pm_BAU_reg_emi_wo_LU_bunkers.cs4r is written by ./scripts/input/prepare_NDC.R and generally not available
- make the inclusion conditional on existance allows for compiling, while not interfering with gms::singleGAMSfile()
- close #1089

- [x] Bug fix 

## Checklist:

- [ ] Changes are commented, particularly in hard-to-understand areas

### DIFF
--- a/modules/45_carbonprice/NDC/datainput.gms
+++ b/modules/45_carbonprice/NDC/datainput.gms
@@ -74,7 +74,9 @@ display p45_histShare;
 Parameter p45_BAU_reg_emi_wo_LU_bunkers(ttot,all_regi) "regional GHG emissions (without LU and bunkers) in BAU scenario"
   /
 $ondelim
+$ifthen exist "./modules/45_carbonprice/NDC/input/pm_BAU_reg_emi_wo_LU_bunkers.cs4r"
 $include "./modules/45_carbonprice/NDC/input/pm_BAU_reg_emi_wo_LU_bunkers.cs4r"
+$endif
 $offdelim
   /             ;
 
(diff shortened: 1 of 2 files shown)


orichters (approved):
seems like a plausible solution for the test problem. I guess if you actually run the code and because of whatever reason the file is missing although it should be there, it would fail later with a "no value assigned" error, right? Then, I think this is good solution.

mikapfl (approved):

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
> I guess if you actually run the code and because of whatever reason the file is missing although it should be there, it would fail later with a "no value assigned" error, right?

No idea how that would happen, but it would fail in `gms::singleGAMSfile()` with
```
Creating full.gms
Error in .insertIncludeFile(code, i, p) : 
  Include file ./modules/46_carbonpriceRegi/NDC/input/pm_BAU_reg_emi_wo_LU_bunkers.cs4r could not be found!
```

PR #1093, opened 2022-11-17
showUpdates & checkDeps from piamenv

pfuehrlich-pik:
# Purpose of this PR
showUpdates and checkDeps are moved from lucode2 to piamenv, so this PR simply gets both functions from piamenv instead of lucode2. See also: https://github.com/pik-piam/lucode2/pull/137

## Type of change

- [x] Refactoring

## Checklist:

### DIFF
--- a/start.R
+++ b/start.R
@@ -219,11 +219,11 @@ if (any(c("--testOneRegi", "--debug", "--quick") %in% flags) & "--restart" %in% 
 }
 
 # Check if dependencies for a model run are fulfilled
-if (packageVersion("lucode2") >= "0.34.0") {
+if (packageVersion("piamenv") >= "0.2.0") {
   piamenv::checkDeps(action = "ask")
 } else {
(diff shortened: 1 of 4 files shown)


start.R
```diff
@@ -220,14 +220,14 @@ if (any(c("--testOneRegi", "--debug", "--quick") %in% flags) & "--restart" %in%
 
 # Check if dependencies for a model run are fulfilled
 if (packageVersion("lucode2") >= "0.34.0") {
-  lucode2::checkDeps(action = "ask")
+  piamenv::checkDeps(action = "ask")
```
mikapfl, line 223: here, you need to check `piamenv` instead of `lucode2`, right? Or does `lucode2 >= 0.34.0` already imply `piamenv`?

mikapfl (changes requested):

start.R
pfuehrlich-pik: Oh good point, will do that

mikapfl (approved):

PR #1096, opened 2022-11-21
Code cleaning, simplify search for old runs

dklein-pik:
# Purpose of this PR

Change the way old runs are searched for:
  - look for old runs (REMIND and MAgPIE) even if no scenario name is provided in "oldrun" in scenario_config_coupled (using the name of the current scenario)
User settings overwrite search results
  - if provided always use "path_gdx" from scenaio_config or scenario_config_coupled instead of any automatically found gdx

## Type of change

- [x] New feature 

## Checklist:

- [ ] All automated model tests compile successfully (`Rscript start.R -g config/scenario_config_AMT.csv`)
- [ ] The model runs successfully (`Rscript start.R -q`)


orichters:
Thanks, David. It all looks very good, but can you tell me what you tested? Because after the introduction of `--parallel`, we noticed many special situations where the script failed, although in normal operation mode everything looked good.

start_bundle_coupled.R
```diff
@@ -260,61 +267,65 @@ for(scen in common){
...
   # Check for existing REMIND and MAgPIE runs and whether iteration can be continued from those (at least one REMIND iteration has to exist!)
-  path_report_found <- NULL
-  start_magpie <- FALSE
+  # Look whether there is already a fulldata.gdx from a former REMIND run (check for old name if provided)
   iter_rem <- 0
-  needle <- scen
-  suche <- paste0(path_remind, "output/", prefix_runname, needle, "-rem-*/REMIND_generic_", prefix_runname, needle, "-rem-*.mif")
-  possibleRemindReport <- mixedsort(Sys.glob(suche))[1]
-  already_rem <- sub(paste0("REMIND_generic_", prefix_runname, needle, "-rem-.*\\.mif"), "fulldata.gdx", possibleRemindReport)
+  suche <- paste0(path_remind, "output/", prefix_runname, scen, "-rem-*/fulldata.gdx")
```
orichters, line 273: funny how the German word `suche` survives all refactoring :)

start_bundle_coupled.R
```diff
@@ -260,61 +267,65 @@ for(scen in common){
...
   } else {
     path_report_found <- normalizePath(already_mag)
     message("Found MAgPIE report here: ", path_report_found)
   }
+  
   # decide whether to continue with REMIND or MAgPIE
+  start_magpie <- FALSE
   if (iter_rem == iter_mag + 1 & iter_rem < max_iterations) {
     # if only remind has finished an iteration -> start with magpie in this iteration using a REMIND report
     start_iter_first  <- iter_rem
```
orichters, line 331: I'm not sure whether `already_rem` is correct here in line 332. In the way you reformulated it, this may not even be defined, right? Or it may point to a run that isn't actually used because you specified a `…/fulldata.gdx` in `path_gdx`. I guess it must be `settings_remind[scen, "path_gdx"]`

start_bundle_coupled.R
```diff
@@ -260,61 +267,65 @@ for(scen in common){
...
+      lookfor <- if (is.na(scenarios_coupled[scen, "oldrun"])) scen else scenarios_coupled[scen, "oldrun"]
+      suche <- paste0(path_remind_oldruns, prefix_oldruns, lookfor, "-rem-*/fulldata.gdx")
       already_rem <- mixedsort(Sys.glob(suche))[1]
     }
   }
-  if (is.na(already_rem)) {
-    message("Nothing found for ", suche, ", starting with ", runname, "-rem-1.")
-  } else {
+  
+  if (.isFileAndAvailable(settings_remind[scen, "path_gdx"], "/fulldata.gdx")) {
```
orichters, line 289: you could put that right at the beginning, right? Only if this is not the case, try to find `already_rem`. Which also avoids cluttering the log with "haven't found anything" messages.

start_bundle_coupled.R
```diff
@@ -260,61 +267,65 @@ for(scen in common){
...
   if (! is.na(already_mag)) {
     iter_mag <- as.integer(sub(".*mag-(\\d.*)/.*","\\1",already_mag))
-  } else if (! is.na(scenarios_coupled[scen, "oldrun"])) {
-    message("Nothing found for ", suche, ", continue with oldrun")
-    if (isTRUE(stringr::str_sub(scenarios_coupled[scen, "oldrun"], -14, -1) == "/report.mif") &&
-        file.exists(scenarios_coupled[scen, "oldrun"])) {
-          already_mag <- c(scenarios_coupled[scen, "oldrun"])
+  } else {
+    message("Nothing found for ", suche, ", continue with path_report")
+    if (.isFileAndAvailable(scenarios_coupled[scen, "path_report"], "/report.mif")) {
```
orichters, line 309: So you prioritize older magpie runs over `path_report`, but prioritize `fulldata.gdx` specified in `path_gdx` over older REMIND runs? We should document that somewhere.

start_bundle_coupled.R
```diff
@@ -435,14 +407,15 @@ for(scen in common){
         path_report <- path_report_found
       }
     } else {
+      # start_coupled.R uses the name of this run to create the name of the MAgPIE report for the subsequent runs (next iteration)
```
orichters, line 410: As I said, I would also be fine replacing it with the filename it will get later, as it is already known. But this is fine like this as well.

dklein-pik:
> Thanks, David. It all looks very good, but can you tell me what you tested? Because after the introduction of `--parallel`, we noticed many special situations where the script failed, although in normal operation mode everything looked good.

I have run `Rscript start_bundle_coupled.R --test" with several different settings in `scenario_config_coupled.csv` especially for `oldrun`, `path_report`, and `path_gdx` and with different situations regarding former runs, meaning: there were runs to be continued in the same output folder, or there were no runs to be continued but old runs the gdx / report should be taken from, with and without ovewriting the automatic search with `path_gdx`.

start_bundle_coupled.R
dklein-pik: Interesting. If I am not mistaken the question whether it is defined or not has not been touched by the changes. If `already_rem` is not defined (no former REMIND run found), `iter_rem` will be zero and this line will not be executed.

orichters: That is true. The problem is instead: This code will be executed only if it doesn't start the first REMIND, but say `rem-2`. So `already_rem` is defined (as it was finding old paths), but you now don't like to use this one, but the one specified in the config file… And so it should catch a wrong one, doesn't it?

start_bundle_coupled.R
dklein-pik: Good question and welcome to applied philosophy of coupling. I also considered this, but then had the following case in mind, where you would still need the search for old runs to yield an iteration number that can be continued: a REMIND iteration went infeasible and you want to repeat it and continue the iteration. One solution could be to provide a gdx from a different REMIND run than the direct iteration-predecessor but take the MAgPIE report from the regular preceeding MAgPIE run and continue the iteration counting. Then you wouldn't want to skip the seach for former runs. Is this described in an understandable way?

orichters: Yes, I got it. The number of use cases is simply annoyingly big :)

dklein-pik: Oh yeah!

start_bundle_coupled.R
dklein-pik: Okay, I changed this now: `path_gdx` *and* `path_report` overwrite whatever was found.

tutorials/04_RunningREMINDandMAgPIE.md
```diff
@@ -136,11 +136,13 @@ Other, optional columns allow you to make a run start only after another has fin
    - `path_gdx`, `path_gdx_ref`, `path_gdx_refpolicycost`, `path_gdx_carbonprice`, `path_gdx_bau`, : Override these same settings in REMIND's `scenario_config`, see [`03_RunningBundleOfRuns`](./03_RunningBundleOfRuns.md) for a detailed explanation.
       - You can set these switches either to the full path of a `fulldata.gdx` file or simply to the name of another scenario in the file (without the "C_"!). So if you want a certain scenario (say `NDC`) to use as starting point the results of a `Base` scenario, you can simply set `path_gdx` to `Base` and it will automatically locate the appropriate `fulldata.gdx` in `Base`, for example `path_remind/C_Base-rem-5/fulldata.gdx` if 5 iterations where requested and you run in "serial" mode.
```
orichters, line 137: great you updated the tutorial. may you also delete the reference to the `serial` mode here (basically, the last half sentence can be deleted)

orichters (approved):
Thanks, I think this is more consistent now.

orichters (approved):

dklein-pik:
> It may be worth starting actual test runs with [this method](https://github.com/pik-piam/discussions/discussions/28). Takes like an hour (`max_iterations <- 2` may suffice), but makes sure REMIND and MAgPIE actually find the right reports.

Thanks for the hint. Successfully tested a bunch of coupled runs using this method.

PR #1097, opened 2022-11-22
use .Platform$path.sep in .Rprofile instead of if-else statement

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
Toady I found out …


mikapfl (approved):
Nice, thanks!

PR #1098, opened 2022-11-22
add lock to --gamscompile, cleanup

orichters:
## Purpose of this PR

- Running `./start.R --gamscompile` now locks the model so that no run tinkers around with the files in the meantime as suggested by @mikapfl
- rename `verboseGamsCompile` to `interactive` because it is not only verbose, but also interactive :)
- better help message
- `prepare_and_run.R` isn't angry anymore that it cannot copy the magicc files if an existing run is started with `--reprepare`. @mikapfl: You suggested `-n`, but is it a problem that is [may not be compatible to all platforms](https://www.unix.com/unix-for-beginners-questions-and-answers/279031-cp-n-command.html)?
- in `start_bundle_config.R`, kick out all the `variable=whatever` stuff and `test` before searching for config files in the command line arguments

## Type of change

- [x] Bug fix 
- [x] no impact on scenarios

## Checklist:

### DIFF
--- a/start.R
+++ b/start.R
@@ -22,7 +22,10 @@ helpText <- "
 #'
 #'    --help, -h:        show this help text and exit
 #'    --debug, -d:       start a debug run with cm_nash_mode = debug
-#'    --gamscompile, -g: compile gms of all selected runs
+#'    --gamscompile, -g: compile gms of all selected runs. Combined with
+#'                       --interactive, it stops in case of compilation errors,
+#'                       allowing the user to fix them and rerun gamscompile;
+#'                       combined with --restart, existing runs can be checked.
 #'    --interactive, -i: interactively select config file and run(s) to be
 #'                       started
 #'    --quick, -q:       starting one fast REMIND run with one region, one
@@ -182,7 +185,7 @@ if (length(argv) > 0) {
 
 if ("--gamscompile" %in% flags) {
   dir.create(file.path("output", "gamscompile"), recursive = TRUE, showWarnings = FALSE)
-  runGamsCompile <- function(modelFile, cfg, verboseGamsCompile) {
+  runGamsCompile <- function(modelFile, cfg, interactive = TRUE) {
     tmpModelFile <- file.path("output", "gamscompile", paste0("main_", cfg$title, ".gms"))
     file.copy(modelFile, tmpModelFile, overwrite = TRUE)
     manipulateConfig(tmpModelFile, cfg$gms)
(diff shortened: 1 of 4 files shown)


start.R
```diff
@@ -22,7 +22,9 @@ helpText <- "
 #'
 #'    --help, -h:        show this help text and exit
 #'    --debug, -d:       start a debug run with cm_nash_mode = debug
-#'    --gamscompile, -g: compile gms of all selected runs
+#'    --gamscompile, -g: compile gms of all selected runs. Combined with
+#'                       --interactive, it given the user time to fix errors;
```
0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q, line 26: > it given the user time to fix errors

Say what?

start_bundle_coupled.R
```diff
@@ -107,7 +107,7 @@ if ("--parallel" %in% flags) {
 }
 
 # load arguments from command line
-argv <- argv[! grepl("^-", argv)]
+argv <- argv[! grepl("^-", argv) & ! grepl("=", argv) & ! argv == "test"]
```
0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q, line 110:
```diff
-argv <- argv[! grepl("^-", argv) & ! grepl("=", argv) & ! argv == "test"]
+argv <- grep('(^-|=|test)', argv, value = TRUE, invert = TRUE)
```

start.R
orichters: is that clearer now?

Combined --interactive, it stops in case of compilation errors, allowing the user to fix them and rerun gamscompile;

start_bundle_coupled.R
orichters: thanks.

start.R
0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q: Yes.  And also it's a sentence.

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q (approved):

start.R
orichters: when I write about compilation errors, my brain automatically switches to introducing syntax errors, I guess.

mikapfl (approved):
Looks good, and I don't care about HP-UX. Macos and GNU have `cp -n`, that's good enough, I think. If people really want to run REMIND using busybox or on HP-UX, we'll solve the problems when they arise.

PR #1100, opened 2022-11-23
Let file.copy overwrite an existing main.gms. Delete unnecessary code.

dklein-pik:
# Purpose of this PR

Avoid undesirable behavior (if main.gms could not be copied because it already existed, this was not reported).
Remove duplicated code.

## Type of change

- [x] Bug fix 

## Checklist:

- [ ]


mikapfl:
Hm, you deleted parts of the checklist, which is not really good practice imho, unless they are not applicable at all. `Rscript start.R -q` is very much applicable, I think. I would not want to merge without knowing that `Rscript start.R -q` still works.

dklein-pik:
Successfully executed `Rscript start.R -q`.

PR #1112, opened 2022-11-29
Faster compile

mikapfl:
# Purpose of this PR

Using the ideas from https://github.com/pik-piam/gms/pull/47 , we can make compiling full.gms significantly faster by using `gams action=c` instead of `gms::singleGAMSfile`. This PR is designed so that we don't have to change anything in the workflow. That means a `full.gms` is still created and is then run (which means the model will be compiled twice by GAMS, once from main.gms and once from full.gms). We can discuss if we really need full.gms or what else we want to change later, let's take the free speed boost now.

Compared to the current develop, this saves over 2 minutes each time a full.gms is built, i.e. per scenario. For me, it is especially welcome for `start.R --quick` runs, which now take under 4 minutes instead of over 6 minutes.

The only visible differences are:
* we now also have a main.lst and main.log file in the output folder, because main.gms is compiled as well.
* if gams compilation fails, no full.gms is produced. Instead, you get a message like:
> Creating full.gms
> Error in prepare() : 
>   Compiling main_testOneRegi.gms failed, stopping. Check main_testOneRegi.log , main_testOneRegi.dmp , and main_testOneRegi.lst for details.
* the singleGAMSfile-specific comments in full.gms are gone. Instead, we have comments added by GAMS with the same meaning but different wording. Check out the [full diff](https://github.com/pik-piam/gms/pull/47/files#diff-3cd5eaa7d21933422bfd26dfab6cb20e47e691a74217f6f8870d8b9ae62e32ce) as an example.

## Type of change

- [x] Refactoring

## Checklist:


0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q (approved):
Me likey!

> We can discuss if we really need full.gms or what else we want to change later, let's take the free speed boost now.

Yes, we do.  It's there to ensure portability of the run.  `main.gms` contains absolute paths to files to include ([e.g.](https://github.com/pik-piam/gms/pull/47/files#diff-3cd5eaa7d21933422bfd26dfab6cb20e47e691a74217f6f8870d8b9ae62e32ceR22)), so you can't copy the run somewhere else (modify it) and run it again, which is i.a. useful for debugging.
So we can do without the `main` files, but not without the `full` ones.

Also would suggest mikapfl/remind#6.

mikapfl:
> Yes, we do. It's there to ensure portability of the run. main.gms contains absolute paths to files to include ([e.g.](https://github.com/pik-piam/gms/pull/47/files#diff-3cd5eaa7d21933422bfd26dfab6cb20e47e691a74217f6f8870d8b9ae62e32ceR22)), so you can't copy the run somewhere else (modify it) and run it again, which is i.a. useful for debugging.
So we can do without the main files, but not without the full ones.

So maybe we can run straight from main.gms, but with dumpOpt=21, so that the equivalent of the full.gms is produced by the run. So after the run finished either way (as long as it compiles), you will have an equivalent of full.gms, but you only need to compile once. But that's an idea for later, I'd say.

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
> but you only need to compile once. But that's an idea for later, I'd say.

People weren't complaining about the three minute preprocessing enough to be taken seriously if they now complain about an additional ten second compilation.
Table that for five years, when there was sufficient personal turn-over for the argument to hold merit ;)

PR #1120, opened 2022-12-06
introduce a couple of flexible switches useful for region-specific analysis  (biomass, trade, agriculture, exogenous demand) and add new standard config for REMIND-EU

fschreyer:
# Purpose of this PR

This PR generalizes a couple of switches and adds a new REMIND-EU config file initiated by [this issue](https://github.com/remindmodel/development_issues/issues/155). 

It contains the following changes

switch changes:

* remove ``c_BaselineAgriEmiRed``, introduce more flexible ``c_agricult_base_shift`` switch that can be used to shift our agricultural emissions trajectories for specific regions
* remove ``cm_bioprod_histlim``, introduce more flexible ``cm_bioprod_regi_lim`` switch that can be used to set bounds on total biomass production in a region (also letting @merfort know) 
* remove ``cm_BioImportTax_EU``, introduce more flexible ``cm_import_tax`` switch that can be used to charge taxes on PE imports per region (e.g. ``pegas`` tax for energy security scenarios)
* introduce flexible ``cm_trade_SE_exog`` to set fixed trade flows between regions for secondary energy carriers (e.g. hydrogen), (letting @PhilippVerpoort know). 
* introduce ``cm_exogDem_scen`` switch that can be used to fix CES nodes (FE/ES demand) to exogenous trajectories/values; currently features steel and cement production trajectories for Germany used in Ariadne project read in via the parameter ``p47_exogDemScen`` (see corresponding [PR](https://github.com/pik-piam/mrremind/pull/333) in mrremind). Note that this switch is supposed to be used outside of calibration runs and will make REMIND deviate from the "optimal" trajectories it is calibrated to in the calibration run. This is mainly useful for projects with fixed demand trajectories (that cannot be met by REMIND's endogous response) focused on the energy system where macro effects are not as relevant. Moreover, it can be useful for testing purposes to reduce the complexity of REMIND's price-elastic behavior of FE/ES demands (also letting  @cchrisgong know). 

general changes:

* the ``prepare_and_run.R`` script was adapted to declare ext_regi before all_regi. This ensures that for the above (and other) switches changes for specific regions (e.g. Germany) overwrite changes for region groups (e.g. EU27). 

config changes:

* add ``scenario_config_EU.csv`` based on the latest ariadne runs and the new switches. This is supposed to become a common basic REMIND-EU config and contains only 4 runs (see [this issue](https://github.com/remindmodel/development_issues/issues/155) and the corresponding table to explain the choice of switches: https://cloud.pik-potsdam.de/index.php/apps/onlyoffice/14419250?filePath=%2FShared%2FREMIND-EU_calibration%2FREMIND_EU_Ariadne_Switches.xlsx (``Nextcloud\Shared\REMIND-EU_calibration\REMIND_EU_Ariadne_Switches.xlsx``). Let me know if you want to discuss some of this in more detail, @Renato-Rodrigues and @robertpietzcker. I don't have strong opinions on all switches but I do support having a simple joint config to base REMIND-EU runs on (which is not too far away from the REMIND-H12 configs). 

## Type of change

- [x] New feature 

## Checklist:

- [x] adapt switches after renaming in other scenario configs (last step before merging)

* Test runs are here: 

test new switches in REMIND-EU baseline run:

``/p/tmp/schreyer/Modeling/remind/Current/output/EU21reg_Base_SwitchTest2_2022-12-05_15.16.34``


modules/47_regipol/regiCarbonPrice/bounds.gms
```diff
@@ -130,6 +130,10 @@ if (cm_EnSecScen_limit gt 0,
     vm_prodPe.up(t,regi,"pegas")$((t.val ge 2025) AND (sameas(regi,"DEU"))) = cm_EnSecScen_limit/pm_conv_TWa_EJ;
 );
 
+*** Fix CES function quantity trajectories to exogenous data if cm_exogDem_scen is activated
```
Renato-Rodrigues, line 133: As discussed previously, fixing the ces quantity directly like you do here is not a good practice to be used in any scenario definition as it affects not only this CES tree node, but also the marginal rate of substitution of any nodes connected to it. 
If you really want to push this to the main REMIND code I would add a warning in every line regarding that you call this switch so it is clear to any user that this switch should only be used if you are aware of the consequences.

config/21_regions_EU11/scenario_config_EU.csv
```diff
@@ -0,0 +1,7 @@
+title;start;.description;regionmapping;extramappings_historic;CES_parameters;optimization;c_testOneRegi_region;cm_nash_mode;subsidizeLearning;trade;c_ccsinjecratescen;cm_bioprod_regi_lim;cm_import_tax;cm_1stgen_phaseout;c_agricult_base_shift;cm_altFeEmiFac;cm_CESMkup_build;cm_CESMkup_ind;cm_rcp_scen;cm_iterative_target_adj;c_budgetCO2from2020;carbonprice;cm_co2_tax_2020;cm_peakBudgYr;cm_taxCO2inc_after_peakBudgYr;cm_CO2priceRegConvEndYr;regipol;cm_emiMktTarget;cm_NucRegiPol;cm_CoalRegiPol;capitalMarket;industry;buildings;transport;cm_EDGEtr_scen;cm_emiscen;cm_POPscen;cm_GDPscen;cm_demScen;cm_startyear;path_gdx;path_gdx_ref
```
Renato-Rodrigues, line 1: I would remove this scenario config file from this PR.
It has being agreed that we shouldn't create more scenario config files in the subfolder 21_regions_EU11.
Also, the default scenario config for EU21 should be added to the default REMIND scenario config file and support the default budget runs supported in REMIND.
We can get together to better define this scenario config switches in an smaller group before adding it to the code.

Renato-Rodrigues (changes requested):

modules/47_regipol/regiCarbonPrice/bounds.gms
robertpietzcker: I think we had agreed that Felix should push this to the trunk now. At a later point in time, it would be interesting to compare runs using the two approaches - my bet would be that the marginals in the rest of the CES tree are the same.

fschreyer: This is what I added to the description. I think it should address your point. If the CES function is forced away from the optimal point, then the prices of the CES nodes do not necessarily align with the price signal from the energy system anymore. 
```
*** cm_exogDem_scen
*** switch to fix FE or ES demand represented in CES function to trajectories 
*** from exgenous sources (not EDGE models) given in file p47_exogDemScen.
*** This switch fixes demand without recalibration of REMIND CES parameters.
*** This should be kept in mind when comparing those runs to baseline runs without fixing
*** as the fixing shifts the CES function away from its optimal point based on the CES parameters used.
```

Renato-Rodrigues: I read the comment the first time I reviewed the PR.
I ask you to please add an extra line to it saying something like:
```
*** warning: the formulation fixing CES quantity nodes in scenarios should be used with care and parsimony. Price and tax induced solutions are preferable from the REMIND formulation perspective, and consequences of fixing CES tree nodes directly still require further investigation. 
```

fschreyer: Ok, added

config/21_regions_EU11/scenario_config_EU.csv
fschreyer: done

Renato-Rodrigues (approved):

PR #1125, opened 2022-12-12
fix incompatibility between techpol scenarios and cm_nucscen 5

Renato-Rodrigues:
# Purpose of this PR

Allow the use of techpol realisations with cm_nucscen = 5.

## Type of change

Disable tnrs lower bound definition in techpol realisations if cm_nucscen is set to 5, allowing complete phase-out of nuclear technology. 

- [x] Bug fix 

## Checklist:


orichters:
@Renato-Rodrigues: I had no idea about the problem. What do you expect me to do as reviewer?

Renato-Rodrigues:
@orichters I added you as a reviewer as you recently worked with the NDCs and so you can give your opinion about these change if you feel like.

The NDC code set lower bounds to IND and CHA between 2025 and 2040 for nuclear power capacity. For the IND case, this capacity increases from 2030 to 2035. 
Meanwhile the cm_nucscen scenario 5 assumes that no new nuclear investments are made after 2020. Therefore if you enable both together you get an infeasibility. 

I am not sure how much of the capacity for both CHA and IND in the NDC `p40_TechBound` parameter is actually already existent capacity or new additions after 2020.
So the change proposed in this PR is simply disabling the NDC nuclear capacity bounds, if you opt to run the model with cm_nucscen = 5.

robertpietzcker (approved):
Looks good. 

Maybe make the comment even more explicit, eg
"!! resetting lower bound to value defined at cm_nucscen switch"
->
"!! we assume: Nucscen (limiting nuclear deployment) overrides NDC targets -> resetting lower bound to value defined at cm_nucscen switch"

Renato-Rodrigues:
> Nucscen (limiting nuclear deployment) overrides NDC targets -> resetting lower bound to value defined at cm_nucscen switch

added suggested change at commit https://github.com/remindmodel/remind/pull/1125/commits/44b1235b167bc98c5351b9c4c3eca27ab65c0129

PR #1133, opened 2022-12-19
convert compile-time `c_CES_calibration_iteration` to run-time scalar

mikapfl:
# Purpose of this PR

After #1112 , calibration runs were broken because `c_CES_calibration_iteration` was a compile-time variable (`$setglobal`), but prepare_and_run tried to change it in `full.gms`. That used to work because `singleGAMSfile` didn't actually compile main.gms to write full.gms, but only evaluated compile-time variables internally to resolve `include`s and `batincludes`. After #1112, main.gms is fully compiled, which evaluates compile-time variables and replaces them by their value in full.gms. Therefore, it is not possible to replace `c_CES_calibration_iteration` in full.gms any more.

The solution for the problem in this PR is to convert `c_CES_calibration_iteration` to a scalar `sm_CES_calibration_iteration`, which is a run-time variable and thus will survive into full.gms. That unfortunately means I had to touch all places where this was used and convert them from compile-time code to run-time code, which is a bit tedious and bug-prone.

I also looked at an alternative way to solve this: converting the calibration loop in `run()` such that it would re-compile from main.gms for subsequent runs. However, this also introduces a lot of changes because then `submit()` has to be split into multiple parts such that only the relevant parts can be re-run, and potentially also reporting needs to be changed to find the results. I personally would find it more coherent if each calibration iteration would be `submit()`ted individually, one after each other, but that would mean individual output folders for each iteration (which would be the point, such that each iteration would be re-startable on its own) and therefore introduce even more changes all over the place. Therefore, I think converting `c_CES_calibration_iteration` is less invasive now.

I started a test calibration run in `/p/tmp/pflueger/remind/output/SSP2EU-AMT-calibrate_2022-12-20_15.17.19`

## Type of change

- [x] Bug fix 

## Checklist:

- [ ] Calibration runs deliver the expected output

### DIFF
--- a/core/declarations.gms
+++ b/core/declarations.gms
@@ -600,6 +600,8 @@ sm_globalBudget_dev                                   "actual level of global cu
 
 sm_eps                                                "small number: 1e-9 "  /1e-9/
 
+sm_CES_calibration_iteration                          "current calibration iteration number, updated from R" /1/
+
 ***----------------------------------------------------------------------------------------
 ***----------------------------------------------trade module------------------------------
 ;
--- a/main.gms
+++ b/main.gms
@@ -1594,7 +1594,6 @@ $setGLobal cm_debug_preloop  off    !! def = off
 $setGlobal cm_APscen  SSP2          !! def = SSP2
 $setglobal cm_CES_configuration  indu_subsectors-buil_simple-tran_edge_esm-POP_pop_SSP2EU-GDP_gdp_SSP2EU-En_gdp_SSP2EU-Kap_debt_limit-Reg_62eff8f7   !! this will be changed by start_run()
 $setglobal c_CES_calibration_iterations  10     !!  def  =  10
-$setglobal c_CES_calibration_iteration  1     !!  def  =  1
 $setglobal c_CES_calibration_industry_FE_target  1
 $setglobal c_testOneRegi_region  EUR       !! def = EUR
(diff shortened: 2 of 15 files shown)


mikapfl:
The `--quick` run in `/p/tmp/pflueger/remind/output/SSP2EU-AMT-calibrate_2022-12-20_15.18.52` looks pretty convincing, so I think we can merge.

mikapfl:
argh, no, reporting is still weird because `iteration` is written as float instead of decimal. :slightly_frowning_face:

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
> argh, no, reporting is still weird because `iteration` is written as float instead of decimal. slightly_frowning_face

Yeah, working around that is not worth the effort.  These files get read by R, and R doesn't care.

mikapfl:
> Yeah, working around that is not worth the effort. These files get read by R, and R doesn't care.

Ah, okay. `reportCEScalib` doesn't work on the `--quick` run, but this could also be because it contains only one region. We'll know for sure if the reporting works once the calibration run finishes tomorrow.

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q (approved):
[This](https://github.com/mikapfl/remind/pull/8), then go!

PR #1113, opened 2022-12-01
new CES parameters and gdx files for EU21_SSP2EU, add inputdataInfo.log

LaviniaBaumstark:
# Purpose of this PR
new CES parameters and gdx files for EU21_SSP2EU for input data revision 6.324 and also added inputdataInfo.log to the CES parameter folder containing information on input data revision the CES calibration is based on. This info will be downloaded together with the CES parameters. Adjusted tutorial 06_Advanced_ChangeInputs and https://redmine.pik-potsdam.de/projects/remind-r/wiki/GDX_and_CES_parameter_Handling accordingly

## Type of change

- [x] Refactoring
- [x] This change requires a documentation update

## Checklist:

- [x] My code follows the coding etiquette
- [x] I have performed a self-review of my own code
- [x] The model runs successfully (`Rscript start.R -q`)

## Further information (optional):

* Test runs are here: 
* Comparison of results (what changes by this PR?):

PR #1134, opened 2022-12-21
Fix compilation error.

mikapfl:
# Purpose of this PR
Fix compilation error introduced in #1113 due to missing value in declaration.


0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
My bad.  I tested a proper run, but not the test.

mikapfl:
Nono, I would like to own this error, I didn't test the PR before merging, which is my responsibility as the owner of the PR. After travelling yesterday with deutsche bahn I would like to cite them:

> We apologize for *any* inconvenience.

PR #1135, opened 2022-12-22
remove all the latestSnapshot and automatic searches from .snapshot.Rprofile

orichters:
## Purpose of this PR

- avoid that people let the script find a suitable snapshot for them. Instead, you have to specify one yourself.
- the snapshots should only be used with coupled runs, and then you want stability, so no automatic searching for snapshots
- has created confusion already
- first wanted to use `2022_12_R4`, but this contains a broken variant of `piamenv` which creates fails. So using a newer snapshot.

## Type of change

- [x] minor change to reduce confusion and failing runs because of bad snapshots picked.

## Checklist:

- [x] All automated model tests pass (`FAIL 0` in the output of `make test`).

### DIFF
--- a/.snapshot.Rprofile
+++ b/.snapshot.Rprofile
@@ -4,13 +4,11 @@
 
 local({ # prevent variables defined here from ending up in the global env
 
-# snapshots created with R 4.1 have a _R4 postfix
-maybeR4 <- if (R.version$major == 4) "_R4" else ""
-# Gather all library snapshots
-snapshots <- sort(list.files(path = '/p/projects/rd3mod/R/libraries/snapshots',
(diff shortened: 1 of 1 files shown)


mikapfl:
> currently, using snapshots in general fails with

Fixed in: https://github.com/pik-piam/piamenv/pull/4

pfuehrlich-pik (approved):

orichters:
> > currently, using snapshots in general fails with
> 
> Fixed in: [pik-piam/piamenv#4](https://github.com/pik-piam/piamenv/pull/4)

Yeah, thanks. But was still in the `2022_12_R4` snapshot, so I went to `2022_12_15_R4` instead…

mikapfl (approved):
I think it makes a lot of sense to explicitly have a default snapshot version in REMIND develop so that there is less guessing which snapshot currently fulfils all version requirements. Thanks!

PR #1143, opened 2023-01-06
bugfix on KWTCint damages

gabriel-abrahao:
# Purpose of this PR
Fixes a domain and a division by zero error in `50_damages/KWTCint` when it was activated

## Type of change

- [x] Bug fix 
## Checklist:

- [X] I have updated the in-code documentation
- [] All automated model tests pass (`FAIL 0` in the output of `make test`)

There was a change in the input data generation where the tropical cyclone damage parameters (`f50_TC_df_tasK` etc) used to be defined only for the countries where these damages happen (set `isoTC`), and now they are defined for all countries (`iso`), with zeroes in countries where they don't apply. This caused a domain violation when trying to read them just in `isoTC`.

This PR changes the set used on reading, and also limits the calculation of `pm_GDPfrac` from `tall` to `ttot` after 2005 to prevent divisions by zero. The years in between timesteps are filled by interpolation in a later step anyway.

* Test runs are here: 
`/p/projects/piam/abrahao/runs_gcsf/phase2_dec/remind/output/SSP2-NDC_2023-01-05_15.37.09`

### DIFF
--- a/modules/50_damages/KWTCint/datainput.gms
+++ b/modules/50_damages/KWTCint/datainput.gms
@@ -24,7 +24,7 @@ p50_damageFuncCoefTC0(isoTC) = 0;
 p50_damageFuncCoefTC1(isoTC) = 0;
 
 *** load TC damage parameter data
-parameter f50_TCconst(isoTC,all_TCpers,all_TCspec)	"damage parameter constant"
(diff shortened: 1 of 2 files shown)


fpiontek (approved):
Looks good. For consistency, maybe change the equation for "combined regional damage" in modules/50_damage/KWTCint/postsolve (line 53) to also be calculated starting in 2005, but that shouldn't matter for the result.

PR #1144, opened 2023-01-06
bug fix - using p_priceCO2forMAC for mac curves instead of pm_priceCO2

Renato-Rodrigues:
# Purpose of this PR

- change code to use `pm_priceCO2forMAC` instead of `p_priceCO2` for `pm_CementAbatementPrice`, `pm_macStep(ttot,regi,"ch4gas")` and `pm_macStep(ttot,regi,"ch4coal")` because the first has support for the regional carbon prices.
- Filling `p_priceCO2` values for co2 prices defined by regional emission targets in case this parameter is used for reporting reasons.
- Controlling `pmpriceCO2forMAC` to avoid assigning values to unused dimension combinations.

## Type of change

- [x] Bug fix 

## Checklist:

- [x] make sure that the remind2 R library supports the renaming of `pm_priceCO2` to `p_priceCO2` and `p_priceCO2forMAC` to `pm_priceCO2forMAC`

outdated comment:
@0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q I removed entirely the fixed_shares presolve file as apparently the same code is already calculated at the core presolve file. 
The only code I was not 100% sure is the line below:
```
*** zero out a ghost
vm_macBase.fx(ttot,regi,emiInd37_fuel) = 0;
```
I guess you control that somehow already for the case of industry subsectors. Please double check if it is necessary to do any code change regarding to this, or if this line needs to be moved to the core presolve.


0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q (changes requested):
> I removed entirely the fixed_shares presolve file as apparently the same code is already calculated at the core presolve file.

Ummm, no, it is not.

```
if ( NOT (cm_IndCCSscen eq 1 AND cm_CCS_cement eq 1),
*** Cement (clinker) production causes process emissions of the order of 
*** 0.5 t CO2/t Cement. As cement prices are of the magnitude of 100 $/t, CO2
*** pricing leads to significant price markups.


  pm_CementAbatementPrice(ttot,regi)$( ttot.val ge 2005 )
  = pm_priceCO2(ttot,regi) / sm_C_2_CO2;
  ```
https://github.com/Renato-Rodrigues/remind/blob/8e3221ee4fc59c70ebf6a48b528ba1a271cd1113/core/presolve.gms#L189-L195
  ```
  if (cm_IndCCSscen eq 1 AND cm_CCS_cement eq 1,


  display "CO2 price applied for Cement Demand Reduction [$/tC]", pm_priceCO2;


  !! lowest price for which abatement equals current abatement
  pm_CementAbatementPrice(ttot,regi)$( ttot.val ge 2005 )
  = max(0,
        smin(steps$(   pm_abatparam_Ind(ttot,regi,"co2cement",steps)
                    ge pm_macAbatLev(ttot,regi,"co2cement") ),
          steps.val - 1.5 !! average upper and lower step
        )
    )
  * sm_dMAC;
  ```
  https://github.com/Renato-Rodrigues/remind/blob/8e3221ee4fc59c70ebf6a48b528ba1a271cd1113/modules/37_industry/fixed_shares/presolve.gms#L13-L25

**Comment by Renato-Rodrigues** [comment]
I reverted the changes made in the fixed shares module and added code check required renames.

File: modules/37_industry/fixed_shares/presolve.gms [hunk 1]
```diff
@@ -12,7 +12,7 @@ vm_macBase.fx(ttot,regi,emiInd37_fuel) = 0;
 *** adjust CO2 cement process emissions
 if (cm_IndCCSscen eq 1 AND cm_CCS_cement eq 1,
 
-  display "CO2 price applied for Cement Demand Reduction [$/tC]", pm_priceCO2;
+***  display "CO2 price applied for Cement Demand Reduction [$/tC], check for co2cement:", pm_priceCO2forMAC;
```
0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q, line 15: Why's that commented out?  Either use it or delete it, I'd say.

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q (approved):

modules/37_industry/fixed_shares/presolve.gms
0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q: No, no way of doing that.

Issue #1149, opened 2023-01-11
pm_SEprice assignment repeated unnecessarily all over the model. 


Renato-Rodrigues:
There is unnecessary code repetition for the definition of `pm_SEprice` in the model. 

It is currently repeated in multiple places in the model. However, most probably only two assignments are really required: one at the core/preloop, and another in the core/postsolve.
 
Repeating the assignment at every power sector module realization, and at every presolve is quite unnecessary, extends the code length and can create potential oversight issues on any future change that can affect this parameter.

Issue #1115, opened 2022-12-01
SSP2EU-NDC runs crash


dbantje:
Currently, the SSP2EU-NDC run crash with an execution error. See for example the automated test results:

```
Run                                 Runtime      inSlurm  RunType      RunStatus           Iter              Conv                   modelstat            Mif       inAppResults
default-AMT-_2022-11-26_00.05.49    6.3 hours    FALSE    nash         Normal completion   40/100            converged              2: Locally Optimal   TRUE      TRUE          
SDP-AMT-Base_2022-11-26_00.09.30    NA           FALSE    nash         Execution error     1/100             NA                     NA                   FALSE     NA            
SSP1-AMT-Base_2022-11-26_00.08.32   6.2 hours    FALSE    nash         Normal completion   39/100            converged              2: Locally Optimal   TRUE      TRUE          
SSP2EU-AMT-Base_2022-11-26_00.08.5  7.9 hours    FALSE    nash         Normal completion   54/100            converged              2: Locally Optimal   TRUE      TRUE          
SSP2EU-AMT-calibrate_2022-11-26_00  2 days       FALSE    Calib_nash   Normal completion   38/100 Clb: 10    converged              2: Locally Optimal   TRUE      TRUE          
SSP2EU-AMT-NDC_2022-11-26_08.15.06  NA           FALSE    nash         Execution error     17/100            722222527222           6: Intermed Infes    FALSE     NA            
SSP5-AMT-Base_2022-11-26_00.09.11   NA           FALSE    nash         Execution error     1/100             NA                     NA                   FALSE     NA            
testOneRegi-AMT-Base_2022-11-26_00  44.3 mins    FALSE    testOneRegi  Normal completion   1/1               NA                     2: Locally Optimal   TRUE      TRUE     

```

or the these two runs I started (also during iteration 17):

```
/p/tmp/davidba/remind/output/SSP2EU-NDC_off_2022-11-24_19.04.11
/p/tmp/davidba/remind/output/SSP2EU-NDC_on_2022-11-24_17.34.03
```

Here's the abort from the second run:
```
----3518138 Run was aborted because the maximum number of consecutive failures was reached in at least one region!
**** Exec Error at line 3518138: Execution halted: abort$5 'Run was aborted because the maximum number of consecutive failures was reached in at least one region!'

```
In that run, first infeasibilities already show up in iteration 2:
```
  2   solvestat              modelstat                 resusd     objval
[...]

Issue #1137, opened 2022-12-29
q39_EqualSecShare_BioSyn cause either slow solving time or direct infeasibilities to the model


Renato-Rodrigues:
I had quite a few warnings recently from conopt about small jacobians related with the equation q39_EqualSecShare_BioSyn, for example:
```
**** ERRORS/WARNINGS IN EQUATION q39_EqualSecShare_BioSyn(2025,REF,fepet,trans,other)
     1 error(s): All Jacobian elements in the row are very small.
...
```

This is a bad sign as it means that the function changes very slowly with changes in the equation, so if the model has infeasibilities regarding this equation it will have a very hard time getting ride of them.

More recently, this got worst as this equation is causing infeasibilities.
Take for example: `/p/projects/remind/users/renatoro/ECEMF/v09_2022_12_16/output/debug_WP1_NPI_2022-12-29_13.18.35`   

You can see the small jacobians problem in the "REF" region, but if you take a look at the "SSA" region, you will see that this equation is causing infeasibilities to the model.

SSA has fepet demand for seliqbio, seliqfos anbd seliqsyn. However, it has zero seliqsyn and seliqbio demand in fedie (in 2090).
As the share equation enforces the seliqbio and seliqsyn shares to be the same in fedie and fepet, this is clearly impossible to the model to solve.

Sometimes the model could get luck and get around this issue in next iterations if something changes in the biomass trades for example, but usually it is much more common that this problem gets worst at each iteration.

A possible work around for this problem would be to replace the share equation with a similar approach to the Chebyschev minmax formulation that I implemented for dealing with biomass switching between sectors.
Another more long term solution would be to deal with the trade-offs between alternative fuels directly in the transport sector code.

Any thoughts about that? @robertpietzcker @fschreyer @pweigmann @johannah-pik  

For now I am having to disable this equation in my production runs entirely, which is not desirable if we want to show transportation detailed numbers.

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
> More recently, this got worst as this equation is causing infeasibilities. Take for example: `/p/projects/remind/users/renatoro/ECEMF/v09_2022_12_16/output/debug_WP1_NPI_2022-12-29_13.18.35`
> 
> You can see the small jacobians problem in the "REF" region, but if you take a look at the "SSA" region, you will see that this equation is causing infeasibilities to the model.
> 
[...]

PR #1150, opened 2023-01-11
fix to cm_flex_tax bug

Renato-Rodrigues:
# Purpose of this PR

Aims to fix the problem discussed here: https://github.com/remindmodel/remind/issues/1115
The proposed solution follows the addition of a non negative `pm_SEprice` parameter and its use in the q32_flexAdj equation.

Test runs with and without the change can be found here (still running):
- `/p/projects/remind/users/renatoro/Debug_trunk/2023_01_11_cm_flex_tax/output/SSP2EU-AMT-Base_2022-12-24_00.08.15`
- `/p/projects/remind/users/renatoro/Debug_trunk/2023_01_11_cm_flex_tax/output/SSP2EU-AMT-NDC_no_change_2023-01-11_10.33.56`

## Type of change

- [x] Bug fix 
- [x] Fundamental change

## Checklist:

## Further information:

- There is unnecessary code repetition for the definition of `pm_SEprice`, currently repeated in multiple places in the model.
- only two assignments are really required: one at the `core/preloop`, and another in the `core/postsolve`. Repeating the assignment at every power sector module and at every presolve is quite unnecessary.
- This refactoring was not done in this PR, and an issue was added to mark this refactoring as pending. https://github.com/remindmodel/remind/issues/1149

### DIFF
--- a/core/declarations.gms
+++ b/core/declarations.gms
@@ -232,7 +232,6 @@ pm_delta_histCap(tall,all_regi,all_te)                "parameter to store data o
 pm_FEPrice(ttot,all_regi,all_enty,sector,emiMkt)      "parameter to capture all FE prices across sectors and markets (tr$2005/TWa)"
 pm_FEPrice_iter(iteration,ttot,all_regi,all_enty,sector,emiMkt) "parameter to capture all FE prices across sectors and markets (tr$2005/TWa) across iterations"
 pm_SEPrice(ttot,all_regi,all_enty)                    "parameter to capture all SE prices (tr$2005/TWa)"
-pm_SEPrice_noNegatives(ttot,all_regi,all_enty)        "non negative values of pm_SEPrice parameter (tr$2005/TWa)"
 pm_PEPrice(ttot,all_regi,all_enty)                    "parameter to capture all PE prices (tr$2005/TWa)"
 
 p_FEPrice_by_SE_Sector_EmiMkt(ttot,all_regi,entySe,all_enty,sector,emiMkt) "parameter to save FE price per SE, sector and emission market (tr$2005/TWa)"
--- a/core/declarations.gms
+++ b/core/declarations.gms
@@ -232,6 +232,7 @@ pm_delta_histCap(tall,all_regi,all_te)                "parameter to store data o
 pm_FEPrice(ttot,all_regi,all_enty,sector,emiMkt)      "parameter to capture all FE prices across sectors and markets (tr$2005/TWa)"
 pm_FEPrice_iter(iteration,ttot,all_regi,all_enty,sector,emiMkt) "parameter to capture all FE prices across sectors and markets (tr$2005/TWa) across iterations"
 pm_SEPrice(ttot,all_regi,all_enty)                    "parameter to capture all SE prices (tr$2005/TWa)"
+pm_SEPrice_noNegatives(ttot,all_regi,all_enty)        "non negative values of pm_SEPrice parameter (tr$2005/TWa)"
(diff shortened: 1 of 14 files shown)


0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
> My suggestion at the IEA-Update channel was in this line, but instead of creating an additional parameter, I suggested to add a dollar condition to the right hand side of the equation so it would be zero when prices are negative. Both should theoretically work in the same way.

Why did you go for the additional parameter?  One more thing in the .gdx file …

Renato-Rodrigues:
I could go either way. 
I tend to agree with you that using less memory is always better, but others prefer to have "easier to read" code. Dollar conditionals are a bit less straight forward to read to some.

I will adjust the PR to whatever solution the reviewers vote for, so if @LaviniaBaumstark  and @robertpietzcker have a preference for the parameter way, I would leave as is. If they are ok with the dollar conditional way, I will make the necessary changes to adapt this PR to that solution instead.

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
Yeah, but … :point_up: there are dollar-conditions already, so not much is lost I would argue.

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q (approved):

robertpietzcker:
Thanks for the fix, Renato!

It seems I am not yet up to full speed - yesterday I somehow thought using a $ conditional in the multiplication in q32_flexAdj to take out the negative values would somehow cause a problem as it would leave vm_flexAdj free in that case (instead of fixing it to 0), but now I don't understand anymore why I thought so. 

I am sorry for the additional hassle!

Please go with the version you prefer, Renato.

Renato-Rodrigues:
> It seems I am not yet up to full speed - yesterday I somehow thought using a $ conditional in the multiplication in q32_flexAdj to take out the negative values would somehow cause a problem as it would leave vm_flexAdj free in that case (instead of fixing it to 0), but now I don't understand anymore why I thought so.

My original post in the IEA channel had a mistake as the dollar conditional was applied to the equation declaration and not to the right hand side of it. It was probably that version that you look at.
I mentioned that issue just after I wrote the original equation in the same mattermost channel: https://mattermost.pik-potsdam.de/rd3/pl/njtnam3mkfbqdcgemdog6s68yy , but after that I editted the original post to avoid the misleading message.

I will set a run with the dollar conditional alternative and compare the solutions. If everything works the same way I will update this PR and use the dollar alternative instead.

LaviniaBaumstark:
I also prefer the dollar condition. In my first test this led to unbounded variables but probably I did a mistake. So, let's wait for your tests and if it is working, go with the dollar condition

Renato-Rodrigues:
I changed the solution changed to use the dollar formulation.
I set a test run at:  `/p/projects/remind/users/renatoro/Debug_trunk/2023_01_11_cm_flex_tax_dollar/output/SSP2EU-AMT-NDC_dollar_2023-01-11_18.10.15`
tomorrow I will check the results to merge this in case of success.

Renato-Rodrigues:
At least the test runs I tried were successful.
The model solution time also reduced at least half hour due to the need for less iterations. It converged in the minimum number of iterations possible (28 or 29 iterations depending of the run) instead of 38 iterations from the previous successful NDC run.

I still got the issue I described here https://github.com/remindmodel/remind/issues/1137 in one of the runs although.

I am merging this PR as is, but we still need to verify if that is enough to solve the issue with the marginals use in equations, and if this issue was causing the REF and MEA consumption losses described by Marian in the validation results.
I am not sure about that yet.

PR #1151, opened 2023-01-11
search for mifs and gdx if output script needs mif

orichters:
## Purpose of this PR
- old: if starting `output.R` and selecting `compareScenarios2` or `policyCosts`, it offers you runs with a `fulldata.gdx` where the reporting `*.mif` is missing (for whatever reason) but then the scripts fail
- now: if a scripts needs the mif, only show runs where both fulldata.gdx and mif file exists
- in `policyCosts`, avoid that two running scripts simultaneously operate on the same tmp folder

## Type of change

- [x] Bug fix 
- [x] Refactoring

## Checklist:

### DIFF
--- a/scripts/output/comparison/policyCosts.R
+++ b/scripts/output/comparison/policyCosts.R
@@ -65,12 +65,11 @@ policy_costs_pdf <- function(policy_costs,
                  "<<echo=false>>=",
                  "options(width=110)",
                  "@")
-
+  tmpfolder <- paste0("tmp_", gsub("[\\. ]*", "", basename(fileName)))
   # Create temporary folder in which to create the policyCost pdf
-  system("mkdir tmp_policyCost")
-
+  dir.create(tmpfolder)
   # Open stream in tmp_folder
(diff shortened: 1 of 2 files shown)


scripts/output/comparison/policyCosts.R
```diff
@@ -65,12 +65,11 @@ policy_costs_pdf <- function(policy_costs,
                  "<<echo=false>>=",
                  "options(width=110)",
                  "@")
-
+  tmpfolder <- paste0("tmp_", gsub("[\\. ]*", "", basename(fileName)))
   # Create temporary folder in which to create the policyCost pdf
-  system("mkdir tmp_policyCost")
-
+  dir.create(tmpfolder)
```
mikapfl, line 70: Why not [with_tempdir()](https://withr.r-lib.org/reference/with_tempfile.html)? That would also take care of deleting the tempdir later on.

mikapfl (approved):
Thanks, looks good! If you fancy a little refactoring, consider my comment below, but you can also just merge, the PR is already a nice improvement.

PR #1155, opened 2023-01-11
Fix coupling issue with path_mif_ghgprice_land

gabriel-abrahao:
# Purpose of this PR
Fix a coupling issue that made the carbon prices read from `path_mif_ghgprice_land` be applied in MAgPIE just on the first coupling iteration, but not on the subsequent ones. 

## Type of change

- [x] Bug fix 

## Checklist:

- [xI have performed a self-review of my own code

The configuration option `path_mif_ghgprice_land` in `scenario_config_coupled_*.csv` is supposed to provide either a `.mif` file or another depended scenario from which to take GHG prices to be passed to MAgPIE. This is in opposition to the standard behavior of using the carbon prices of the same scenario, and is used to essentially set separate exogenous carbon prices to MAgPIE only (independent of the carbon price behavior in REMIND). However, a condition in `start_bundle_coupled.R` was only configuring the relevant settings in MAgPIE on the first coupling iteration, leading to different or no GHG prices on MAgPIE in later iterations. This PR simply removes that condition.


orichters (approved):
Thanks, as discussed…

PR #1160, opened 2023-01-13
fastest possible phaseout of traditional biomass for SDP_XX scenarios

bs538:
# Purpose of this PR
SDP_EI, SDP_MC, SDP_RC now also have the fastest possible phaseout of traditional biomass (no new capacities allowed). Previously this was just the case for the original "SDP".

## Type of change
- [x] Bug fix 

## Checklist:
("make test" reports: "nothing to be done")

Only affects the SDP_XX scenarios. Faster phaseout of traditional biomass in IND, SSA, balanced by a slight increase in (non-traditional) biomass and coal. Other variables virtually unchanged.

Note: This is a quick fix, jumping ahead of https://github.com/remindmodel/remind/pull/1054 where further discussion is needed.

### DIFF
--- a/core/bounds.gms
+++ b/core/bounds.gms
@@ -84,8 +84,11 @@ $endif
   );
 );
 
-* quickest phaseout in SDP (no new capacities allowed), quick phaseout in SSP1 und SSP5
-$if %cm_GDPscen% == "gdp_SDP"  vm_deltaCap.fx(t,regi,"biotr","1")$(t.val gt 2020) = 0. * vm_deltaCap.lo(t,regi,"biotr","1");
+* quickest phaseout in SDP scenarios (no new capacities allowed), quick phaseout in SSP1 und SSP5
+$if %cm_GDPscen% == "gdp_SDP" vm_deltaCap.fx(t,regi,"biotr","1")$(t.val gt 2020) = 0;
+$if %cm_GDPscen% == "gdp_SDP_EI" vm_deltaCap.fx(t,regi,"biotr","1")$(t.val gt 2020) = 0;
+$if %cm_GDPscen% == "gdp_SDP_MC" vm_deltaCap.fx(t,regi,"biotr","1")$(t.val gt 2020) = 0;
+$if %cm_GDPscen% == "gdp_SDP_RC" vm_deltaCap.fx(t,regi,"biotr","1")$(t.val gt 2020) = 0;
 $if %cm_GDPscen% == "gdp_SSP1" vm_deltaCap.fx(t,regi,"biotr","1")$(t.val gt 2020) = 0.5 * vm_deltaCap.lo(t,regi,"biotr","1");
 $if %cm_GDPscen% == "gdp_SSP5" vm_deltaCap.fx(t,regi,"biotr","1")$(t.val gt 2020) = 0.5 * vm_deltaCap.lo(t,regi,"biotr","1");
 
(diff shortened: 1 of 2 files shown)


merfort (approved):
Looks all good!

Renato-Rodrigues:
You definitely do not need compilation time conditionals to define bounds. Maybe in the future avoid using those.

bs538:
@Renato-Rodrigues: casually shifting blame to whoever did the original SSP implementation :)
@merfort: perhaps we can take this up in your PR?

dklein-pik:
If I remember correctly we tied the choice of the bounds to the `$setglobal cm_GDPscen` because this setglobal exists anyway and we didn't want to create an additional switch, which we would then have had to make dependent on the same setglobal. But I am happy to learn a better solution.

Renato-Rodrigues:
in the case of this PR you could use this:
`if(sameas("gdp_SDP","%cm_GDPscen%") vm_deltaCap.fx(t,regi,"biotr","1")$(t.val gt 2020) = 0;);`
instead of this
`$if %cm_GDPscen% == "gdp_SDP" vm_deltaCap.fx(t,regi,"biotr","1")$(t.val gt 2020) = 0;`

or even better:
```
set SDP_GDP_scenarios(all_GDPscen) /
  gdp_SDP, gdp_SDP_EI, gdp_SDP_MC, gdp_SDP_RC
/;

if (SDP_GDP_scenarios("%cm_GDPscen%"),
  vm_deltaCap.fx(t,regi,"biotr","1")$(t.val gt 2020) = 0;
);
```

Global defined variables are replaced in compilation time by their values in the code, so you can use normal assignments like the ones above.

Compilation time conditionals should be mainly used when you want to (1) set a global variable, (2) insert literal code, e.g. the code modules that we insert in remind or external data saved on other files, or (3) to exclude a chunk of code from the lst file if the switch is not enabled.

Although we used to use a lot the compilation time conditionals for the case 3, we should avoid that as we saw from recent discussions in the remind and rse channels that this type of conditionals have very strong limitations on supporting logical operators, leading to hidden bugs in the code.
In the case of this PR for example, the case 3 is not even the case, as we are not saving any lines in the code.

PS: all code in this message was not tested and only serves as a guideline.

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
> ```
> set SDP_GDP_scenarios(all_GDPscen) /
>   gdp_SDP, gdp_SDP_EI, gdp_SDP_MC, gdp_SDP_RC
> /;
> 
> if (SDP_GDP_scenarios("%cm_GDPscen%"),
>   vm_deltaCap.fx(t,regi,"biotr","1")$(t.val gt 2020) = 0;
> );
> ```
@bs538: from all the solutions discussed, this is the cleanest in my opinion. 
>  the grepl-like solution would have the advantage that if I add future SDP scenarios with different names ("SDP_whatever"), they would still get the same behaviour I don't have to manually adjust the code again.

But it is also quite opaque to operate on literal set names instead of explicit set membership.
```diff
 all_demScen    "all possible demand scenarios"
   /
+  !! add additional SDP scenarios to SDP_GDP_scenarios too
   gdp_SDP         "SDP fastGROWTH medCONV"
   gdp_SDP_EI
   gdp_SDP_MC
   gdp_SDP_RC 

   gdp_SSP1        "SSP1 fastGROWTH medCONV"
   …
```
Problem solved.

Issue #1157, opened 2023-01-12
Wrong / outdated parameters in scenario configs.


mikapfl:
Probably related to the code cleaning, we have quite a lot of scenario configs with wrong or outdated parameters, which will be ignored in new REMIND runs. Thanks to @orichters for writing the necessary code to find that out cleanly!

Affected scenario configs and parameters are:

* `scenario_config_DeepEl.csv`:
   - [x] `cm_buildings_scen`: https://github.com/remindmodel/remind/pull/1174
* `scenario_config_EDGE-T_NDC_NPi_pkbudget.csv`
   - [x] `cm_shSynTrans`: https://github.com/remindmodel/remind/pull/1169
* `scenario_config_GCS.csv`:
   - [x] `cm_CO2priceRegConvYr`: https://github.com/remindmodel/remind/pull/1174
   - [x] `cm_INCONV_PENALTY_bioSwitch`: https://github.com/remindmodel/remind/pull/1168
* `scenario_config_NAVIGATE_300.csv`:
   - [x] `cm_bioenergy_tax`: 
   - [x] `cm_bioprod_histlim`: https://github.com/remindmodel/remind/pull/1184
   - [x] `cm_BioImportTax_EU`: https://github.com/remindmodel/remind/pull/1184
   - [x] `c_BaselineAgriEmiRed`: https://github.com/remindmodel/remind/pull/1184
   - [x] `cm_ariadne_VRECapFac_adj`: https://github.com/remindmodel/remind/pull/1184
   - [x] `cm_OILRETIRE`: https://github.com/remindmodel/remind/pull/1174
* `21_regions_EU11/scenario_config_21_EU11_ARIADNE.csv`:
   - [x] `cm_biotrade_phaseout`
* `21_regions_EU11/scenario_config_21_EU11_ECEMF.csv`:
   - [x] `cm_bioprod_histlim`: https://github.com/remindmodel/remind/pull/1184
   - [x] `cm_BioImportTax_EU`: https://github.com/remindmodel/remind/pull/1184
   - [x] `cm_build_costDecayStart` https://github.com/remindmodel/remind/pull/1169
   - [x] `cm_shSynTrans`
* `21_regions_EU11/scenario_config_21_EU11_Fit_for_55_sensitivity.csv`:
   - [x] `cm_bioprod_histlim`
   - [x] `cm_BioImportTax_EU`: https://github.com/remindmodel/remind/pull/1184
[...]

PR #1162, opened 2023-01-16
Enable scenario_config tests

mikapfl:
# Purpose of this PR

Thanks to @orichters we now have scenario_config consistency tests - which still fail for some scenario configs (see #1157). Once they are all fixed, we should enable the scenario config tests for all scenarios.

## Type of change

- [x] New feature 

## Checklist:


orichters (approved):
should be good to go now:
```
✔ |        32 | 01-readCheckScenarioConfig [3.4s]
```
Can't check the other tests, because the cluster is down.

mikapfl:
> Can't check the other tests, because the cluster is down.

Hah! Just run it locally! :wink: 
I did the normal tests (`make test`) and they are all green.

PR #1172, opened 2023-01-17
bugfix: overwrite negative values in f04_IO_input/output

robertpietzcker:
# Purpose of this PR
bugfix: the new IEA data contains negative entries in f04_IO_input/output for some regions, which causes the model to fail. 
This PR overwrites them with 0 as a short-term solution to get the model running. In the midterm, the input data/mapping needs to be checked why these (small) negative values occur.

## Type of change

- [x] Bug fix 

## Checklist:


modules/04_PE_FE_parameters/iea2014/datainput.gms
```diff
@@ -13,6 +13,27 @@ $include "./modules/04_PE_FE_parameters/iea2014/input/f04_IO_input.cs4r"
...
+  put "te", @15, "regi", @20, "value"//;
+  loop(t,
+    loop(regi,
+      loop(pe2se(entyPe,entySe,te),
+        if( f04_IO_input(t,regi,entyPe,entySe,te)  lt 0,
+          put_utility "msg" / "**""** input data problem: f04_IO_input has negative values that are overwritten to still allow model solving." ;
+          put_utility "msg" / "Check input data. More details in the file diagnosis_f04_input_negatives";
+          put te.tl, @ 15, regi.tl, @20, f04_IO_input(t,regi,entyPe,entySe,te):10:8 /;
+        );
+      );
```
0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q, line 30: You want the message to the .lst file in the loop?  How many negative values are there?

robertpietzcker: no, right, I'll also take it outside. only the writing to the put file needs to be in the loop. 
Thanks!

(and how do you get notified of draft pull requests for which I  haven't even assigned a reviewer? :-)

mikapfl: > and how do you get notified of draft pull requests for which I haven't even assigned a reviewer? 

Change your notification settings in the top right corner:

image

And enjoy the spam :slightly_smiling_face:

Renato-Rodrigues (approved):
It looks fine to me after the changes you added to report only once the error.
Considered it approved as long as you set a test run and it worked fine.

If you could, please create or update an issue in the remind issues list to track the need to fix this in the input data side instead, and/or contact IEA about the data input problem.

robertpietzcker:
> Considered it approved as long as you set a test run and it worked fine.

well - it solved the FRA problem. 
There are other infeasibilities in other regions, but regions that don't have the "negative input values" problem. 
/p/tmp/robertp/REMIND_3px/3p1_2023-01-13_debug_FRA_Renato/testVersion/output/xx_R3_DIAG-Base_2023-01-17_16.59.12

so even though the run still has infeasibilities, I merge now because the specific issue is (short-term) solved with it. 

I will add an issue tomorrow

modules/04_PE_FE_parameters/iea2014/datainput.gms
0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q: Stasi-Mode.

PR #1175, opened 2023-01-18
remove feature to force start scripts to use old run as path_gdx…

orichters:
## Purpose of this PR
- this way, people can name their scenario `____asdf___` if they like

## Type of change

- [x] Remove feature 
- [x] Tutorial was updated

## Checklist:


Renato-Rodrigues (approved):
I don't control much of this code part, but every change you did looks reasonable. 
Thk you for the very quick reaction to the message by the way!

orichters:
Basically, the `_?` meant: You can have an additional `_` after the string in the cell and before the datetime stuff. Now, you must have this `_`, which means using `Base_` in the cell will not find `Base` runs anymore.

Renato-Rodrigues:
Couldn't you simply escape any special regex characters from the title before using then in this line to avoid this behavior?
That way the _ should be considered.

orichters:
> Couldn't you simply escape any special regex characters from the title before using then in this line to avoid this behavior? That way the _ should be considered.

I don't understand this comment, to be honest. The `_` is there because the folder name `Base_2022-06-23_18.53.19` always has an additional `_` between the scenario title and the date.

PR #1177, opened 2023-01-18
Delete realization 15_climate/box and cm_emiscen = 2, 3, 5, 8

mikapfl:
# Purpose of this PR

The realization `box` of the module `15_climate` is broken and can't be used any more. Also, we think that if someone comes along who needs a simple climate model in REMIND and is willing to commit resources to that, they should not repair the existing box model, instead they should implement something else (like FaIR) from scratch in a new realization. Therefore, there is no value in keeping the existing box model around anymore (not usable now, won't be repaired even if someone needs an SCM in REMIND).

## Type of change

- [x] Refactoring

## Checklist:


mikapfl:
We have a somewhat deeper problem if we stop pretending that we have an endogenous climate model: The emissions scenarios which need an endogenous climate model definitely don't make any sense any more. In particular, that's 
* `cm_emiscen = 2`: temperature cap
* `cm_emiscen = 5`: forcing target from 2010 (not to exceed)
* `cm_emiscen = 8`: forcing target from 2100 onwards (overshoot scen)
* maybe `cm_emiscen = 3`: CO2 concentration cap – but that one is dead already anyway

Should we delete the code for them as well? I guess they are very untested because for a couple of years now, the `box` model doesn't work, and those emission scenarios need it, so those emission scenarios can't be run for a couple of years now.

Opinions? Insights?

@LaviniaBaumstark  @fpiontek

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
> Therefore, there is no value in keeping the existing box model around anymore

Just nobody tell that to Elmar :p

LaviniaBaumstark:
after @fbenke-pik has checked and/or adjusted the reporting, I am looking forward to this merge!

core/datainput.gms
```diff
@@ -108,9 +108,6 @@ Execute_Loadpoint 'input'      vm_cesIO, vm_invMacro;
 pm_gdp_gdx(ttot,regi)    = vm_cesIO.l(ttot,regi,"inco");
 p_inv_gdx(ttot,regi)     = vm_invMacro.l(ttot,regi,"kap");
 
-*** permit price initilization
-pm_pricePerm(ttot) = 0;
```
LaviniaBaumstark, line 112: this was only used in box?

LaviniaBaumstark (approved):

core/datainput.gms
mikapfl: yes.

fbenke-pik:
I could only find one instance of a deleted parameter used in remind2 `pm_ttot_val`. I removed it in [this PR](https://github.com/pik-piam/remind2/pull/396)

PR #1190, opened 2023-01-23
add use of rs2 without any options

robertpietzcker:
# Purpose of this PR

Add info to the tutorial


orichters (approved):
good point, Robert.

PR #1191, opened 2023-01-24
fail on whitespaces in path_gdx entries, restrict scenario titles to A-z0-9_-

orichters:
## Purpose of this PR
- avoid that files are not found because you have some spurious ` ` behind the file name/run name.
- restrict scenario titles to letters, digits, `_` and `-`
- also make a backup of the the full.lst file while restarting as desired by Robert here: https://mattermost.pik-potsdam.de/rd3/pl/pepp7th6obfjugjhj69ybe6sby

## Type of change

- [x] Bug fix 

## Checklist:

### DIFF
--- a/scripts/start/readCheckScenarioConfig.R
+++ b/scripts/start/readCheckScenarioConfig.R
@@ -27,18 +27,27 @@ readCheckScenarioConfig <- function(filename, remindPath = ".", testmode = FALSE
             paste0(rownames(scenConf)[regionname], collapse = ", "),
             " – Titles with three capital letters or 'glob' may be confused with region names by magclass. Stopping now.")
   }
-  containsdot <- grep("\\.", rownames(scenConf))
+  containsdot <- grepl("\\.", rownames(scenConf))
   if (any(containsdot)) {
     warning("These titles contain a dot: ",
             paste0(rownames(scenConf)[containsdot], collapse = ", "),
             " – GAMS would not tolerate this, and quit working at a point where you least expect it. Stopping now.")
   }
+  whitespaceerrors <- 0
+  for (path_gdx in intersect(names(path_gdx_list), names(scenConf))) {
+    haswhitespace <- grepl(" ", scenConf[[path_gdx]])
+    if (any(haswhitespace)) {
+      warning("The ", path_gdx, " cells of these runs contain whitespaces: ", paste0(rownames(scenConf)[haswhitespace], collapse = ", "),
+              " – scripts will fail to find corresponding runs and gdx files. Stopping now.")
+      whitespaceerrors <- whitespaceerrors + sum(haswhitespace)
+    }
+  }
+  errorsfound <- sum(toolong) + sum(regionname) + sum(containsdot) + whitespaceerrors
   if ("path_gdx_ref" %in% names(scenConf) && ! "path_gdx_refpolicycost" %in% names(scenConf)) {
     scenConf$path_gdx_refpolicycost <- scenConf$path_gdx_ref
     message("In ", filename,
         ", no column path_gdx_refpolicycost for policy cost comparison found, using path_gdx_ref instead.")
   }
-  errorsfound <- sum(toolong) + sum(regionname) + sum(containsdot)
   scenConf[, names(path_gdx_list)[! names(path_gdx_list) %in% names(scenConf)]] <- NA
   knownColumnNames <- c(names(cfg$gms), names(path_gdx_list), "start", "output", "description", "model",
                         "regionmapping", "extramappings_historic", "inputRevision", "slurmConfig",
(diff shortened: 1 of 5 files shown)


mikapfl (commented):
I think this is an improvement, but I agree with @0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q that it would be even better to just stop if there is whitespace. Otherwise, we have to make sure that in every place where we deal with scenario configs and names, we use the same "canonicalization". Sounds like a lot of coordination or chasing "oh, in the script to compare scenario configs, we don't remove white space in names, then the diffs are unwieldy, let's add it here" cases.

orichters:
Seems plausible. It was a bit tricky, but I implemented it such that it also tells you where the whitespace is found. If you want to change the regex to something more general, more specific or whatever, let me know.

scripts/start/readCheckScenarioConfig.R
```diff
@@ -27,18 +27,27 @@ readCheckScenarioConfig <- function(filename, remindPath = ".", testmode = FALSE
...
-  containsdot <- grep("\\.", rownames(scenConf))
+  containsdot <- grepl("\\.", rownames(scenConf))
   if (any(containsdot)) {
     warning("These titles contain a dot: ",
             paste0(rownames(scenConf)[containsdot], collapse = ", "),
             " – GAMS would not tolerate this, and quit working at a point where you least expect it. Stopping now.")
   }
+  whitespaceerrors <- 0
+  for (path_gdx in intersect(names(path_gdx_list), names(scenConf))) {
+    haswhitespace <- grepl(" ", scenConf[[path_gdx]])
```
mikapfl, line 38:
```suggestion
    haswhitespace <- grepl("\\s", scenConf[[path_gdx]])
```
Will also match on [protected spaces](https://en.wikipedia.org/wiki/Non-breaking_space) and other stuff people can come up with to break remind.

mikapfl (approved):

scripts/start/readCheckScenarioConfig.R
```diff
@@ -27,18 +27,27 @@ readCheckScenarioConfig <- function(filename, remindPath = ".", testmode = FALSE
...
+  whitespaceerrors <- 0
+  for (path_gdx in intersect(names(path_gdx_list), names(scenConf))) {
+    haswhitespace <- grepl("\\s", scenConf[[path_gdx]])
+    if (any(haswhitespace)) {
+      warning("The ", path_gdx, " cells of these runs contain whitespaces: ", paste0(rownames(scenConf)[haswhitespace], collapse = ", "),
+              " – scripts will fail to find corresponding runs and gdx files. Stopping now.")
+      whitespaceerrors <- whitespaceerrors + sum(haswhitespace)
+    }
+  }
+  errorsfound <- sum(toolong) + sum(regionname) + sum(containsdot) + whitespaceerrors
```
0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q, line 45: No half measures.
```
path_name_errors <- 0
for (path_gdx in intersect(names(path_gdx_list), names(scenConf))) {
  has_illeagal_chars <- grepl("[^[:alnum:]_\\.-]", scenConf[[path_gdx]])
  if (any(has_illeagal_chars)) {
    warning("The ", path_gdx, " cells of these runs contain illeagal characters: ",
            paste0(rownames(scenConf)[has_illeagal_chars], collapse = ", "),
            " – scripts will fail to find corresponding runs and gdx files. Stopping now.")
    path_name_errors <- path_name_errors + sum(has_illeagal_chars)
  }
}
errorsfound <- sum(toolong) + sum(regionname) + sum(containsdot) + path_name_errors
```
Allow characters, numbers, "-", "_", and ".".  I have never seen anything else being used.

orichters: I mean, at least `/` should be allowed, and for Windows users `\` or so, I guess :)

orichters: And I would rather treat whitespaces independently, because they are quite difficult to recognize, at while editing csv files with Excel/OpenOffice. And we should at least tell the users which characters are legal :)

orichters: And I wonder whether it is necessary. If the file `@asdf.gdx` exists and the scripts finds it, what is the problem? If it doesn't find it, it will fail anyway.

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q: > I mean, at least / should be allowed, and for Windows users \ or so, I guess :)

True.  I was thinking of scenario names.

scripts/start/readCheckScenarioConfig.R
```diff
@@ -27,18 +27,27 @@ readCheckScenarioConfig <- function(filename, remindPath = ".", testmode = FALSE
...
-  containsdot <- grep("\\.", rownames(scenConf))
-  if (any(containsdot)) {
-    warning("These titles contain a dot: ",
-            paste0(rownames(scenConf)[containsdot], collapse = ", "),
-            " – GAMS would not tolerate this, and quit working at a point where you least expect it. Stopping now.")
+  illegalchars <- grepl("[^[:alnum:]_-]", rownames(scenConf))
+  if (any(illegalchars)) {
+    warning("These titles contain illegal characters: ",
+            paste0(rownames(scenConf)[illegalchars], collapse = ", "),
+            " – Please use only [:alnum:]_- to avoid errors. Stopping now.")
```
0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q, line 34: You and I know what `[:alnum:]` is, the median REMIND user does not …

orichters: Better: `A-z0-9_-`?

mikapfl (approved):

PR #1192, opened 2023-01-24
re-add renv scripts with notes

pfuehrlich-pik:


orichters (approved):
Thanks, that is very helpful.

PR #1195, opened 2023-01-25
improve error text for bio switches

orichters:
## Purpose of this PR

- improve the message, because renaming alone is not what you should do… (I did :/)


merfort (approved):
Ah, I am very sorry! Yes, simple renaming was not enough. Thanks for correcting the message.

PR #1198, opened 2023-01-25
Add support for Python virtual environments

mikapfl:
# Purpose of this PR

To prepare for coupling REMIND with models based on Python (e.g., magicc, pyPSA), add support for Python virtual environments to REMIND.
The handling of the virtual environments is similar to the renv handling, but Python is entirely optional. If no Python installation is discovered, no virtual environment is created, and using Python from REMIND only works after setting `cfg$pythonEnabled` to "on" (default is "off").

## Type of change

- [x] New feature 

## Checklist:


.gitignore
```diff
@@ -71,3 +71,8 @@ plotRuntimeDependencies/
 
 # ignore RStudio project files
 /*.Rproj
+
+# Main Python virtual environment
+.venv/
+# venv.lock (renv.lock equivalent for Python)
+venv.lock
```
mikapfl, line 78:
```suggestion
```
Doesn't exist any more.

gabriel-abrahao:
Thanks for this Mika! It's been quite hard to test this on my local machine due to some crazy clashes in libraries between preexisting python, R and conda things, but I believe that's to be expected. I did manage to get it working eventually though.

I'm still trying a bunch of things, but one that I believe is relevant here is how it should handle new packages in `requirements.txt`, which is not very clear to me. A quick test adding `climate-assessment` there (which we will need eventually need) and running `start.R` showed the usual `pip` warning message, indicating that it did understand that python should be used, but did not attempt to attempt to install the dependencies. Manually running `piamenv::updatePythonVirtualEnv()` did attempt it though. 

Is that the intended behaviour? I understand that, if we need some python dependency, it should be installed into the `venv` automatically when starting the model just like the `renv` ones are, no?

mikapfl:
Dear Gabriel,

Thanks for the test! Did you use `pythonEnabled <- "on"` in your test? Without it, it won't try to install new dependencies from requirements.txt.

Cheers,

Mika

gabriel-abrahao:
Hi Mika, yes, I have `pythonEnabled <- "on"`. Tested again with an additional package and it actually worked, so maybe it was just a freak first time bug?

mikapfl:
I couldn't reproduce the error - but during reproduction I noticed that it wasn't possible to specify `pythonEnabled` in a scenario config, which clearly is something we want to be able to do. I fixed that, now you can have scenarios with python without changing default.cfg.

gabriel-abrahao:
I couldn't reproduce it either, sorry! Tried now with a fresh copy and the installation of packages from `requirements.txt` seems to be working fine with setting `pythonEnabled` just on a `scenario_config.csv`. Also did some tests with `reticulate` and seems like the environments and python versions are all being set as expected. Again, thanks for doing this!

gabriel-abrahao (approved):

tutorials/01_GettingREMIND.md
```diff
@@ -15,6 +15,7 @@ REQUIREMENTS
 - Windows only: RTools
 - LaTeX
 - pandoc
+- for couplings to other models: Python >= 3.7
```
LaviniaBaumstark, line 18: only relevant for coupling with models which need Python, maybe specify here?

LaviniaBaumstark (approved):

tutorials/01_GettingREMIND.md
LaviniaBaumstark: I think this describes current situation better

LaviniaBaumstark (approved):

tutorials/01_GettingREMIND.md
gabriel-abrahao: If it helps to have an "initial egg", one of the first Python things we will likely do is integrate the [climate-assessment](https://github.com/iiasa/climate-assessment) toolchain. It requires just Python 3, works with 3.7 and recommends 3.9 (which is what they test on). So I think requiring >=3.7 is quite reasonable for now.

PR #1205, opened 2023-01-27
make the q32_flexAdj equation more error-proof

robertpietzcker:
## Purpose of this PR

Prevent the use of faulty marginals from a previous run to calculate vm_flexAdj. 

pm_SEPrice is only used for the calculation of vm_flexAdj as long as the pm_SEPrice is below 230$/MWh (equal to 2T$/TWa), otherwise 230$/MWh is used as maximum value.  In the real world, having electricity prices > such a value for a 5 year period seems hardly imaginable.  (the energy crisis in the EU forced power prices up to 200-400 €/MWh for a year, but then they dropped down to below 200€/MWh)

## Type of change

- [x] Bug fix 

## Checklist:

### DIFF
--- a/modules/32_power/IntC/equations.gms
+++ b/modules/32_power/IntC/equations.gms
@@ -278,9 +278,12 @@ q32_flexPriceBalance(t,regi)$(cm_FlexTaxFeedback eq 1)..
 *** while inflexible technologies are penalized (v32_flexPriceShare > 1).  
 *** Flexibility tax is switched only if cm_flex_tax = 1 and is active from 2025 onwards. 
 q32_flexAdj(t,regi,te)$(teFlexTax(te))..
-	vm_flexAdj(t,regi,te) 
-	=e=
-	(1-v32_flexPriceShare(t,regi,te)) * pm_SEPrice(t,regi,"seel")$((cm_flex_tax eq 1) AND (t.val ge 2025) and (pm_SEPrice(t,regi,"seel") gt 0))
+  vm_flexAdj(t,regi,te) 
+  =e=
+  (1-v32_flexPriceShare(t,regi,te)) 
+  * ( pm_SEPrice(t,regi,"seel")$((cm_flex_tax eq 1) AND (t.val ge 2025) and (pm_SEPrice(t,regi,"seel") gt 0) and (pm_SEPrice(t,regi,"seel") lt 2))
+      + 2$((cm_flex_tax eq 1) AND (t.val ge 2025) and (pm_SEPrice(t,regi,"seel") gt 0) and (pm_SEPrice(t,regi,"seel") ge 2)) !! limit maximum applied electricity price to 230$/MWh = 2T$/TWa 
+    )
 ;
 
 *** EOF ./modules/32_power/IntC/equations.gms


Renato-Rodrigues (approved):
I am second-guessing our choice to have this as an equation condition instead of a additional price parameter now. :)

robertpietzcker:
> I am second-guessing our choice to have this as an equation condition instead of a additional price parameter now. :)

I had exactly the same thought :-)

modules/32_power/IntC/equations.gms
```diff
@@ -278,9 +278,12 @@ q32_flexPriceBalance(t,regi)$(cm_FlexTaxFeedback eq 1)..
...
 q32_flexAdj(t,regi,te)$(teFlexTax(te))..
-	vm_flexAdj(t,regi,te) 
-	=e=
-	(1-v32_flexPriceShare(t,regi,te)) * pm_SEPrice(t,regi,"seel")$((cm_flex_tax eq 1) AND (t.val ge 2025) and (pm_SEPrice(t,regi,"seel") gt 0))
+  vm_flexAdj(t,regi,te) 
+  =e=
+  (1-v32_flexPriceShare(t,regi,te)) 
+  * ( pm_SEPrice(t,regi,"seel")$((cm_flex_tax eq 1) AND (t.val ge 2025) and (pm_SEPrice(t,regi,"seel") gt 0) and (pm_SEPrice(t,regi,"seel") lt 2))
+      + 2$((cm_flex_tax eq 1) AND (t.val ge 2025) and (pm_SEPrice(t,regi,"seel") gt 0) and (pm_SEPrice(t,regi,"seel") ge 2)) !! limit maximum applied electricity price to 230$/MWh = 2T$/TWa 
+    )
```
0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q, line 286: Why not simply
```
q32_flexAdj(t,regi,te)$(teFlexTax(te))..
  vm_flexAdj(t,regi,te) 
  =e=
    (1 - v32_flexPriceShare(t,regi,te))
  * max(0, min(2, pm_SEprtive(t,regi,"seel")))$(    cm_flex_tax eq 1
                                                AND t.val ge 2025    )
;
```?

File: modules/32_power/IntC/equations.gms [hunk 2]
```diff
@@ -278,9 +278,12 @@ q32_flexPriceBalance(t,regi)$(cm_FlexTaxFeedback eq 1)..
 *** while inflexible technologies are penalized (v32_flexPriceShare > 1).  
 *** Flexibility tax is switched only if cm_flex_tax = 1 and is active from 2025 onwards.
```
0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q, line 279: Good thing you deleted the
> - [ ] I checked that the [in-code documentation](https://github.com/remindmodel/remind/blob/develop/main.gms#L120) is up-to-date

part from the template.

Still better would have been to explain why this is limited to 2, and how that relates to the 200 $/TWa in the comment.

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q (commented):
Might be worth a follow-up commit …

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
@cchrisgong, you might want to carry this over to https://github.com/remindmodel/remind/blob/064cf53fa412df0a5d9a9977aa06448bb091955c/modules/32_power/DTcoup/equations.gms#L225-L236

modules/32_power/IntC/equations.gms
0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q: Oh, I see that was part of the long line nobody is going to read.  Still, comment up top with the rest would be better.

modules/32_power/IntC/equations.gms
robertpietzcker: I thought you could only use max/min-commands in data preparation, but not in equations when using NLP and not DNLP?

but if it works, your formulation definitely is shorter

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q: You can't use them with _variables_.  At the time of model generation, all the parameters are known to GAMS and collapsed together with all the numbers in the equation before being send off to the solver.

PR #1213, opened 2023-02-10
Ensure library requirements are installed before running tests; add make target

mikapfl:
## Purpose of this PR

If a new requirement (e.g. new library or newer version) is added to develop, and then someone else pulls develop and runs `make test`, it just hangs because remind asks about installing the new library and the tests hide the question. Most recently, this happened with `reticulate`.

With this PR, we now make sure that the REMIND requirements are met before running tests.

## Type of change

- [x] Bug fix 
- [x] Refactoring

## Checklist:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,6 +26,12 @@ update-renv-all: ## Upgrade all packages (including CRAN packages) in your renv
 	[ -e ".venv/bin/python" ] && .venv/bin/python -mpip install --upgrade pip wheel
 	[ -e ".venv/bin/python" ] && .venv/bin/python -mpip install --upgrade --upgrade-strategy eager -r requirements.txt
 
+ensure-reqs:     ## Ensure the REMIND library requirements are fulfilled
+                 ## by installing updates and new libraries as necessary. Does not
+                 ## install updates unless it is required.
+	@Rscript -e 'source("scripts/start/ensureRequirementsInstalled.R"); ensureRequirementsInstalled(rerunPrompt="make ensure-reqs")'
+	@[ -e ".venv/bin/python" ] && .venv/bin/python -mpip -qq install -r requirements.txt
+
 archive-renv:    ## Write renv.lock into archive.
 	Rscript -e 'piamenv::archiveRenv()'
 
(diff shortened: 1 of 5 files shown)


orichters (approved):
What I did for testing it:
- uninstall `piamenv` package -> reasonable error message
- install old version of `piamenv` -> reasonable error message
- uninstall `remind2` -> `make test` manages to fix that
- install old version of `remind2` -> `make test` manages to fix that

I'm not sure whether it makes sense that I can say `no` to the question by `make test` whether I want to update but it proceeds with the tests – and it seems this is not how it is intended in `setup_updateRenv.R`.
```
  remind2 >= 1.104.1 is required, but 1.104.0 is installed - please update: remind2
Try to fix automatically? (Y/n): n
make[1]: Leaving directory '/p/tmp/oliverr/remind-mikapfl'
✔ | F W S  OK | Context
✔ |         3 | 01-combine_slurmConfig
✔ |     1   1 | 01-gms_readSettings [0.1s]
…
```

mikapfl:
> I'm not sure whether it makes sense that I can say no to the question by make test whether I want to update but it proceeds with the tests – and it seems this is not how it is intended in setup_updateRenv.R.

True, that is not really useful, but it is the same when you try to directly start remind, no? It asks you if it should fix it, and if you say no, it just carries on.

mikapfl:
Actually, my intention was that tests don't ask, just install necessary updates automatically. Fixed.

orichters:
I'm not sure whether this is the right choice. I'm not sure about other libraries, but for example `remind2` can be a bit sensitive because you may want to have harmonized reporting. In particular, because the script is not installing only the required package version, but the newest it can find (is that intended, by the way?). I'm reluctant to updating packages automatically in case you run an innocent `make test` (and I can imagine people having strong opinions about that :)). Or, framed differently: If we accept that people run `./start.R` without updating the libraries (so there seem to be a reason we don't always update automatically), then `make test` shouldn't do it automatically.

mikapfl:
I don't feel strongly either way, so I just reverted it to the asking behaviour, but now intended. :wink:

orichters (approved):
fine from my side, big improvement compared to endless tests that do not even fail

PR #1215, opened 2023-02-15
allow to specify copyConfigFrom column in csv files 

orichters:
## Purpose of this PR
- allow to add an `copyConfigFrom` column to `scenario_config.csv` files
- If you put a scenario name into this cell, it copies all the settings from that scenario, but not for settings that are specified in the current row
- this allows to easily generate scenario variations and to improve consistency in the scenario config
- the assignment also works in a nested way, so you can have `NDC` using `Base` as `copyConfigFrom`, and `Policy1` uses `NDC` and `Policy2` uses `Policy1` etc.
- Cleanup of the NGFS template
- if you prefer a different column name or so, we can of course discuss that

## Type of change

- [x] New feature 
- [x] The tutorial was updated

## Checklist:


orichters:
@christophbertram: I made a cleanup of the NGFS config files:
- use the `copyFromConfig` stuff for both the normal scenarios and the damage runs
- delete the differentiation with `bIT` and `BIT` to get the config a bit more manageable again. With the `copyFromConfig` feature, one can very quickly rebuild such a distinction, if required in Phase V or later.
- To check the differences, run the following in an `R` session in `/p/tmp/oliverr/remind/`:
```
library(dplyr)
source("scripts/start/readCheckScenarioConfig.R")
source("scripts/start/path_gdx_list.R")
old <- readCheckScenarioConfig('config/scenario_config_NGFS_v3_old.csv')
new <- readCheckScenarioConfig('config/scenario_config_NGFS_v3.csv')
old <- old[order(row.names(old)), ]
new <- new[order(row.names(new)), ]
print(bind_cols(row.names(old), row.names(new), .name_repair = "unique"), n = 100)
for (n in names(old)) { 
  message(n)
  if (! identical(old[[n]], new[[n]])) {
    print(bind_cols(old[[n]], new[[n]], .name_repair = "unique"), n = 100)
  }
}
```

scripts/start/readCheckScenarioConfig.R
```diff
@@ -43,13 +43,41 @@ readCheckScenarioConfig <- function(filename, remindPath = ".", testmode = FALSE
...
+      if (! is.na(copyConfigFrom)) {
+        scenConf[run, is.na(scenConf[run, ])] <- scenConf[copyConfigFrom, is.na(scenConf[run, ])]
+      }
+    }
+    scenConf$copyConfigFrom <- NULL
+  }
+
+  errorsfound <- sum(toolong) + sum(regionname) + sum(illegalchars) + whitespaceerrors +
+                 length(copyFromMissing) + sum(copyFromLater)
+
```
mikapfl, line 79: Could you put this into a separate function, which takes `scenConf` as an argument and fills it out according to `copyConfigFrom` and returns it? I think that way it would be much easier to understand what is going on, and also it would be possible to write some unit tests for this function, asserting that the filling works as intended.

mikapfl (changes requested):
I have no idea what is going on in the NGFS configs :wink: 

Regarding the changes in start scripts: looks to me like a step forward, thanks. I think a separate function would be better, though.

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
Nice idea.  But a couple of points:
1. Changing code and changing configuration should be two different merge requests.
2. All for factoring this in a separate function.
3. This can get complicated soon.  A function to normalise a scenario, i.e. to get some overview what exactly would end up in a configuration, would be very beneficial.
4. Currently, it does not seem to work.  With 

   scenario | start | copyFromConfig | cm_CCS_cement | cm_CCS_chemicals | cm_CCS_steel
   -- | -- | -- | -- | -- | --
   foo | 0 |   | 23 | 42 | 3.14
   bar | 0 | foo |   | 13 | 2.718
   bazz | 1 | bar | -1 |   |  

   I get 
   ```
   $ grep cm_CCS_ cfg.txt 
   cm_CCS_cement: -1
   cm_CCS_chemicals: '1'
   cm_CCS_steel: '1'
   ```
   so `cm_CCS_chemicals` and `cm_CCS_steel` are not carried over as advertised.
5. I trust you did consider that somebody who is very close to a project deadline and needs runs _now_ will create an infinite loop of configurations to copy from, and that it is your responsibility to detect that and give a proper error message ;)

orichters:
Answer to @0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q.

@1, @2: ok.

@3: Something like that: Probably needs a nice wrapper.
```
source("scripts/start/readCheckScenarioConfig.R")
source("scripts/start/path_gdx_list.R")
readCheckScenarioConfig("config/scenario_config.csv")["SSP2EU-Base", ]
```

@4: Sorry, it has to be `copyConfigFrom`, not `copyFromConfig`. I wrote that differently in the code and the PR description. My fault.

@5: Yes. The code checks whether the scenario you refer to is defined above in the config. If not, it spits out a `The following scenarios references in copyConfigFrom column a scenario defined only later` warning.

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
:+1:

orichters:
- I moved it to an own function `copyConfigFrom`
- I added tests, not for the individual function, but rather expanding the config to be tested. For me, this seemed safer, given you also want to check whether the function is called correctly etc. I don't see the need for further tests at the moment
- I first left the NGFS stuff inside, because there I basically implement the `copyConfigFrom` column, so you can also see how it looks like.

scripts/start/readCheckScenarioConfig.R
```diff
@@ -131,3 +116,25 @@ readCheckScenarioConfig <- function(filename, remindPath = ".", testmode = FALSE
   if (errorsfound > 0) if (testmode) warning(errorsfound, " errors found.") else stop(errorsfound, " errors found, see explanation in warnings.")
   return(scenConf)
 }
+
+copyConfigFrom <- function(scenConf) {
```
mikapfl, line 120: Our [coding etiquette](https://github.com/remindmodel/remind/blob/develop/tutorials/07a_Advanced_ChangeStartingScripts.md#coding-etiquette-structuring) says:
> one file per function (you can have multiple functions in one file, but only the topmost should be used from other files), file name is the same as function name

So, I think you should move copyConfigFrom into `copyConfigFrom.R`.

mikapfl (changes requested):

scripts/start/readCheckScenarioConfig.R
orichters: Why is "you can have multiple functions in one file, but only the topmost should be used from other files" not satisfied?

mikapfl: It is, I'm just confused. Thanks for double-checking.

mikapfl (approved):
https://github.com/orichters/remind/pull/6 and then go!

LaviniaBaumstark:
nice, thanks. What happens in coupled runs? It is working for REMIND configuration and does not interrupt and other processes, right?

orichters:
This PR only affects the process of reading and processing the scenario config file. If `copyFromConfig` exists in this file, the configs from the "parent" scenario are copied to the empty cells of the "child" scenario. Then, the completed config is returned as before.

Regarding coupled configs: Well, you can use that feature also in the `scenario_config_coupled.csv` if you like, but that was maybe not what you wanted to knowß

LaviniaBaumstark:
yes and no, nice that it can be used in couples configs. I also wanted to know if you tested it for coupling and if it might cause some problems there

orichters:
I did not run actual coupled runs. But: My PR only changes `readCheckScenarioConfig()`. And you can easily check that the old and the new implementation yields exactly the same return value for this function for coupled scripts, which is no surprise given that if `copyConfigFrom` does not exist in these files.

```
source("/p/tmp/oliverr/remind-develop/scripts/start/path_gdx_list.R")
source("/p/tmp/oliverr/remind-develop/scripts/start/readCheckScenarioConfig.R")
oldcoupled <- readCheckScenarioConfig("/p/tmp/oliverr/remind-develop/config/scenario_config_coupled_NGFS_v3.csv")
old <- readCheckScenarioConfig("/p/tmp/oliverr/remind-develop/config/scenario_config_NGFS_v3.csv")
source("/p/tmp/oliverr/remind/scripts/start/readCheckScenarioConfig.R")
newcoupled <- readCheckScenarioConfig("/p/tmp/oliverr/remind-develop/config/scenario_config_coupled_NGFS_v3.csv")
new <- readCheckScenarioConfig("/p/tmp/oliverr/remind-develop/config/scenario_config_NGFS_v3.csv")
print(identical(old, new))
print(identical(oldcoupled, newcoupled))
```

So I don't see any way this could impact coupled scripts. And once you introduce a `copyConfigFrom` column into your (coupled) config file, you have to check as I described [above](https://github.com/remindmodel/remind/pull/1215#issuecomment-1431670127) whether the settings are still correct.

LaviniaBaumstark (approved):

PR #1222, opened 2023-02-20
bugfix that baseline emission peaks in 2020 already in China

cchrisgong:
this potentially fixes issue https://github.com/remindmodel/development_issues/issues/180

## Type of change

- [X ] Bug fix 
- [ ] Fundamental change

## Checklist:

- [ X] My code follows the [coding etiquette](https://github.com/remindmodel/remind/blob/develop/main.gms#L80)
- [ X] I performed a self-review of my own code
- [ X] I explained my changes within the PR, particularly in hard-to-understand areas
- [ X] I checked that the [in-code documentation](https://github.com/remindmodel/remind/blob/develop/main.gms#L120) is up-to-date
- [ X] I adjusted the reporting in [`remind2`](https://github.com/pik-piam/remind2) where it was needed
- [ X] I adjusted `forbiddenColumnNames` in [readCheckScenarioConfig.R](https://github.com/remindmodel/remind/blob/develop/scripts/start/readCheckScenarioConfig.R) in case the PR leads to deprecated switches

* Test runs are here: 
after change:
 /p/tmp/chengong/cha_coalphaseout/remind/output/SSP2EU-AMT-Base_after_2023-02-19_15.33.34
 /p/tmp/chengong/cha_coalphaseout/remind/output/SSP1-AMT-Base_after_2023-02-19_15.33.16
before change:
 /p/tmp/chengong/cha_coalphaseout/remind/output/SSP2EU-AMT-Base_before_2023-02-19_13.36.20
 /p/tmp/chengong/cha_coalphaseout/remind/output/SSP1-AMT-Base_before_2023-02-19_13.36.00

/p/tmp/chengong/cha_coalphaseout/remind/compScen-cha_baseline_fix-2023-02-20_09.58.42-H12

### DIFF
--- a/core/datainput.gms
+++ b/core/datainput.gms
@@ -533,8 +533,12 @@ $ENDIF.tech_earlyreti
 *SB* Time-dependent early retirement rates in Baseline scenarios
 $ifthen.Base_Cprice %carbonprice% == "none"
 $ifthen.Base_techpol %techpol% == "none"
-*** Allow no early retirement in future periods under baseline for China; disallowing early retirement of coal all regions runs into EUR infesibility under H12
-pm_regiEarlyRetiRate(t,"CHA","pc")$(t.val ge 2020) = 0;
+*** CG: Allow no early retirement in future periods under baseline for developing countries; disallowing early retirement of coal all regions runs into EUR infesibility under H12
+loop(regi,
+if ( p_developmentState("2015",regi) < 1,
+pm_regiEarlyRetiRate(t,regi,"pc")= 0;
+);
+);
 $endif.Base_techpol
 $endif.Base_Cprice
 
--- a/core/datainput.gms
+++ b/core/datainput.gms
@@ -533,8 +533,8 @@ $ENDIF.tech_earlyreti
 *SB* Time-dependent early retirement rates in Baseline scenarios
 $ifthen.Base_Cprice %carbonprice% == "none"
 $ifthen.Base_techpol %techpol% == "none"
-*** Allow very little early retirement future periods
-pm_regiEarlyRetiRate(t,regi,"pc")$(t.val gt 2025) = 0.01;
+*** Allow no early retirement in future periods under baseline
+pm_regiEarlyRetiRate(t,"CHA","pc")$(t.val ge 2020) = 0;
 $endif.Base_techpol
 $endif.Base_Cprice
 
--- a/core/datainput.gms
+++ b/core/datainput.gms
@@ -533,7 +533,7 @@ $ENDIF.tech_earlyreti
 *SB* Time-dependent early retirement rates in Baseline scenarios
 $ifthen.Base_Cprice %carbonprice% == "none"
 $ifthen.Base_techpol %techpol% == "none"
-*** CG: Allow no early retirement in future periods under baseline for developing countries; disallowing early retirement of coal all regions runs into EUR infesibility under H12
+*** CG: Allow no early retirement in future periods under baseline for developing countries
 loop(regi,
 if ( p_developmentState("2015",regi) < 1,
 pm_regiEarlyRetiRate(t,regi,"pc")= 0;
--- a/core/datainput.gms
+++ b/core/datainput.gms
@@ -533,7 +533,7 @@ $ENDIF.tech_earlyreti
 *SB* Time-dependent early retirement rates in Baseline scenarios
 $ifthen.Base_Cprice %carbonprice% == "none"
 $ifthen.Base_techpol %techpol% == "none"
-*** Allow no early retirement in future periods under baseline
+*** Allow no early retirement in future periods under baseline for China; disallowing early retirement of coal all regions runs into EUR infesibility under H12
 pm_regiEarlyRetiRate(t,"CHA","pc")$(t.val ge 2020) = 0;
 $endif.Base_techpol
 $endif.Base_Cprice


cchrisgong:
@robertpietzcker 
red: after
blue: before

SE|Electricity|Coal for CHA still drops too suddenly:
image
Cap|Electricity|Coal for CHA looks smoother though:
image

cchrisgong:
total emission goes up a bit:
image

cchrisgong:
electricity supply emission is less sharply decreasing, but still somewhat suddenly decreasing (it may be due to the fact that capacity factor is hard coded to goes down and no capacity can be added under baseline (due to the way REMIND works)? 
image
image

We could force add capacity(?) - I don't know how to do this, or hard code cf to be flat (=57.8%) under baseline, which could ensure a smoother transition

also seeking @christophbertram 's opinion

cchrisgong:
btw I'm not sure how to not hard code region in this, @LaviniaBaumstark. It is not an input data per se, but just a bound for a single region. Including all regions cause infes for EUR, so either way some regions have to be hard-coded.

robertpietzcker:
> electricity supply emission is less sharply decreasing, but still somewhat suddenly decreasing (it may be due to the fact that capacity factor is hard coded to goes down

@stephenbi I saw that you added this reduction of CF down to 0.5 in mrremind. 
What was the motivation for this? larger VRE shares, pushing out coal dispatch?

It looks like there are lots of calculations on coal power plant CF - would it be possible to simply shift this CF reduction back by 5 years or so, to slow down this coal phaseout today? or would you argue that this dispatch phaseout is happening in most regions of the world already so strongly?

robertpietzcker:
> btw I'm not sure how to not hard code region in this, @LaviniaBaumstark. It is not an input data per se, but just a bound for a single region. Including all regions cause infes for EUR, so either way some regions have to be hard-coded.

I would propose you can loop over regi and use p_developmentState:
loop(regi,
  if ( p_developmentState("2015",regi) < 1, 
    pm_regiEarlyRetiRate(t,regi,"pc")= 0;
  );
);

I would likely limit ER in all timesteps, not just gt 2020?

(I did the checking of  p_developmentState for 2015, as CHA reaches 1 (= fully developed) in 2020 already ;-)

stephenbi:
We decided (with @Renato-Rodrigues @nicobauer @christophbertram ) on the 50% capacity factor assumption 3-4 years ago. I'd say it seems to hold up pretty well against reality. China's capacity factor has actually been <=50% since 2015 (see Figure below), so we already slightly overestimate its historical CF. 

image

One thing we could to is set a lower bound on near-term capacity. I implemented 2025 coal capacity scenarios in mrremind for my PPCA paper by varying assumptions on completion of the project pipeline (data from the Global Coal Plant Tracker). For example, the 'neutral' scenario (all countries complete projects at historical rates) might make sense for Base and NPi. For NDC, 'green' might be more appropriate (assumes all countries cut their historical completion rates in half). Note that the figure's outdated - new data was released a couple weeks ago. Does that sound reasonable? 

image
image

cchrisgong:
@robertpietzcker yes, I can limit it for all time steps. And yes, this bound is only applicable under baseline. Thanks for the suggestion for using the set - let me change that

core/datainput.gms
```diff
@@ -533,8 +533,12 @@ $ENDIF.tech_earlyreti
 *SB* Time-dependent early retirement rates in Baseline scenarios
 $ifthen.Base_Cprice %carbonprice% == "none"
 $ifthen.Base_techpol %techpol% == "none"
-*** Allow no early retirement in future periods under baseline for China; disallowing early retirement of coal all regions runs into EUR infesibility under H12
-pm_regiEarlyRetiRate(t,"CHA","pc")$(t.val ge 2020) = 0;
+*** CG: Allow no early retirement in future periods under baseline for developing countries; disallowing early retirement of coal all regions runs into EUR infesibility under H12
```
LaviniaBaumstark, line 536: thanks for this general formulation. I would delete the 2. sentence as this only hold now and might confuse in future

LaviniaBaumstark (approved):

robertpietzcker:
> We decided (with @Renato-Rodrigues @nicobauer @christophbertram ) on the 50% capacity factor assumption 3-4 years ago. I'd say it seems to hold up pretty well against reality. China's capacity factor has actually been <=50% since 2015 (

Thanks for the info, Steve!
Then I would say it would be good to pull this reduction forward, so that we already see it reaching ~0.5 in 2020 - currently in CHA, CF goes up from 0.52 in 2015 to 0.58 in 2020 in REMIND. and then drops down to 0.54 in 2025, creating this strong peak in 2020 in REMIND that seems to be unrealistic.

I don't know where the "historic" CFs in REMIND come from, but they seem a bit different from the EMBER values that you posted, Steve - would it be possible to consolidate them. or at least ensure we don't get an increase from 2015-2020 and lower 2020 values?

@Renato-Rodrigues   @stephenbi

stephenbi:
From what I gather, our CF data is mostly ported from REMIND_11_regi and thus not historical after 2010(?). Then we have some 2015 IRENA data for low carbon techs. The EMBER data looks to be reasonably complete (215 countries, 2000-2021), so should we maybe just fully transition to it? 

And just for clarification, @robertpietzcker are you suggesting that we move the convergence to 0.5 forward or leave that as is?

In any case, since the 2020 peak persists even with early retirement disallowed, should I implement the aforementioned lower bounds on 2025 capacity? 

And since 2020 coal power is pretty low compared to historical data in China and several other regions, should we increase the current historical lower bound of 0.9 * p_PE_histCap? And should we do it for all te?

robertpietzcker:
> From what I gather, our CF data is mostly ported from REMIND_11_regi and thus not historical after 2010(?).

@Renato-Rodrigues do you remember where these region-differentiated pc capacity factors after 2010 come from? 

I remember we had added concrete assumptions for EU and JPN, but I don't understand where the current changes in CF from 2010-2020 eg in IND and CHA etc come from. 

from my current (incomplete) understanding, I would propose using the EMBER values (averaged 2013-2017 for 2015; averaged 2018-2021 for 2020) to set the values in REMIND - or is there something that speaks against this?

Renato-Rodrigues:
The old capacity factor values from the remind 11 regions input data can be found at: `/p/projects/rd3mod/inputdata/sources/REMIND_11Regi`
`f_cf-global.csv` -> which includes the global values for capacity factors
`f_cf-rules.csv` -> which included some specific regions and years modified values that, if I am not mistaken, at the time were hard-coded at the REMIND gams code to JPN, EUR and USA.

Unfortunately, we don't have version control for these files, and my memory is not reliable enough to affirm with conviction if and when those numbers were edited. The `f_cf-rules.csv` file does not have IND specific values, and most of the changes were probably only added in the end of the file to REF and CHA pc factors.

After that, the capacity factors are also changed in the mrremind code, in the `calcCapacityFactor` function by using data from GCPT for coal. This change was added a bit more then two years ago to the input data creation: https://github.com/pik-piam/mrremind/pull/28

stephenbi:
Thanks Renato. So to me, this all speaks in favor of fully transitioning to EMBER data for all historical periods through 2020. The only question is whether there is any reason why the 2015 IRENA numbers might be preferable for RE, or if it was just a stopgap solution. Also, is there a standard in place for how to handle the year 2020 in the aggregation of data to the 2020 timestep? If not, I'd propose assigning a weight of 0.5 to 2020 and 1.5 to the other 3 years with available data.

robertpietzcker:
> So to me, this all speaks in favor of fully transitioning to EMBER data for all historical periods through 2020.

sounds good to me, Steve! Could you take a look at the CFs that would result from EMBER data?

some questions to check:
1. for which technology can we derive CFs from EMBER data?
2. does the EMBER data include CHP capacities and output?

stephenbi:
What we could do instead to directly address the issue at hand is add the data from this Ember [graphic](https://github.com/remindmodel/remind/pull/1222#issuecomment-1437364722)  to f_cf-rules.csv (or a separate file)

robertpietzcker:
> What we could do instead to directly address the issue at hand is add the data from this Ember [graphic](https://github.com/remindmodel/remind/pull/1222#issuecomment-1437364722) to f_cf-rules.csv (or a separate file)

well, the question is whether the CFs from the graphic help us if they are inconsistent with the capacity data that we would be comparing to? so if we compare to capacity data without the <30MW plants, then we should also use CFs w/o <30MW plants in order to get the right energy amounts out, right? or am I overlooking something?

still, if producing this data that you put in the table is easy and can easily be done for our REMIND regions, it might be good to implement these values for coal in REMIND. As we currently have some temporary changes in the model that seem to date back a long time, I think overwriting that would be great:


<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=Excel.Sheet>
<meta name=Generator content="Microsoft Excel 15">
<link id=Main-File rel=Main-File
href="file:///C:/Users/robertp/AppData/Local/Temp/msohtmlclip1/01/clip.htm">
<link rel=File-List
href="file:///C:/Users/robertp/AppData/Local/Temp/msohtmlclip1/01/clip_filelist.xml">
<style>

</style>
</head>

<body link="#0563C1" vlink="#954F72">



pc | CHA-REMIND | CHA-Ember | USA | IND | EUR | OAS | REF | JPN
-- | -- | -- | -- | -- | -- | -- | -- | --
2005 | 0.69 | 0.68 | 0.72 | 0.78 | 0.60 | 0.66 | 0.40 | 0.83
2010 | 0.60 | 0.59 | 0.63 | 0.74 | 0.53 | 0.65 | 0.41 | 0.70
2015 | 0.52 | 0.55 | 0.55 | 0.64 | 0.54 | 0.66 | 0.45 | 0.77
2020 | 0.58 | 0.56 | 0.60 | 0.60 | 0.54 | 0.62 | 0.50 | 0.74
2025 | 0.54 |   | 0.63 | 0.57 | 0.51 | 0.58 | 0.44 | 0.69
2030 | 0.52 |   | 0.52 | 0.53 | 0.49 | 0.54 | 0.47 | 0.60
2035 | 0.50 |   | 0.50 | 0.50 | 0.50 | 0.50 | 0.50 | 0.50



</body>

</html>

stephenbi:
> well, the question is whether the CFs from the graphic help us if they are inconsistent with the capacity data that we would be comparing to? so if we compare to capacity data without the <30MW plants, then we should also use CFs w/o <30MW plants in order to get the right energy amounts out, right? or am I overlooking something?
Good point.

> still, if producing this data that you put in the table is easy and can easily be done for our REMIND regions, it might be good to implement these values for coal in REMIND. As we currently have some temporary changes in the model that seem to date back a long time, I think overwriting that would be great:
True. Looks like this may smooth the CHA 2020 peak a bit indeed. And yes it's easy enough, but just to tie in the parallel IEA update discussion, are we going to use the new IEA generation data? Because then I'd say it makes sense to get CFs from that too (there are discrepancies with Ember on coal capacity).

cchrisgong:
Sorry for only following the conversation partially, but just a bit clarification -

1) on smoothing Chinese emission peak, as I understand, it is more about implementing coal additions in REMIND than adjusting CF values? (because it looks like the current CFs and EMBER CFs are not that far apart, therefore should not change the results that much?) Unless I missed something...
image

2) importantly our coal capacity data for CHA should be updated to EMBER data (see below) (?) As Robert discussed with me before and now Steve agreed to do this along with the inputdata update. Is this data distorted by the "lack of <30MW plant" issue, if so, do we need to increase this number?

if (cm_startyear le 2020,
vm_cap.lo("2015","CHA","pc","1")            = 0.904;  !! Ember says 903.85 GW in 2015, 1069.52 GW in 2020 
vm_cap.lo("2020","CHA","pc","1")            = 1.07;  
);

just for reference:

before adding above cap.lo bound (latest AMT):
image
after adding this bound (on my own branch):
image

robertpietzcker:
> 1. on smoothing Chinese emission peak, as I understand, it is more about implementing coal additions in REMIND than adjusting CF values?

I think it is both - a 10% increase of CFs from 2015 to 2020 in CHA in REMIND will produce problems when trying to match energy/emission values. So it would be great to fix this (Ember only goes from 0.55 to 0.56)

robertpietzcker:
> And yes it's easy enough, but just to tie in the parallel IEA update discussion, are we going to use the new IEA generation data? Because then I'd say it makes sense to get CFs from that too (there are discrepancies with Ember on coal capacity).

good point. 
I just commented in the IEA2022 update channel. if the region resolution is similar, and no source has a clear quality advantage, I guess using IEA would be more reasonable.

stephenbi:
> importantly our coal capacity data for CHA should be updated to EMBER data (see below) (?) As Robert discussed with me before and now Steve agreed to do this along with the inputdata update. Is this data distorted by the "lack of <30MW plant" issue, if so, do we need to increase this number?

I think the historical input data for capacity is fine as is, but the constraint in REMIND needs to be more stringent (currently vm_cap.lo = 0.9 * pm_PE_histCap). I'll set them equivalent instead, unless someone objects? I don't think the <30MW plants are very significant, especially since the CF data suggests that they aren't used much at all.

> on smoothing Chinese emission peak, as I understand, it is more about implementing coal additions in REMIND than adjusting CF values? (because it looks like the current CFs and EMBER CFs are not that far apart, therefore should not change the results that much?) Unless I missed something...

Yeah, forcing 2025 capacity additions will have a larger effect. Seeing as China permitted 100 GW of new coal plants this past year, I don't think a 2020 peak is very reasonable :/. Will get to this soon!

> if the region resolution is similar, and no source has a clear quality advantage, I guess using IEA would be more reasonable.

It's looking like Ember wins on spatial resolution.

robertpietzcker:
> but the constraint in REMIND needs to be more stringent (currently vm_cap.lo = 0.9 * pm_PE_histCap).

I would try to avoid setting bounds at historic levels, otherwise the model will easily come into overproduction, which renders all marginals useless. maybe go up to 0.95 * pm_PE_histCap, but not higher. 

But I think the main problem is that that until now, the model doesn't have 2020 capacity values for most regions and techs. once Chris added pc data for 2020 for CHA, the model actually went above the EMBER values (if I read her plot correctly). 
so I wouldn't tighten the bounds, but rather add the necessary data :-)

stephenbi:
> I would try to avoid setting bounds at historic levels, otherwise the model will easily come into overproduction, which renders all marginals useless. maybe go up to 0.95 * pm_PE_histCap, but not higher.

Duly noted.

> But I think the main problem is that that until now, the model doesn't have 2020 capacity values for most regions and techs. once Chris added pc data for 2020 for CHA, the model actually went above the EMBER values (if I read her plot correctly).

The model _does_ have 2020 values for pecoal capacity though. I added them a couple years ago to p_PE_histCap (from global coal plant tracker data). CHA is at 1.034, which should increase when I update the data revision. So I'm still in favor of tightening the bound.

cchrisgong:
> I would try to avoid setting bounds at historic levels, otherwise the model will easily come into overproduction, which renders all marginals useless. maybe go up to 0.95 * pm_PE_histCap, but not higher.

@robertpietzcker just for clarifications, do you mean that baseline the coal for "2020" point is above historical 2020 data (figure below)? If model 2020 means 2018-2022, and our historical is only until 2020, so it is possible that Chinese real-world data for 2021 and 2022 would be higher (?)
image
Also, our model should be allow to overshoot as well as undershoot on 2020 (since it is model results, no?) I am worried that since 1) coal is never profitable under policy runs, setting 95% bounds would mean we will still have a gap with historical data because it will for sure undershoot 2) we never look at baseline results so overshooting there should not be an issue

cchrisgong:
> > 1. on smoothing Chinese emission peak, as I understand, it is more about implementing coal additions in REMIND than adjusting CF values?
> 
> I think it is both - a 10% increase of CFs from 2015 to 2020 in CHA in REMIND will produce problems when trying to match energy/emission values. So it would be great to fix this (Ember only goes from 0.55 to 0.56)

Thanks  - I see this now!

robertpietzcker:
> just for clarifications, do you mean that baseline the coal for "2020" point is above historical 2020 data (figure below)?

no, with "overproduction" I mean the situation where bounds on capacities lead to higher production than the demand that we calibrated the model to. This will lead to zero or negative prices, which produce all sorts of negative effects in the model. 

So I would say it is a tradeoff between "getting historic values exactly right" and "keeping the model dynamics working" - and in this tradeoff, I would say getting to within 2-5% of historic values is absolutely sufficient, while creating a situation with slack/overproduction in the model is very painful for model dynamics and later analysis.

robertpietzcker:
> But then this would be only a wrong fix.. because ideally we want a correct demand and then we could also shoot for a correct supply without marginals collapsing

given the different sources we need for different parts of the model, and the differences for the same variable between different sources, I would say it is impossible to be "correct". getting within 2-5% (of large values) is totally sufficient for all things I could imagine we could want to do with REMIND

cchrisgong:
@robertpietzcker sorry I thought you commented without seeing this addition, I added it back now :p

I think this is possible, but what is also likely is that near-term model dynamics can be wrong, especially REMIND CES always does partial demand reduction as opposed to mitigation... To solve it, in my opinion would mean making near-term elasticity really stiff, so for near-term dynamics very little demand reduction is possible. Then when we do correct historical fixings, we will in theory have good marginals...

I feel like now I am bashing my head against the "demand too low" question again in REMIND XD

robertpietzcker:
> Then when we do correct historical fixings, we will in theory have good marginals...

I repeat: there is no "correct" historic fixing across sources. it is simply impossible. they never agree.
So we will always need to leave the model some wriggle room :-)

PR #1227, opened 2023-02-28
add R script for cascading existing scenario configs

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
- [x] New feature 

## Checklist:


orichters:
And I wonder whether it would make sense to have something like this in the script, to be sure.
```
source("../start/path_gdx_list.R")
source("../start/readCheckScenarioConfig.R")
old <- readCheckScenarioConfig(file)
... run the script
new <- readCheckScenarioConfig(file)
if (! identity(old, new)) {
   stop("Files are not identical, something went wrong")
}
```

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
> Now keen to see where I just got it completely wrong :)

Nowhere.  I missed some thing from my commit.  Try again.

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
> And I wonder whether it would make sense to have something like this in the script, to be sure.
> 
> ```
> source("../start/path_gdx_list.R")
> source("../start/readCheckScenarioConfig.R")
> old <- readCheckScenarioConfig(file)
> ... run the script
> new <- readCheckScenarioConfig(file)
> if (! identity(old, new)) {
>    stop("Files are not identical, something went wrong")
> }
> ```

- Jeeez, these are awful functions …
  - They do assume some working directory.  People should have their wages cut when they write stuff like that.
  - They use row names, which is just cumbersome and all around bad.
- I guess you meant `identical()`, not `identity()`.
- Better to capture the output from `all.equal()` and test with `isTRUE()`, this way you are told that 
  ```
  [1] "Component “slurmConfig”: 'is.NA' value mismatch: 25 in current 27 in target"
  ```

What you end up with is `SSP2-Base_d50_bIT` and `SSP2-Base_d95_bIT` inheriting the `SSP2-Base_d95high_bIT` value for `slurmConfig` of five instead of pestering the user.  No way around that.

orichters:
>     * Jeeez, these are awful functions …
>       
>       * They do assume some working directory.  People should have their wages cut when they write stuff like that.
>       * They use row names, which is just cumbersome and all around bad.

Too bad, my contract was just extended…
- @1: You need to specify a `remindPath`, and this has a default of `.`. Specify otherwise if you like.
- @2: I have no preference on that, but the rownames should be unique anyway for config files, so I don't see any immediate problem caused by that.

>     * I guess you meant `identical()`, not `identity()`.

Yes.

>     * Better to capture the output from `all.equal()` and test with `isTRUE()`, this way you are told that
>       ```
>       [1] "Component “slurmConfig”: 'is.NA' value mismatch: 25 in current 27 in target"
>       ```
> 
> What you end up with is `SSP2-Base_d50_bIT` and `SSP2-Base_d95_bIT` inheriting the `SSP2-Base_d95high_bIT` value for `slurmConfig` of five instead of pestering the user. No way around that.

I think that is ok, but I think worth telling the user. Because in case of slurmConfig it may not matter, but in other cases it may be of relevance. As far as I see, this happens if say `SSP2-Base` has a value declared in a cell, but `SSP2-Base_d50_bIT` hasn't (and would therefore use the default from main.gms, if it wasn't slurmConfig). Now, it uses the value from `SSP2-Base`, which indeed changes the switch (and potentially the model behavior). Yes, I think the user should be warned about that.

I tried to take the scenario config and ran:
```
cascade_scenarios("scenario_config_NGFS_v3.csv", list("SSP2-Base_bIT" = c("SSP2-Base_d95high_bIT", "SSP2-Base_d50_bIT", "SSP2-Base_d95_bIT", "SSP2-Base_d50high_bIT")))
cascade_scenarios("scenario_config_NGFS_v3.csv", list("SSP2-Base_bIT" = c("SSP2-Base_d95high_bIT", "SSP2-Base_d50_bIT", "SSP2-Base_d95_bIT", "SSP2-Base_d50high_bIT"), "h_ndc_bIT" = c("h_cpol_bIT", "o_1p5c_bIT", "o_1p5c_bIT_46NDC", "o_2c_bIT")))
```
but the second lead to an error:
```
Error in `pivot_wider()`:                                                                                                                                             
! Names must be unique.
x These names are duplicated:
  * "copyConfigFrom" at locations 2 and 4.
```
Should I never run this script on a config that already has a `copyConfigFrom` column? But running the first command a second time works without problems, seems it just does nothing then. So I also cannot use the script to overwrite existing `copyFromConfig` settings, right?

The other really minor thing what I noticed with the NGFS config is that we have several outcommented scenarios, and they don't line up anymore after running the script, because they don't get the column added. But maybe not that important after all.

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
> I tried to take the scenario config and ran:
> 
> ```
> cascade_scenarios("scenario_config_NGFS_v3.csv", list("SSP2-Base_bIT" = c("SSP2-Base_d95high_bIT", "SSP2-Base_d50_bIT", "SSP2-Base_d95_bIT", "SSP2-Base_d50high_bIT")))
> cascade_scenarios("scenario_config_NGFS_v3.csv", list("SSP2-Base_bIT" = c("SSP2-Base_d95high_bIT", "SSP2-Base_d50_bIT", "SSP2-Base_d95_bIT", "SSP2-Base_d50high_bIT"), "h_ndc_bIT" = c("h_cpol_bIT", "o_1p5c_bIT", "o_1p5c_bIT_46NDC", "o_2c_bIT")))
> ```
> 
> but the second lead to an error:
> 
> ```
> Error in `pivot_wider()`:                                                                                                                                             
> ! Names must be unique.
> x These names are duplicated:
>   * "copyConfigFrom" at locations 2 and 4.
> ```

Fixed.

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
> Because in case of slurmConfig it may not matter, but in other cases it may be of relevance. As far as I see, this happens if say `SSP2-Base` has a value declared in a cell, but `SSP2-Base_d50_bIT` hasn't (and would therefore use the default from main.gms, if it wasn't slurmConfig). Now, it uses the value from `SSP2-Base`, which indeed changes the switch (and potentially the model behavior). Yes, I think the user should be warned about that.

There is just functionality lacking to do that properly.  What would be necessary is a function to normalise the configurations.  So read the defaults from `main.gms` and `default.cfg`, read a .csv file, replace all missing values either by those from the the scenario referenced in `copyConfigFrom` or from the default scenario, and return the complete configurations.
Little point in telling users there's a difference in the configuration when one can't tell them whether it is relevant or not.

```
full_join(
    old %>% 
        as_tibble(rownames = 'title') %>% 
        pivot_longer(cols = -'title', values_to = 'old', 
                     values_transform = as.character),
    
    new %>% 
        as_tibble(rownames = 'title') %>% 
        pivot_longer(cols = -'title', values_to = 'new', 
                     values_transform = as.character),
    
    c('title', 'name')
) %>% 
    group_by(title, name) %>% 
    filter(!identical(old, new)) %>% 
    ungroup() %>% 
    mutate(value = paste(old, new, sep = ' -> ')) %>% 
    select(-old, -new) %>% 
    pivot_wider()
  ```

orichters:
I did some more checks and I think it works on scenario configs, as long as you don't have a `copyConfigFrom` already in place, as replacing cascades does not seem to work. So I think it would be good to add some more hints to the user how to use it.

PR #1232, opened 2023-03-01
bugfix CO2 tax convergence calculation in regipol

Renato-Rodrigues:
## Purpose of this PR
 - fix CO2 tax convergence calculation with multiple targets to avoid mismatch between emission market prices
 -  update ECEMF scenario config

## Type of change

- [x] Bug fix 

## Checklist:


robertpietzcker (approved):
I assume you checked the resulting CO2 prices and they are now equal across sectors when there is a uniform carbon price :-)

Renato-Rodrigues:
yes, the changes are tested and working on two test runs I made, but I will only merge it tomorrow once all policy runs finish to make sure no unforeseen issue remains.

PR #1236, opened 2023-03-06
add non-convergence to debugging tutorial, cleanup help text

orichters:
## Purpose of this PR

- expand debugging tutorial to explain what one might check if REMIND did not convergence
- harmonize formatting of help text in output.R, start.R, start_bundle_coupled.R
- use actual GAMS names in `nash_info_convergence.csv` such that one can search them more easily

## Type of change

- [x] documentation update

## Checklist:

### DIFF
--- a/tutorials/10_DebuggingREMIND.md
+++ b/tutorials/10_DebuggingREMIND.md
@@ -9,8 +9,11 @@ First, find out the state of your run by executing this in the run directory:
 Rscript -e "modelstats::loopRuns('.')"
 ```
 PIK cluster users can access information on all their runs in all directories by executing `rs2 -c` (or `rs2 -a` for only active runs).
-If you want information on all (or some) runs in a given output folder, no matter who ran them, simply type `rs2` and then select either `1` for all runs or just the runs you want information about. 
-In case of errors, this tutorial should help you.
+If you want information on all (or some) runs in a given output folder, no matter who ran them, simply type `rs2` and then select either `1` for all runs or just the runs you want information about.
+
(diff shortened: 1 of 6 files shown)


tutorials/10_DebuggingREMIND.md
```diff
@@ -134,8 +137,29 @@ If infeasibilities show up already in the first iteration, it may be related to
...
+
+If the run was mostly converged in that iteration (which depends on your standards), one can set `cm_nash_autoconverge` to `3`, which relaxes the requirements on market clearing further. This should then lead to the restarted run actually converging in that iteration. You can run REMIND for more iterations by modifying `cm_iteration_max` in [`80_optimization/nash/datainput.gms`](../modules/80_optimization/nash/datainput.gms) and the set iteration in [`core/sets.gms`](../core/sets.gms), which by default only goes to 200.
```
LaviniaBaumstark, line 153: this is not recommended for scenarios for paper or project submission. Maybe also the existing solution is sufficient for the purpose if only a few markets fail in late time steps. Would be even faster than restarting with lower tolerance criteria

LaviniaBaumstark (commented):
thanks

tutorials/10_DebuggingREMIND.md
robertpietzcker: I would propose to add "This should not be used for paper runs, but can be ok for project runs if p80_defic_sum_rel is low (<0.02 or so)".

Often enough, the solution in iteration 100 can be quite bad, which is why I would usually rather go for restarting with relaxed limits. On using the run for a project: To my understanding the current thresholds are arbitrary, so doubling increases the imbalance, but doesn't necessarily make the runs "wrong" or "bad". I would say for almost all project uses this is totally sufficient - only when the paper submission time approaches it might be safer to go for a tighter convergence criterion.

LaviniaBaumstark (approved):

PR #1237, opened 2023-03-06
avoid unnecessary fails if runs are restarted automatically

orichters:
## Purpose of this PR

- if a coupled run is run on `standby`, is aborted and restarted automatically, it fails because the folder already contains files (no surprise). After this PR, it only prints a warning.
- I remember adding this `stop` when implementing #794 rather as a check that no coupled run was started that was already on his way, but `start_bundle_coupled.R` either avoids that or deletes the folder, so this serves no functionality anymore.

## Type of change

- [x] Bug fix

## Checklist:

* The error was identified here: `/p/projects/magpie/users/florianh/projects/paper/LandEnergy/modelruns/remind/output/C_SSP2EU-PkBudg630-EatLancetPure-rem-3/`


orichters:
FYI @flohump

PR #1238, opened 2023-03-06
Remove the splitting of deltacap to two time steps

robertpietzcker:
## Purpose of this PR

Remove the splitting of deltacap to two time steps: from now on, capacity additions in one time step are (again) directly accounted for in the capacities in this time step.
To prevent the model strongly changing the 2025 results in an ambitious climate policy scenario (which would be unrealistic as today is 2023 and changes to policies and investments take time to implement), this PR introduces a new mechanism so that the model sees (budget-neutral) costs when it changes the results in cm_startyear compared to the reference run from which it starts. The costs are a quadratic function of the relative change, to allow small changes put prevent large changes. The relevant equations and variables all contain "changeProdStartyear" in their names. 

The parameters c_changeProdCost and c_SlackMultiplier in main.gms determine how close the model wants to stay to the values in the reference run. 
Currently, the default is set to c_changeProdCost  = 5 and c_SlackMultiplier = 1, but this may be adjusted in the validation in the coming weeks. 
As time progresses, these values should be adjusted: c_changeProdCost should be increased in 2024 and 2025 (as the resulting changes possible will become smaller and smaller), until in 2026, cm_startyear is shifted to 2030, at which point c_changeProdCost should be reduced to allow larger deviations from the reference run in 2030.  

## Type of change

- [x] Fundamental change
(in the 3.0.0 documentation, the old q_cap and q_transPe2se are printed in full, and they still contain 
+  pm_dt(ttot)/2 / pm_dataeta(ttot,regi,te)
                * pm_omeg(regi,"2",te)
                * vm_deltaCap(ttot,regi,te,rlf) 
this should be removed/ will be automatically removed when running goxygen again. 

## Checklist:

* Test runs are here:  /p/projects/remind/users/robertp/save_for_pullRequests/deltacaphalf_v1k-tax
* Comparison of results (what changes by this PR?): a compScen comparing Base/NDC/PkBudg500 before and after the changes can be found in /p/projects/remind/users/robertp/save_for_pullRequests/deltacaphalf_v1k-tax; 
some findings are described in issue https://github.com/remindmodel/development_issues/issues/185


core/bounds.gms
```diff
@@ -519,5 +519,12 @@ $IFTHEN.transpGDPscale not "%cm_transpGDPscale%" == "on"
   vm_transpGDPscale.fx(t,regi) = 1;
 $ENDIF.transpGDPscale
 
+***----------------------------------------------------------------------------
+*'  Limit slack variable and uncontrolled variable values for adj costs that limit changes to REF in cm_startyear
```
Renato-Rodrigues, line 523: could you change "REF" to "reference" to avoid people confusing this with the REF region?

core/declarations.gms
```diff
@@ -228,6 +228,14 @@ pm_share_CCS_CCO2(ttot,all_regi)                      "share of stored CO2 from
 
 pm_delta_histCap(tall,all_regi,all_te)                "parameter to store data of historic capacity additions [TW/yr]"
 
+p_prodSeReference(ttot,all_regi,all_enty,all_enty,all_te)         "read from gdx the output value in the reference run"
```
Renato-Rodrigues, line 231: I would change the description of these parameter to something like: "secondary energy in reference run", "final energy in reference run",...
Also, please add the units if possible to the parameter definition in between brackets.

core/equations.gms
```diff
@@ -55,7 +55,7 @@ q_costInv(t,regi)..
 q_costInvTeDir(t,regi,te)..
   v_costInvTeDir(t,regi,te)
   =e=
-  vm_costTeCapital(t,regi,te) * sum(te2rlf(te,rlf), vm_deltaCap(t,regi,te,rlf) )
+  vm_costTeCapital(t,regi,te) * sum(te2rlf(te,rlf), vm_deltaCap(t,regi,te,rlf)) * 1.05**(pm_ts(t) / 2)
```
Renato-Rodrigues, line 58: could you add a line to the equation comments explaining the 1.05 value here? Could this number be included as a scalar/parameter instead of hard codded to the equation?

core/equations.gms
```diff
@@ -66,7 +66,9 @@ v_adjFactorGlob.fx(t,regi,te) = 0;
 q_costInvTeAdj(t,regi,teAdj)..
   v_costInvTeAdj(t,regi,teAdj)
   =e=
-  vm_costTeCapital(t,regi,teAdj) * ( (p_adj_coeff(t,regi,teAdj) * v_adjFactor(t,regi,teAdj)) + (p_adj_coeff_glob(teAdj) * v_adjFactorGlob(t,regi,teAdj) ) )
+  vm_costTeCapital(t,regi,teAdj) * (
+    (p_adj_coeff(t,regi,teAdj) * v_adjFactor(t,regi,teAdj)) + (p_adj_coeff_glob(teAdj) * v_adjFactorGlob(t,regi,teAdj))
+  ) * 1.05**(pm_ts(t) / 2)
```
Renato-Rodrigues, line 71: same as previous, if possible please add a comment expaling the choice for the 1.05 value, or create a scalar to store it and add a description to the scalar.

core/equations.gms
```diff
@@ -838,16 +832,65 @@ q_limitCO2(ttot+1,regi) $((pm_ttot_val(ttot+1) ge max(cm_startyear,2055)) AND (t
...
+    / ( pm_ttot_val(ttot) - pm_ttot_val(ttot-1) )
+  , 2)
+  / ( sum(te2rlf(te,rlf), vm_deltaCap(ttot-1,regi,te,rlf)) + p_adj_seed_reg(ttot,regi) * p_adj_seed_te(ttot,regi,te)  
+      + p_adj_deltacapoffset("2010",regi,te)$(ttot.val eq 2010) + p_adj_deltacapoffset("2015",regi,te)$(ttot.val eq 2015)
+      + p_adj_deltacapoffset("2020",regi,te)$(ttot.val eq 2020) + p_adj_deltacapoffset("2025",regi,te)$(ttot.val eq 2025)
+    )
+;
+
+***---------------------------------------------------------------------------
+*' Calculate changes to REF in cm_startyear - needed to limit them via refunded adj costs
```
Renato-Rodrigues, line 853: please change REF to reference to avoid confusion with the region names convention

core/equations.gms
```diff
@@ -838,16 +832,65 @@ q_limitCO2(ttot+1,regi) $((pm_ttot_val(ttot+1) ge max(cm_startyear,2055)) AND (t
...
+      + p_adj_deltacapoffset("2020",regi,te)$(ttot.val eq 2020) + p_adj_deltacapoffset("2025",regi,te)$(ttot.val eq 2025)
+    )
+;
+
+***---------------------------------------------------------------------------
+*' Calculate changes to REF in cm_startyear - needed to limit them via refunded adj costs 
+***---------------------------------------------------------------------------
+*' calculating the absolute change of output with respect to the value in REF for each te (counting SE, FE, UE and CCS)
+q_changeProdStartyear(t,regi,te)$( (t.val gt 2005) AND (t.val eq cm_startyear ) )..
+  v_changeProdStartyear(t,regi,te)
```
Renato-Rodrigues, line 857: this is very minor, but as by default you only want to define v_changeProdStartyear to a single year always, you could remove the time dimension from the variable and equation, and use a sum to control the time dimension in vm_prodSe, vm_prodFE, ...
It wouldn't change in any way the way your formulation works, but it would reduce unnecessary dimensions from memory allocation and gdx files.
This is very minor in this case, but I just wanted to mention to keep in mind for alternative formulations in the future.

core/declarations.gms
```diff
@@ -354,7 +362,9 @@ vm_costCESMkup(ttot,all_regi,all_in)                  "CES markup cost to repres
 vm_taxrevimplicitQttyTargetTax(ttot,all_regi)        "quantity target bound implemented through implict tax"
 vm_taxrevimplicitPriceTax(ttot,all_regi,entySe,all_enty,sector)   "final energy price target implemented through implict tax"
 vm_taxrevimplicitPePriceTax(ttot,all_regi,all_enty)  "primary energy price target implemented through implict tax"
-
+v_changeProdStartyear(ttot,all_regi,all_te)          "absolute change of output with respect to the reference run for each te"
```
Renato-Rodrigues, line 365: could you add the units for these variables in between brackets?

main.gms
```diff
@@ -1095,6 +1095,16 @@ parameter
 *' * (0)                default, equals "off", no limit imposed
 *' * (any other number) limit of gas demand from 2025 on in Germany in EJ/yr
 *'
+parameter
+  cm_SlackMultiplier   "sets the slack size for v_changeProdStartyearSlack"
+;
+  cm_SlackMultiplier = 1; !! def 1
```
Renato-Rodrigues, line 1101: could you add the unit for the slack multiplier? if it is just a factor, or it is in percentage terms?

core/equations.gms
```diff
@@ -838,16 +832,65 @@ q_limitCO2(ttot+1,regi) $((pm_ttot_val(ttot+1) ge max(cm_startyear,2055)) AND (t
...
+  =e=
+  sum(pe2se(enty,enty2,te),   vm_prodSe(t,regi,enty,enty2,te)  - p_prodSeReference(t,regi,enty,enty2,te) )
+  + sum(se2se(enty,enty2,te), vm_prodSe(t,regi,enty,enty2,te)  - p_prodSeReference(t,regi,enty,enty2,te) )
+  + sum(se2fe(enty,enty2,te), vm_prodFE(t,regi,enty,enty2,te)  - p_prodFEReference(t,regi,enty,enty2,te) )
+  + sum(fe2ue(enty,enty2,te), vm_prodUe(t,regi,enty,enty2,te)  - p_prodUeReference(t,regi,enty,enty2,te) )
+  + sum(ccs2te(enty,enty2,te), sum(teCCS2rlf(te,rlf), vm_co2CCS(t,regi,enty,enty2,te,rlf) - p_co2CCSReference(t,regi,enty,enty2,te,rlf) ) )
+;
+
+*' calculating the relative change 
+q_relChangeProdStartYear(t,regi,te)$( (t.val gt 2005) AND (t.val eq cm_startyear ) )..
```
Renato-Rodrigues, line 867: could you move the denominator of the division to the left hand side of the equation as a multiplication term?
I am not sure this will ever be the case, but just to be on the safe side we avoid divisions by zero error in extreme cases, like for example very small numbers in the denominator. So it is easier to just transform the equation into multiplication on both sides instead of adding a division term, as `10^-12 * 0` is defined, meanwhile `# / 10^-12` can be not defined depending of the zero tolerance we set in gams.

Renato-Rodrigues (changes requested):
My comments are mainly small changes either improve documentation or be on the safe side to avoid division by zero errors.

core/equations.gms
robertpietzcker: right. I didn't think about this but just automatically added the time dimension (while trying to make sure that only one entry will ever be filled in a run), but you are right - especially if one loads other gdxes from runs fixed until other times, there might be residual information carried over in the gdx, which would be confusing...

I will check if I can remove the time dimension today

core/equations.gms
robertpietzcker: did so

core/bounds.gms
robertpietzcker: did so

core/equations.gms
robertpietzcker: I changed it to: 

  * (1.02 + pm_prtp(regi) ** (pm_ts(t) / 2) !! This increases the investments as if the money was actually borrowed 
  !! half a time step earlier, using an interest rate of pm_prtp + 2%, which is close to the model-endogenous interest rate. 
  !! We do this to reduce the difference to the previous version where the effect of deltacap on capacity was split 
  !! half to the current and half to the next time. 

is this ok?

core/equations.gms
robertpietzcker: as above

Renato-Rodrigues (approved):

core/equations.gms
```diff
@@ -838,16 +841,65 @@ q_limitCO2(ttot+1,regi) $((pm_ttot_val(ttot+1) ge max(cm_startyear,2055)) AND (t
...
+*' calculating the absolute effect size: (relative change)^2 * value in the reference run * construction time (as proxy for "how easy to change on short notice") 
+q_changeProdStartyearAdj(t,regi,te)$( (t.val gt 2005) AND (t.val eq cm_startyear ) )..
+  v_changeProdStartyearAdj(t,regi,te)
+  =e=
+  power( v_relChangeProdStartYear(t,regi,te), 2 )  !! taking the square to a) treat increase and decrease the same; b) to penalize larger changes 
+  * ( p_prodAllReference(t,regi,te) + p_adj_seed_reg(t,regi) * p_adj_seed_te(t,regi,te) ) !! tie back to the absolute change 
+  * ( pm_data(regi,"constrTme",te)$(pm_data(regi,"constrTme",te) gt 0) + 2$(pm_data(regi,"constrTme",te) eq 0)) !! take construction time 
+;
+
+*' calculating the resulting costs (which are applied as a tax in module 21, so they have no budget effect but only incluence REMIND choices)
```
Renato-Rodrigues, line 895:
```suggestion
*' calculating the resulting costs (which are applied as a tax in module 21, so they have no budget effect but only influence REMIND choices)
```

Renato-Rodrigues (approved):

core/declarations.gms
robertpietzcker: done

core/declarations.gms
robertpietzcker: done

main.gms
robertpietzcker: done

core/equations.gms
robertpietzcker: in this case the parameters "p_adj_seed_reg(t,regi) * p_adj_seed_te(t,regi,te)" should ensure that the denominator never gets very small, but I agree that it is always better to formulate relative variables with a multiplication, so I changed as you proposed.

lecfab:
Would someone feel qualified to write some more documentation for these switches?
I'm not sure which values we should use, even after reading the [mattermost conversation](https://mattermost.pik-potsdam.de/rd3/pl/i9fia9y37fge7qk1qefrndhcga).

PR #1239, opened 2023-03-07
adding EU21 scenarios and doing some cleaning in the default config

Renato-Rodrigues:
## Purpose of this PR

- adding EU21 scenarios: SSP2EU-EU21-calibrate, SSP2EU-EU21-NPi-calibrate, SSP2EU-EU21-Base, SSP2EU-EU21-NDC, SSP2EU-EU21-NPi, SSP2EU-EU21-PkBudg500, SSP2EU-EU21-PkBudg1150
- adding comment separators (#) between groups of runs 
- changing NPi and NDC definition of Europe carbon pricing in all SSP2 scenarios to follow EU21 representation of European targets (NPi = 40% reduction of emissions by 2030, NDC = 55% reduction of emissions by 2030, both following lower target defined for energy efficiency - cm_implicitQttyTarget).   
- removing values if they follow the default REMIND definition to make the differences between scenarios more clear    
- removing switches that are exaclty the same between all scenarios: cm_taxCO2inc_after_peakBudgYr, cm_CO2priceRegConvEndYr, trade, cm_import_EU, cm_adj_coeff_multiplier

## Open questions:

- [x] is it intentional for only SSP2_NPi `c_tech_earlyreti_rate` to change the value from `CHA_regi.pc 0.13` to `CHA_regi.pc 0.14`? 
  Values were harmonized between policy scenarios using the 0.13 value.
- [x] as we are using water heating to all scenarios, shouldn't we simply change it to the default module realisation?
  This will be discussed in the REMIND meeting, and the column could be removed to simplify the scenario config file in case we adopt `heating` as the new `water` module realization default.   

## Type of change

- [x] Refactoring

## Checklist:


config/scenario_config.csv
```diff
@@ -1,45 +1,56 @@
...
+title;start;CES_parameters;slurmConfig;regionmapping;extramappings_historic;cm_rcp_scen;cm_iterative_target_adj;subsidizeLearning;c_budgetCO2from2020;carbonprice;cm_co2_tax_2020;cm_peakBudgYr;cm_emiscen;c_regi_earlyreti_rate;c_tech_earlyreti_rate;cm_fetaxscen;cm_co2_tax_growth;cm_bioenergy_SustTax;cm_maxProdBiolc;c_ccsinjecratescen;c_ccscapratescen;cm_CESMkup_build;cm_CESMkup_ind;techpol;regipol;cm_implicitQttyTarget;cm_loadFromGDX_implicitQttyTargetTax;cm_emiMktTarget;cm_NucRegiPol;cm_CoalRegiPol;cm_altFeEmiFac;cm_POPscen;cm_GDPscen;cm_demScen;cm_oil_scen;cm_gas_scen;cm_coal_scen;c_techAssumptScen;cm_nucscen;cm_so2tax_scen;cm_multigasscen;cm_LU_emi_scen;cm_tradecostBio;cm_1stgen_phaseout;c_SSP_forcing_adjust;cm_APscen;water;cm_EDGEtr_scen;cm_startyear;path_gdx;path_gdx_ref;path_gdx_bau;path_gdx_refpolicycost;description
+# H12 SSP2;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
```
LaviniaBaumstark, line 2: let`s call it SSP2EU

config/scenario_config.csv
```diff
@@ -1,45 +1,56 @@
...
+SSP2EU-PkBudg1150;1;;;;;rcp26;9;globallyOptimal;1150;diffCurvPhaseIn2Lin;60;2100;9;;;;;;;;;;;NDC;;2030.EUR_regi.tax.t.FE_wo_b_wo_n_e.all 1.2809;on;;on;on;EUR_regi, NEU_regi;;;;;;;;;;;;;;;;heat;Mix3;2025;;SSP2EU-NDC;SSP2EU-Base;;SSP2EU-PkBudg1150: This climate policy scenario follows the Shared Socioeconomic Pathways 2 called Middle of the Road. The stylized climate policy scenario assumes a peak budget of 1150 Gt?CO2 on total CO2 emissions from 2015 to 2100. This is a weel below 2??C scenario, peak warming is allowed to be at or slightly above 2??C, at median climate sensitivity but returns to values below 2??C in at least 67?% of scenarios by the end of the century.
+# EU21 SSP2;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
```
LaviniaBaumstark, line 10: and this is also SSP2EU

LaviniaBaumstark (commented):
Thanks for adjusting EU21 scenarios and deleting settings which are the same for all scenarios.
I personally really prefer to have the values in the scenario_config even if they are the default. Makes it more precise for me and makes it possible to compare scenario_config files between runs with different default settings.

LaviniaBaumstark:
regarding "is it intentional for only SSP2_NPi c_tech_earlyreti_rate to change the value from CHA_regi.pc 0.13 to CHA_regi.pc 0.14" @christophbertram or @orichters should know.

LaviniaBaumstark:
regarding "as we are using water heating to all scenarios, shouldn't we simply change it to the default module realisation" let's discuss it at the next REMIND meeting

Renato-Rodrigues:
> I personally really prefer to have the values in the scenario_config even if they are the default. Makes it more precise for me and makes it possible to compare scenario_config files between runs with different default settings.

I find it much easier to find the switches that matter for each scenario and the differences between scenarios with a cleaner csv without including the default values. 
For example, I only found the `c_tech_earlyreti_rate` different values which is either a mistake or needs to be updated in other scenarios using the cleaner csv file. The same for the switches I removed because they do not change between any scenario and use the default values, and the `water` common switch use that I only identified after looking at the version without defaults included.

For comparing different scenarios in detail, I always use the cfg.txt file.

Renato-Rodrigues:
> I personally really prefer to have the values in the scenario_config even if they are the default. Makes it more precise for me and makes it possible to compare scenario_config files between runs with different default settings.

I find it much easier to find the switches that matter for each scenario and the differences between scenarios with a cleaner csv without including the default values. 
For example, I only found the `c_tech_earlyreti_rate` different values which is either a mistake or needs to be updated in other scenarios using the cleaner csv file. The same for the switches I removed because they do not change between any scenario and use the default values, and the `water` common switch use that I only identified after looking at the version without defaults included.

For comparing different scenarios in detail, I always use the cfg.txt file.

Renato-Rodrigues:
> regarding "is it intentional for only SSP2_NPi c_tech_earlyreti_rate to change the value from CHA_regi.pc 0.13 to CHA_regi.pc 0.14" @christophbertram or @orichters should know.

@cchrisgong as it affects China, maybe Chris knows something about it also.

christophbertram:
I don't know, in my view 13 or 14 % does not really make a difference, so it should rather be harmonized...

orichters:
> I find it much easier to find the switches that matter for each scenario and the differences between scenarios with a cleaner csv without including the default values.

Agree. This PR here will allow to get the full set of settings returned nevertheless, filling each empty cell: https://github.com/remindmodel/remind/pull/1234

Renato-Rodrigues:
> I don't know, in my view 13 or 14 % does not really make a difference, so it should rather be harmonized...

I will harmonized it for now then, and we can review this if anybody else brings new info about this change.

Renato-Rodrigues:
> > I find it much easier to find the switches that matter for each scenario and the differences between scenarios with a cleaner csv without including the default values.
> 
> Agree. This PR here will allow to get the full set of settings returned nevertheless, filling each empty cell: #1234

Thank you Oliver for introducing once again another useful feature to our arsenal!

@LaviniaBaumstark Would you be ok with the alternatives that both me and Oliver mentioned, or do you want to further discuss this?

LaviniaBaumstark:
yes, fine for me

Renato-Rodrigues:
I will merge as soon as you give the approval for the PR then.

LaviniaBaumstark (approved):

PR #1250, opened 2023-03-10
CES calibration pdf report - support for multiple pages tables

Renato-Rodrigues:
## Purpose of this PR

- Split quantity and prices outliers tables into multiple pages if necessary

## Type of change

- [x] Refactoring
- [x] New feature


robertpietzcker (approved):
I really don't know anything about formatting via R, so I can't comment if this is the best way to do it. Anyway, if it does what it is supposed to do - great!

PR #1274, opened 2023-03-30
Make the magicc version configurable

mikapfl:
## Purpose of this PR

As a first step towards updating our magicc coupling, add configuration switches and configs to run REMIND with MAGICC7 instead of MAGICC6. We only have a magicc7 configuration for the default settings, not the full set of configs as we have for magicc6. Also, this PR does not include AR6-compatible pre-processing, we continue using the current code for that.

One pressing reason why we need magicc7 is that most likely magicc6 will not work on the new cluster (or only with a lot of effort).

The default magicc version remains magicc 6 for now, but this PR adds the option to change it to magicc7.

## Type of change

- [x] New feature 

## Checklist:

## Further information:

We still need to do test runs. @LaviniaBaumstark


0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
> One pressing reason why we need magicc7 is that most likely magicc6 will not work on the new cluster (or only with a lot of effort).

I'm no friend of MAGICC's, but the last time we moved clusters, that wasn't a big deal …  Any specific reason for that?

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
Well, then … Kill it, kill it with fire!

gabriel-abrahao:
@LaviniaBaumstark @tonnrueter For the cluster change, I think it's better to merge this here than to rush the climate-assessment PR. I think that one will be done in time, but nevertheless I don't think we want to retire the old capability so quickly.

LaviniaBaumstark:
@tonnrueter how to proceed with this PR?

PR #1287, opened 2023-04-12
improve searching for old and new runs in start_bundle_coupled, expand coupling tests

orichters:
## Purpose of this PR

- avoid that if you use `/p/tmp/oliverr/magpie` as magpie folder, the script does not find the right place because a final `/` is missing 😠
- improve the search for old runs, not finding folders such as `C_Base-rem-5-failed` etc. anymore that have some postfix after -rem-[0-9]+.
- update snapshot to one that actually works ;)
- expand `make test-coupled` substantially, checking plausibility of RData files, check `path_mif_ghgprice_land`, check whether mif files are written etc.
- add errors if `path_mif_ghgprice_land` was not found, adapt colors such that basing this on current runs is blue, not red because the mif file not yet exists.
- avoid that folders are deleted if there were previous errors
- run tests on a single core, not 13 (requires --mem=8000)
- write `cfg_rem$gms$cm_nash_autoconverge ` already in `start_bundle_coupled` correctly such that it can be tested
- use quitte instead of magclass for combining the runs to avoid error because MAgPIE empty fulldata.gdx only has data until 2100.
- don't run convergence documents, because they fail because MAgPIE empty fulldata.gdx only has data until 2100.
- avoid that `--gamscompile` prints success story about started runs if it hasn't started any

## Type of change

- [x] Bugfix
- [x] Refactoring

## Checklist:

- [x] All automated model tests pass (`FAIL 0` in the output of `make test-coupled`)


fbenke-pik (approved):

dklein-pik (approved):
Great, thank you!

start_bundle_coupled.R
```diff
@@ -203,14 +203,14 @@ if ("max_iterations" %in% colnames(scenarios_coupled)) {
...
-common <- intersect(rownames(settings_remind),rownames(scenarios_coupled))
+common <- intersect(rownames(settings_remind), rownames(scenarios_coupled))
 knownRefRuns <- apply(expand.grid(prefix_runname , common, "-rem-", seq(max_iterations)), 1, paste, collapse="")
 if (! identical(common, character(0))) {
   message("\n################################\n")
   message("The following ", length(common), " scenarios will be started:")
   message("  ", paste(common, collapse = ", "))
 } else {
-  stop("No scenario found with start=", startgroup, " in ", basename(path_settings_coupled), ".")
+  stop("No scenario found with start=", startgroup, " in ", basename(path_settings_coupled), " that is also defined in ", basename(path_settings_remind), ".")
```
orichters, line 213: added that because the error message was not explaining the actual problem in one case

dklein-pik (approved):
Wow, thanks. The tests are cool!

gabriel-abrahao:
Thanks Oliver, great work! Unfortunately running `make test-coupled` in a fresh REMIND clone failed for me, see `/p/projects/piam/abrahao/scratch/coupling/remind/failed.log`. 

To me it looks like `gamscompile` breaks because the input data is not automatically downloaded and GAMS can't handle that, the tests did seem to work after I manually started a run in the same folder and the input data was distributed. 

To avoid that, maybe add something like `Rscript start.R config/tests/scenario_config_compile.csv` in the Makefile before the test, like you did in #1223 @orichters ?

orichters:
Thanks, Gabriel, for catching this. I rather prefer a different option: To compile if you run `start_bundle_coupled.R` with `--gamscompile`. This also helps if you do that manually, so not as part of the tests.

gabriel-abrahao (approved):

PR #1292, opened 2023-04-17
Update adjustment costs

robertpietzcker:
## Purpose of this PR

increase adjustment costs to get a more realistic scale-up over the next decade

## Type of change

## Checklist:

* Test runs are here: /p/projects/remind/users/robertp/save_for_pullRequests/2024-04-13_CHAcapCF2/output : "NPiFlex_CHAcapCF3_adj" vs "CHAcapCF3"
* Comparison of results: slight slowing down of upscaling of PV and other new techs.


core/datainput.gms
```diff
@@ -1061,7 +1068,7 @@ $elseif not "%cm_adj_coeff%" == "off"
   p_adj_coeff(t,regi,te)$p_new_adj_coeff(te) = p_new_adj_coeff(te);
 $endif
 
-p_adj_coeff(ttot,regi,te)            = 25 * p_adj_coeff(ttot,regi,te);  !! Rescaling all adjustment cost coefficients
```
LaviniaBaumstark, line 1064: why for all? what is the meaning of 25/32?

robertpietzcker: The general perception that REMIND is a bit too flexible in its upscaling.
So the commit 
a) slows down PV, wind, wind offshore
b) treats all <x>2h2, <x>2liquids similar (with a gas faster than coal faster than biomass due to complexity of the plant due to material handling and preprocessing) - before there were simply techs missing in the list
c) makes everything 28% higher

LaviniaBaumstark (approved):

core/datainput.gms
```diff
@@ -1008,25 +1008,32 @@ $ENDIF.WindOff
...
   p_adj_coeff(ttot,regi,"coalftrec")       = 0.6;
-  p_adj_coeff(ttot,regi,"coalftcrec")      = 0.8;
-  p_adj_coeff(ttot,regi,"spv")             = 0.08;
-  p_adj_coeff(ttot,regi,"wind")            = 0.15;
-  p_adj_coeff(ttot,regi,"geohe")            = 0.6;
+  p_adj_coeff(ttot,regi,"bioftrec")        = 0.65;
+  p_adj_coeff(ttot,regi,"gash2")           = 0.35;
+  p_adj_coeff(ttot,regi,"coalh2")          = 0.55;
+  p_adj_coeff(ttot,regi,"bioh2")           = 0.6;
+  p_adj_coeff(ttot,regi,teCCS)             = 1.0;
```
Renato-Rodrigues, line 1028: `coalftcrec` is part of `teCCS`. 
As `teCCS` definition is now written after the `coalftcrec` definition, you are redefining its value from 0.6 to 1.
If this is not intentional, i.e. `coalftcrec` should be equal to 0.6, please move the `teCCS` line above `coalftcrec ` definition. If it is intentional, please remove the `coalftcrec` definition to avoid confusion.

PS: I didn't checked if this also happens with other techs.

PR #1294, opened 2023-04-17
Update early retirement

robertpietzcker:
## Purpose of this PR

- make ngt and chp plants slower to retire
- ensure that the equations linking early retirements in consecutive time steps are applied correctly (q_limitCapEarlyReti, q_smoothphaseoutCapEarlyReti). 

## Type of change

- [x] bugfix

## Checklist:

* Test runs are here: /p/projects/remind/users/robertp/save_for_pullRequests/2024-04-13_CHAcapCF2/output : "NDC" vs "NDC_CHAcapCF2"
* Comparison of results (what changes by this PR?): some regions quickly phased out chp and ngt until 2020 - this is now prevented. Also, the model was free in increasing its 2015 early retirements, which is now also solved

### DIFF
--- a/core/equations.gms
+++ b/core/equations.gms
@@ -888,17 +888,18 @@ q_changeProdStartyearCost(t,regi,te)$( (t.val gt 2005) AND (t.val eq cm_startyea
 ***---------------------------------------------------------------------------
 *' The use of early retirement is restricted by the following equations:
 ***---------------------------------------------------------------------------
-q_limitCapEarlyReti(ttot,regi,te)$(ttot.val lt 2109 AND pm_ttot_val(ttot+1) ge max(2010, cm_startyear))..
-        vm_capEarlyReti(ttot+1,regi,te)
+q_limitCapEarlyReti(ttot,regi,te)$(ttot.val lt 2120 AND pm_ttot_val(ttot) ge max(2010, cm_startyear)).. !! 2000 doesn't have capacity, so for cm_startyear = 2005 the equation should not be applied
+        vm_capEarlyReti(ttot,regi,te)
         =g=
-        vm_capEarlyReti(ttot,regi,te);
+        vm_capEarlyReti(ttot-1,regi,te);
 
-q_smoothphaseoutCapEarlyReti(ttot,regi,te)$(ttot.val lt 2120 AND pm_ttot_val(ttot+1) gt max(2010, cm_startyear))..
-        vm_capEarlyReti(ttot+1,regi,te)
+q_smoothphaseoutCapEarlyReti(ttot,regi,te)$(ttot.val lt 2120 AND pm_ttot_val(ttot) ge max(2010, cm_startyear)).. !! 2000 doesn't have capacity, so for cm_startyear = 2005 the equation should not be applied
+        vm_capEarlyReti(ttot,regi,te)
         =l=
-        vm_capEarlyReti(ttot,regi,te) + (pm_ttot_val(ttot+1)-pm_ttot_val(ttot)) * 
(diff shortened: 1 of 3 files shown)


core/equations.gms
```diff
@@ -888,17 +888,18 @@ q_changeProdStartyearCost(t,regi,te)$( (t.val gt 2005) AND (t.val eq cm_startyea
 ***---------------------------------------------------------------------------
 *' The use of early retirement is restricted by the following equations:
 ***---------------------------------------------------------------------------
-q_limitCapEarlyReti(ttot,regi,te)$(ttot.val lt 2109 AND pm_ttot_val(ttot+1) ge max(2010, cm_startyear))..
-        vm_capEarlyReti(ttot+1,regi,te)
+q_limitCapEarlyReti(ttot,regi,te)$(ttot.val lt 2120 AND pm_ttot_val(ttot) ge max(2010, cm_startyear)).. !! 2000 doesn't have capacity, so for cm_startyear = 2005 the equation should not be applied
```
Renato-Rodrigues, line 891: Do you want to have early retirement also in 2110? Or `ttot` in this equation should be applied only for 2010 to 2100?
If it is the first, I would change the `ttot.val lt 2120` to `ttot.val le 2100`.

PS: Besides removing the 2110 time step, I would use also equality conditionals, as at least in my opinion they are more straight forward to see the affected years in the equation.

core/preloop.gms
```diff
@@ -188,4 +188,9 @@ if (cm_startyear gt 2005,
 execute_load "input_ref.gdx", pm_PEPrice, pm_SEPrice, pm_FEPrice;
 );
 
+*** load vm_capEarlyReti(ttot,regi,te) from reference gdx to have a reference point for q_smoothphaseoutCapEarlyReti and q_limitCapEarlyReti
```
Renato-Rodrigues, line 191: maybe we should think about adding the feature of suing an alternative reference gdx to load these values to avoid the issue we had with the outdated feasibility frontier runs?

robertpietzcker: not sure I understand?

core/equations.gms
robertpietzcker: Thanks for checking this. 

I agree that this could be shifted to "le 2100".

I don't have an opinion about this, whether it is better to run until 2100 or 2110 - I don't think it matters, neither for the results nor the runtime (which was the reason why it was limited at all). I know it shouldn't run longer than 2110 as then it would collide with this line in core/bounds.gms:
 `vm_capEarlyReti.up(ttot,regi,te)$(ttot.val lt 2009 or ttot.val gt 2111) = 0;`
 
 If I hadn't merged already I would change as you propose, but now I guess I wouldn't open a new PR. Or do you feel strongly about it?

core/preloop.gms
Renato-Rodrigues: sorry, my comment was related with the `q_eqadj` and has no connection with this PR. We can talk about it a bit further the next time we meet.

core/equations.gms
Renato-Rodrigues: I don't have strong feelings about it. But maybe it would be nice to have the change added just to keep consistence. 
Maybe you could create an issue pointing to this, so the next person that messes up with the early retirement code could take that into account for any future changes.

robertpietzcker: > I don't have strong feelings about it. But maybe it would be nice to have the change added just to keep consistence. Maybe you could create an issue pointing to this, so the next person that messes up with the early retirement code could take that into account for any future changes.

I put it in my next PR: https://github.com/remindmodel/remind/pull/1295

PR #1315, opened 2023-05-05
bugfix - improving support for targets set on regions that intersect with each other

Renato-Rodrigues:
## Purpose of this PR

- Fix bug in the regipol module that could make parent regions, e.g. EUR_regi, control the CO2 price updates for child regions, e.g. DEU, even if the child region have an independent regional target set.
- Improve handling of targets that oscillate between convergence states because iteration between multiple yearly targets

## Type of change

- [x] Bug fix 

## Checklist:

* Test runs are here: 
/p/projects/remind/users/renatoro/ECEMF/v11_2023_04_27/output/11_WP1_Nzero_2023-05-10_09.20.28


modules/47_regipol/regiCarbonPrice/sets.gms
```diff
@@ -20,6 +20,7 @@ emi_type_47 "emission type used in regional target"
 $ifThen.emiMkt not "%cm_emiMktTarget%" == "off" 
   regiEmiMktTarget(ext_regi)               "regions with emiMkt targets" / /
   regiANDperiodEmiMktTarget_47(ttot,ext_regi) "regions and periods with emiMkt targets" / /
+  regiEmiMktTarget2regi_47(ext_regi,all_regi)  "regions controlled by emiMkt market set to ext_regi" / /
```
robertpietzcker, line 23: sorry, maybe my mind is slow today. 
Does this simply spell out the regions that are in ext_regi for which EmiMktTarget applies?
so if an EmiMktTarget applies to EUR, then this mapping will contain the entries "(EUR,DEU), (EUR,FRA), (EUR, ENC), ..."?
and if you have separate EmiMktTargets for EUR and DEU, it would contain the entries "(DEU,DEU), (EUR,DEU), (EUR,FRA), (EUR, ENC), ..."?

If that is the case, then maybe describe it as "mapping from all ext_regi for which EmiMktTargets apply to their sub-regions"?

modules/47_regipol/regiCarbonPrice/postsolve.gms
```diff
@@ -195,7 +195,7 @@ pm_emiMktTarget_dev_iter(iteration, ttot,ttot2,ext_regi,emiMktExt) = pm_emiMktTa
 loop((ext_regi,ttot2)$regiANDperiodEmiMktTarget_47(ttot2,ext_regi),
   p47_targetConverged(ttot2,ext_regi) = 1;
   loop((ttot,emiMktExt,target_type_47,emi_type_47)$((pm_emiMktTarget(ttot,ttot2,ext_regi,emiMktExt,target_type_47,emi_type_47))),
-    if((abs(pm_emiMktTarget_dev(ttot,ttot2,ext_regi,emiMktExt)) > 0.01), !! if any emiMKt target did not converged
+    if((abs(pm_emiMktTarget_dev(ttot,ttot2,ext_regi,emiMktExt)) > 0.01), !! if emiMKt target did not converged
       p47_targetConverged(ttot2,ext_regi) = 0;
```
robertpietzcker, line 199: not sure if I overlook something, but if the target is specific for emiMktExt shouldn't also the parameter storing whether or not it converged contain this dimension? and accordingly, also the target adjustment only happen for this emiMktExt?

modules/47_regipol/regiCarbonPrice/declarations.gms
```diff
@@ -57,6 +57,7 @@ Parameter
   p47_firstTargetYear(ext_regi)                      "first year with a pre defined policy emission target in the region [year]"
   p47_lastTargetYear(ext_regi)                       "last year with a pre defined policy emission target in the region [year]"
   p47_currentConvergencePeriod(ext_regi)             "auxiliar parameter to store the current target year being executed by the convergence algorithm [year]"
+  p47_currentConvergence_iter(iteration,ttot,ext_regi) "auxiliar parameter to store if emission target is being executed by the convergence algorithm in the specific iteration [0 or 1]"
```
robertpietzcker, line 60: a) I don't fully get the description. 

would "auxiliary parameter to store if carbon price adjustment in response to a specific emission target is being executed by the convergence algorithm in the specific REMIND iteration [0 or 1]" also describe it? then I would prefer something along those lines. 

b) does this describe whether ANY price adjustment is being done, or whether or not a price adjustment is done due to one specific emission target? (so something like "there is an adjustment due to the ETS target for 2030, but no adjustment due to the ESR target for 2030"). in that case, the parameter would need to contain the emiMktExt as well, right?

only asking - if this would complicate things maybe it doesn't matter

modules/47_regipol/regiCarbonPrice/datainput.gms
```diff
@@ -26,13 +26,20 @@ p47_nonEnergyUse("2030",ext_regi)$(sameas(ext_regi, "EU27_regi")) = 0.11;
...
 *** Auxiliar parameters based on emission targets information 
   loop((ttot,ttot2,ext_regi,emiMktExt,target_type_47,emi_type_47)$pm_emiMktTarget(ttot,ttot2,ext_regi,emiMktExt,target_type_47,emi_type_47), !!calculated sets that depends on data parameter
     regiEmiMktTarget(ext_regi) = yes;
     regiANDperiodEmiMktTarget_47(ttot2,ext_regi) = yes;
   );
 
+  loop((ext_regi,ttot,ttot2,emiMktExt,target_type_47,emi_type_47)$pm_emiMktTarget(ttot,ttot2,ext_regi,emiMktExt,target_type_47,emi_type_47), !!calculated sets that depends on data parameter
+    loop(regi$regi_groupExt(ext_regi,regi),
+      regiEmiMktTarget2regi_47(ext_regi2,regi) = NO;
+      regiEmiMktTarget2regi_47(ext_regi,regi) = YES;
```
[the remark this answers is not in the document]
Renato-Rodrigues: Let me try to explain with an example:

We want to set a emission target at EU level, and another one at German level. 
To do that we can set `cm_emiMktTarget` to `2020.2050.EU27_regi.all.year.netGHG 0.01, 2020.2045.DEU.all.year.netGHG 0.01`

In the code, we want the `DEU` target to be responsible to determine the German carbon tax, and the `EU27_regi` target to set the price for all other EU regions, except for Germany.

This code is designed to create a set tupple that reflects that. 
So the result in this example would be to create a set like the one bellow:
```
set regiEmiMktTarget2regi_47(ext_regi,regi) /
 EU27 (ENC,EWN,ECS,ESC,ECE,FRA,ESW)  !!EU27 without DEU because the carbon price for it should be defined by the DEU target  
 DEU DEU
/;
```
in which `ext_regi` corresponds to the dimension that you defined your emission targets, and `regi` corresponds to the regions that have their carbon price controlled by it.
 

Now lets talk about how the code above achieves that:

Firstly, we loop through all possible emissions targets set at `cm_emiMktTarget`.
However, pay attention that in the loop I am looping first in the dimension `ext_regi`. I am doing this to follow an specific order of evaluation, going through the most aggregated regions to the less aggregated ones.

Any statement in GAMs is executed following the order defined at the `GAMS entry order` list. 
As the set `ext_regi` is defined from the biggest level of regional aggregation (`GLO`), to the smallest one (model region `USA`, `DEU`, ...), as long as no item of this set is previously used in another set in the REMIND code, the `GAMS entry order` will respect the order I declared in this set.
As consequence, if I loop over `ext_regi`, we will first runt the code for `GLO`, then the aggregated regions (e.g. `EUR_regi`, `EU27_regi`, ...), and only at last the more disaggregated regions that correspond to the regions actually used in the optimization code (`USA`, `DEU`, and so on).

So, back to our example, we are going to firstly run the loop for `EU27_regi` and only after to `DEU`.

The inner loop (`loop(regi$regi_groupExt(ext_regi,regi),`) is responsible to translate `ext_regi` into the model used regions (`regi`). So when you are running the code for `EU27_regi`, you are actually running it for every region contained in `EU27_regi` (`ENC,EWN,ECS,ESC,ECE,FRA,ESW,DEU`). When you run the code for a model region, e.g. `DEU`, you will only run it for `DEU`.

As we are executing in the example above the aggregated region first, when we run `EU27_regi` the statement `regiEmiMktTarget2regi_47(ext_regi2,regi) = NO;` will do nothing, but the second statement `regiEmiMktTarget2regi_47(ext_regi,regi) = YES;` will include all regions inside `EU27_regi` to the calculated set.
The result after running only the `EU27_regi` will be:

```
set regiEmiMktTarget2regi_47(ext_regi,regi) /
 EU27 (ENC,EWN,ECS,ESC,ECE,FRA,ESW,DEU)
/;
```

But as we said before, we don't want to have `DEU` carbon prices controlled by the `EU27_regi` target. 

So let's keep going in the loop.

The next item in the loop will be to run it for the non aggregated region `DEU`.

When we run the same code for this region, the first code line will remove `DEU` from any other possible item in the set, in this case remove  `DEU` from the `EU27_regi` item above, meanwhile the second statement will add a new entry `DEU DEU`, to reflect that targets that use DEU should control the German carbon price.

As as result, we will have:
```
set regiEmiMktTarget2regi_47(ext_regi,regi) /
 EU27 (ENC,EWN,ECS,ESC,ECE,FRA,ESW)  !!EU27 without DEU because the carbon price for it should be defined by the DEU target  
 DEU DEU
/;
```
which is exactly what we wanted in the first place.

Now in the carbon price update code I have a set that determines exactly which regions should have their carbon prices updated for a given target, without intersections anymore as it was happening before.

modules/47_regipol/regiCarbonPrice/postsolve.gms
Renato-Rodrigues: The `p47_targetConverged` parameter is reflecting if all targets set to a given year converged. If this happens I can move on to deal with the next period target instead in the iteration.

For example, if I set two parallel targets in 2030, one for the ETS and one for the ESR, if only the 2030 ETS target converged in this iteration, I still need to run in this iteration the rescaling tax algorithm to the 2030 prices, as the ESR target did not converged. 
I will only move on to the next year target (e.g. 2050), if all targets set to 2030 are converging in the given iteration.

modules/47_regipol/regiCarbonPrice/datainput.gms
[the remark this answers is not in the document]
robertpietzcker: wow, impressive, and thanks for the detailed explanation!

a) I guess this would be a perfect section for a description/tutorial page for regipol. Do we already have such a document for regipol where we can first compile such information snippets and then later bring it into a nice structure?

b) I think a line mentioning at the beginning of these 8 lines code that the following code makes sure only the smallest (in a regional sense) emission target applicable to a region is applied in the model, and that it relies on the set  `ext_regi` to always be defined in such a way that subsets of regions are defined AFTER their supersets would be helpful in preventing future modelers to not understand it/destroy it by accident. 
Similarly, such a heads-up in core/sets would be good :-)

modules/47_regipol/regiCarbonPrice/postsolve.gms
robertpietzcker: ok, and if one of the targets is already achieved, the value in `pm_factorRescaleemiMktCO2Tax(ttot,ttot2,ext_regi,emiMktExt)`  for this emiMktExt will be close to 1, so 
 `pm_taxemiMkt(ttot2,regi,emiMkt) = pm_taxemiMkt(ttot2,regi,emiMkt) * pm_factorRescaleemiMktCO2Tax(ttot,ttot2,ext_regi,emiMktExt);` won't change it anymore. 
 understood.

modules/47_regipol/regiCarbonPrice/datainput.gms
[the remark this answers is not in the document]
robertpietzcker: and one needs to keep in mind that targets for subset of regions should be tighter than the overall target. 
So if we set the ECEMF USA runs, we should NOT add an US target as it would be weaker than the global target, and that would (at least in my understanding) not be the aim of the protocol.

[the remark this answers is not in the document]
Renato-Rodrigues: Nice idea of having a more detailed description/tutorial section, but I am not sure the format that I should follow for that. I should get in contact with RSE to clarify that first.

I will also go through your suggestions for better comments later in the afternoon so I can add them to this PR.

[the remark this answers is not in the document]
Renato-Rodrigues: > and one needs to keep in mind that targets for subset of regions should be tighter than the overall target. So if we set the ECEMF USA runs, we should NOT add an US target as it would be weaker than the global target, and that would (at least in my understanding) not be the aim of the protocol.

You can actually add a less stringent target for an specific region like USA. As long as other regions have a way to compensate that to not create infeasibilities, the results will reflect that your USA carbon prices are lower than the other regions carbon pricing, which could be the desired effect for some specific scenario.

modules/47_regipol/regiCarbonPrice/postsolve.gms
Renato-Rodrigues: yes, the value will be still updated in the ETS in the example I gave above, but as the deviation is small, the rescale factor will not fluctuate much from 1. 

In the corner case that the model has a discontinuity in this point and an small change to the ETS carbon price causes the ETS target to not converge anymore, the same code lines in the next iteration will tell that the model that ETS target needs to be updated for 2030. 

Only when all targets are converging in a given period that I move to the next one, 2050 in the example. 

Going further, if after updating the 2050 target in a previous iteration any of my previous targets do not converge anymore in the current iteration (e.g. 2030 ESR or ETS previously converged but not anymore because 2050 prices are lower), in this iteration I will keep the updated 2050 prices unchanged and will get back to solve the 2030 targets convergence assuming the new situation of decreased long term prices (2050 prices).

modules/47_regipol/regiCarbonPrice/datainput.gms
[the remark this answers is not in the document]
Renato-Rodrigues: addressed at https://github.com/remindmodel/remind/pull/1315/commits/cb7953ffda7c42ac9699f3aeb77b95f0f790ede3

modules/47_regipol/regiCarbonPrice/sets.gms
Renato-Rodrigues: addressed at https://github.com/remindmodel/remind/pull/1315/commits/cb7953ffda7c42ac9699f3aeb77b95f0f790ede3

modules/47_regipol/regiCarbonPrice/declarations.gms
robertpietzcker: thanks!

PR #1321, opened 2023-05-31
differentiate gdx_ref and gdx_refprices

orichters:
## Purpose of this PR

- see https://github.com/pik-piam/remind2/pull/410
- the reference runs for the prices (< cm_startyear) and policy costs are different

## Type of change

- [x] Bug fix 

## Checklist:

### DIFF
--- a/scripts/output/single/reporting.R
+++ b/scripts/output/single/reporting.R
@@ -14,7 +14,8 @@ library(edgeTransport)
 library(quitte)
 ############################# BASIC CONFIGURATION #############################
 gdx_name     <- "fulldata.gdx"             # name of the gdx
-gdx_ref_name <- "input_refpolicycost.gdx"  # name of the reference gdx (for policy cost calculation)
+gdx_ref_name <- "input_ref.gdx"            # name of the ref for < cm_startyear
+gdx_refpolicycost_name <- "input_refpolicycost.gdx"  # name of the reference gdx (for policy cost calculation)
 
 
 if(!exists("source_include")) {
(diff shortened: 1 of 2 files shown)


scripts/output/single/reporting.R
```diff
@@ -14,22 +14,25 @@ library(edgeTransport)
...
 if(!exists("source_include")) {
    # Define arguments that can be read from command line
    outputdir <- "."
-   readArgs("outputdir", "gdx_name", "gdx_ref_name")
+   readArgs("outputdir", "gdx_name", "gdx_ref_name", "gdx_refpolicycost_name")
 }
 
-gdx      <- file.path(outputdir,gdx_name)
-gdx_ref  <- file.path(outputdir,gdx_ref_name)
-if (!file.exists(gdx_ref)) { gdx_ref <- NULL }
```
LaviniaBaumstark, line 28: there is no need for running new REMIND version with old reporting, right? Would be avoided by the package version anyway but I also do not see the need for it.

LaviniaBaumstark (approved):

scripts/output/single/reporting.R
orichters: > "I also do not see the need for it". 

For what? The check whether the file exists is just that the reporting does not crash if you don't have `path_gdx_ref` specified. Then, you should pass `gdx_ref = NULL` to remind2 in order for everything to work well.

LaviniaBaumstark: for running new REMIND versions with old reporting

orichters: That is indeed not supported, but I absolutely see no reason for doing that. I also would not really know how to implement that, you would somehow need to check which remind2 version is used beforehand. That is why I added the dependency on new remind2 into the DESCRIPTION and I think that is sufficient.

LaviniaBaumstark: I agree

PR #1322, opened 2023-06-02
enable coupling via renv

pfuehrlich-pik:
## Purpose of this PR
Enables the remind magpie coupling to be run using renv instead of snapshots

## Type of change

- [ ] Bug fix 
- [ ] Refactoring
- [x] New feature 
- [ ] Fundamental change

## Checklist:

* Test runs are here: 


start_bundle_coupled.R
```diff
@@ -613,8 +618,6 @@ for (scen in common) {
 }
 
 if (! "--test" %in% flags && ! "--gamscompile" %in% flags) {
-  system(paste("cp", file.path(path_remind, ".Rprofile "), file.path(path_magpie, ".Rprofile")))
```
orichters, line 616: Should be kept but only executed if `is.null(renv::project())` in order to keep snapshots running.

start_bundle_coupled.R
```diff
@@ -152,9 +158,8 @@ message("run_compareScenarios:  ", run_compareScenarios)
 if (! file.exists("output")) dir.create("output")
 
 # Check if dependencies for a REMIND model run are fulfilled
-# Use ensureRequirementsInstalled(rerunPrompt="start_bundle_coupled.R") when coupled runs are using renv.
```
orichters, line 155: I would maybe make that conditional on `is.null(renv::project())`. Or are snapshots discouraged in the sense "don't use that at all"? Then, line 8 should not be a warning, but a `stop`

orichters (commented):
Some first comments…

We should decide whether snapshots should still be an option to be used, or whether we kick them out completely. Given that it seems to require just some few `if` statements to keep this possibility for a while, we should maybe do that? Or at least ask in the REMIND meeting if everyone is fine with renv. Personally, I probably will never go back to snapshots if I can avoid it.

If we kick snapshots out completely, also the `.snapshot.Rprofile` can be deleted.

Tutorials still need to be updated.

pfuehrlich-pik:
I agree we should keep snapshots for now, I'll adjust as you suggested. Thanks for the review!

orichters (approved):
Assuming you have tested that and it works, I'm fine with the changes you made.

MAgPIE seems to take the `cfg_mag$coupling_renv_lock` and base its renv on that. So if I start a run, the new renv will be used by MAgPIE, it seems. That all seems very plausible ;)

pfuehrlich-pik:
I updated the renv tutorial. With "first MAgPIE version that supports it" you mean a release, right?

orichters:
Can you update [Tutorial 04 on REMIND-MAgPIE](https://github.com/remindmodel/remind/blob/develop/tutorials/04_RunningREMINDandMAgPIE.md#create-snapshot-of-r-libraries), please? I think, basically deleting the two paragraphs will do it, or just stating that `renv` will automatically detect the dependencies of MAgPIE and REMIND and install the necessary packages, or something like that.

orichters:
> I updated the renv tutorial. With "first MAgPIE version that supports it" you mean a release, right?

Release or commits that I could cherry-pick, if that is possible.

dklein-pik (approved):

PR #1336, opened 2023-06-15
add BECCS qtty target to regipol

orichters:
## Purpose of this PR

- expand the implicit quantity target calculation in regipol such that CCS and BECCS can be differentiated
- You can enter your target in Mt CO2 in `cm_implicitQttyTarget`
- Renato told me that adding too many time-steps in the switch may lead to strange behavior and bad convergence, but dragging down at both ends with values a bit below the actual goals worked very well.
- the run below is with `2060.GLO.tax.t.CCS.biomass 4300, 2080.GLO.tax.t.CCS.biomass 4300`, and it nailed it quite well, I think
image

- the run below is with `2060.GLO.tax.t.CCS.biomass 3800, 2080.GLO.tax.t.CCS.biomass 3800` and that also looks very convincing.
image

## Type of change

- [x] New feature 

## Checklist:

- [x] I did not adjust the reporting in [`remind2`](https://github.com/pik-piam/remind2) as it was not necessary

### DIFF
--- a/modules/47_regipol/regiCarbonPrice/postsolve.gms
+++ b/modules/47_regipol/regiCarbonPrice/postsolve.gms
@@ -363,45 +363,59 @@ $ENDIF.emiMkt
 $ifthen.cm_implicitQttyTarget not "%cm_implicitQttyTarget%" == "off"
 
 *** saving previous iteration value for implicit tax revenue recycling
+*** the same line exists in presolve.gms, don't forget to update there
 p47_implicitQttyTargetTax_prevIter(t,regi,qttyTarget,qttyTargetGroup) = p47_implicitQttyTargetTax(t,regi,qttyTarget,qttyTargetGroup);
 p47_implicitQttyTargetTax0(t,regi) =
   sum((qttyTarget,qttyTargetGroup)$p47_implicitQttyTargetTax(t,regi,qttyTarget,qttyTargetGroup),
-    ( p47_implicitQttyTargetTax(t,regi,qttyTarget,qttyTargetGroup) * sum(entyPe$energyQttyTargetANDGroup2enty(qttyTarget,qttyTargetGroup,entyPe), sum(pe2se(entyPe,entySe,te), vm_demPe.l(t,regi,entyPe,entySe,te))) 
-    )$(sameas(qttyTarget,"PE")) 
-    +
-    ( p47_implicitQttyTargetTax(t,regi,qttyTarget,qttyTargetGroup) * sum(entySe$energyQttyTargetANDGroup2enty(qttyTarget,qttyTargetGroup,entySe), sum(se2fe(entySe,entyFe,te), vm_demSe.l(t,regi,entySe,entyFe,te))) 
-    )$(sameas(qttyTarget,"SE")) 
-    +
-    ( p47_implicitQttyTargetTax(t,regi,qttyTarget,qttyTargetGroup) * sum(entySe$energyQttyTargetANDGroup2enty("FE",qttyTargetGroup,entySe), sum(se2fe(entySe,entyFe,te), sum((sector,emiMkt)$(entyFe2Sector(entyFe,sector) AND sector2emiMkt(sector,emiMkt)), vm_demFeSector.l(t,regi,entySe,entyFe,sector,emiMkt)))) 
-    )$(sameas(qttyTarget,"FE") or sameas(qttyTarget,"FE_wo_b") or sameas(qttyTarget,"FE_wo_n_e") or sameas(qttyTarget,"FE_wo_b_wo_n_e"))
-    +
-    ( p47_implicitQttyTargetTax(t,regi,qttyTarget,qttyTargetGroup) * sum(ccs2te(ccsCO2(enty),enty2,te), sum(teCCS2rlf(te,rlf),vm_co2CCS.l(t,regi,enty,enty2,te,rlf)))
-    )$(sameas(qttyTarget,"CCS"))    
+    p47_implicitQttyTargetTax(t,regi,qttyTarget,qttyTargetGroup) * (
+      ( sum(entyPe$energyQttyTargetANDGroup2enty(qttyTarget,qttyTargetGroup,entyPe), sum(pe2se(entyPe,entySe,te), vm_demPe.l(t,regi,entyPe,entySe,te)))
+      )$(sameas(qttyTarget,"PE"))
+      +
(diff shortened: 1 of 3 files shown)


modules/47_regipol/regiCarbonPrice/postsolve.gms
```diff
@@ -455,9 +469,9 @@ loop((ttot,ext_regi,taxType,targetType,qttyTarget,qttyTargetGroup)$pm_implicitQt
 loop((ttot,ext_regi,taxType,targetType,qttyTarget,qttyTargetGroup)$pm_implicitQttyTarget(ttot,ext_regi,taxType,targetType,qttyTarget,qttyTargetGroup),
   if(sameas(taxType,"tax"),
     if(iteration.val lt 15,
-      p47_implicitQttyTargetTaxRescale(ttot,ext_regi,qttyTarget,qttyTargetGroup) = (1 + pm_implicitQttyTarget_dev(ttot,ext_regi,qttyTarget,qttyTargetGroup) ) ** 4;
```
orichters, line 458: I did that because I got some `x**y` error when running it. And replacing `** 2` with `sqr()`  and `** 4` with sqr(sqr())` should avoid that.

modules/47_regipol/regiCarbonPrice/postsolve.gms
```diff
@@ -363,45 +363,59 @@ $ENDIF.emiMkt
...
-    +
-    ( p47_implicitQttyTargetTax(t,regi,qttyTarget,qttyTargetGroup) * sum(entySe$energyQttyTargetANDGroup2enty(qttyTarget,qttyTargetGroup,entySe), sum(se2fe(entySe,entyFe,te), vm_demSe.l(t,regi,entySe,entyFe,te))) 
-    )$(sameas(qttyTarget,"SE")) 
-    +
-    ( p47_implicitQttyTargetTax(t,regi,qttyTarget,qttyTargetGroup) * sum(entySe$energyQttyTargetANDGroup2enty("FE",qttyTargetGroup,entySe), sum(se2fe(entySe,entyFe,te), sum((sector,emiMkt)$(entyFe2Sector(entyFe,sector) AND sector2emiMkt(sector,emiMkt)), vm_demFeSector.l(t,regi,entySe,entyFe,sector,emiMkt)))) 
-    )$(sameas(qttyTarget,"FE") or sameas(qttyTarget,"FE_wo_b") or sameas(qttyTarget,"FE_wo_n_e") or sameas(qttyTarget,"FE_wo_b_wo_n_e"))
-    +
-    ( p47_implicitQttyTargetTax(t,regi,qttyTarget,qttyTargetGroup) * sum(ccs2te(ccsCO2(enty),enty2,te), sum(teCCS2rlf(te,rlf),vm_co2CCS.l(t,regi,enty,enty2,te,rlf)))
-    )$(sameas(qttyTarget,"CCS"))    
+    p47_implicitQttyTargetTax(t,regi,qttyTarget,qttyTargetGroup) * (
```
orichters, line 370: I moved `p47_implicitQttyTargetTax(t,regi,qttyTarget,qttyTargetGroup) *` out of the individual calculation.

modules/47_regipol/regiCarbonPrice/postsolve.gms
Renato-Rodrigues: could you do a test to see if you avoid the issue by using instead `exp(log(a) * b) = a ** b`?
e.g. (non tested code:)
```
p47_implicitQttyTargetTaxRescale(ttot,ext_regi,qttyTarget,qttyTargetGroup) = exp(log(1 + pm_implicitQttyTarget_dev(ttot,ext_regi,qttyTarget,qttyTargetGroup)) * 4 );
```
According tests made by Michaja some time ago, this was a better approach for doing exponentials in GAMs with smaller values when a similar issue happened in the calibration:
https://github.com/remindmodel/remind/pull/970

And this formulation is more flexible for which exponent we choose to use in the future if we find the need to change the pace of the convergence update.

modules/47_regipol/regiCarbonPrice/postsolve.gms
Renato-Rodrigues: Thks! 
Now to make the code less repetitive we only need to move all the contents inside the parenthesis to its own parameter value, so we don't need to repeat this full block of code in three different places, but this does not need to be done in this PR.

modules/47_regipol/regiCarbonPrice/postsolve.gms
orichters: The problem was that I got some a < 0 error. I think that was because I messed up the unit in the switch (Mt vs. Gt), and therefore the scripts thought it was very far off. So that would not be solved with `log(a)`.

Renato-Rodrigues: could you use instead of two `sqr`, instead the power function: `power(a,b)`
`p47_implicitQttyTargetTaxRescale(ttot,ext_regi,qttyTarget,qttyTargetGroup) = power(1 + pm_implicitQttyTarget_dev(ttot,ext_regi,qttyTarget,qttyTargetGroup),4);`

orichters: done.

orichters:
Implemented the change you requested, compiled and it looks good.

Renato-Rodrigues (approved):

modules/47_regipol/regiCarbonPrice/postsolve.gms
```diff
@@ -363,45 +363,59 @@ $ENDIF.emiMkt
...
+      )$(sameas(qttyTarget,"SE"))
+      +
+      ( sum(entySe$energyQttyTargetANDGroup2enty("FE",qttyTargetGroup,entySe), sum(se2fe(entySe,entyFe,te), sum((sector,emiMkt)$(entyFe2Sector(entyFe,sector) AND sector2emiMkt(sector,emiMkt)), vm_demFeSector.l(t,regi,entySe,entyFe,sector,emiMkt))))
+      )$(sameas(qttyTarget,"FE") or sameas(qttyTarget,"FE_wo_b") or sameas(qttyTarget,"FE_wo_n_e") or sameas(qttyTarget,"FE_wo_b_wo_n_e"))
+      +
+      ( sum(ccs2te(ccsCO2(enty),enty2,te), sum(teCCS2rlf(te,rlf),vm_co2CCS.l(t,regi,enty,enty2,te,rlf)))
+      )$(sameas(qttyTarget,"CCS") AND sameas(qttyTargetGroup,"all"))
+      +
+      (( !! Supply side BECCS
+        sum(emiBECCS2te(enty,enty2,te,enty3),vm_emiTeDetail.l(t,regi,enty,enty2,te,enty3))
```
Renato-Rodrigues, line 384: you need to remove the `.l` here so the model knows there is a trade-off between the tax and the level of `vm_emiTeDetail` here.

PR #1340, opened 2023-06-19
run all output scripts on one slurm, improve logging

orichters:
## Purpose of this PR

- add `--vanilla` to `Rscript` calls within GAMS to suppress hundreds of `Global .Rprofile loaded! (R version 4.1.2 (2021-11-01))` lines in the logs.
- In `output.R`, if you select `single` but then choose more than one output script (say: `reporting` + `rds_report` + `emulator` for a coupled run), they were all run on separate `slurm` jobs, sometimes interfering with each other, or one script needed the output of the other, which creates a big mess. I propose to run each `outputdir` in one slurm job, but all the scripts selected one after the other in this one job. This allows you to actually select a chain of output scripts that need each other, instead of having them interfere with each other.
- in `scripts/input/exoGAINSAirpollutants.R`, improve the logging which file is used.
- in `start.R`, also look into `config` if you find the config file there. I struggled quite a bit with the warning that the file could not be found, although it was clearly in the config folder, except I forgot to add the `config/` part in the command line. Make sure this does not happen anymore to anyone.

## Type of change

- [x] Bug fix
- [x] Refactoring

## Checklist:

- [x] I did not adjust the reporting in [`remind2`](https://github.com/pik-piam/remind2)

### DIFF
--- a/modules/11_aerosols/exoGAINS/postsolve.gms
+++ b/modules/11_aerosols/exoGAINS/postsolve.gms
@@ -19,7 +19,7 @@ if((o_modelstat le 2),
 );
 
 *** Calculate AP emissions
-Execute "Rscript exoGAINSAirpollutants.R";
+Execute "Rscript --vanilla exoGAINSAirpollutants.R";
 
 *** Read input ref results for tall with following dimensions: pm_emiAPexsolve(tall,all_regi,all_sectorEmi,emiRCP)
 if((cm_startyear gt 2005),
--- a/output.R
+++ b/output.R
@@ -199,7 +199,7 @@ if (comp %in% c("comparison", "export")) {
   if (! exists("source_include")) {
     # for selected output scripts, only slurm configurations matching these regex are available
     slurmExceptions <- if ("reporting" %in% output) "--mem=[0-9]*[0-9]{3}" else NULL
-    if (all(output %in% outputUsingDirect)) slurmConfig <- "direct"
+    if (any(output %in% outputUsingDirect)) slurmConfig <- "direct"
     # if this script is not being sourced by another script but called from the command line via Rscript let the user
     # choose the slurm options
     if (!exists("slurmConfig")) {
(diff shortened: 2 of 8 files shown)


output.R
```diff
@@ -257,38 +257,38 @@ if (comp %in% c("comparison", "export")) {
...
+          slurmcmd <- paste0("sbatch ", slurmConfig, " --job-name=", logfile, " --output=", logfile,
+                       " --mail-type=END --comment=output.R --wrap='", Rscripts, "'")
+          message("Sending to slurm: ", paste(name, collapse = ", "), ". Find log in ", logfile)
+          system(slurmcmd)
         }
+        # finished
+        message("\nFinished ", ifelse(slurmConfig == "direct", "", "starting job for "), "output generation for ", outputdir, "!\n")
+      } else {
+        warning("Skipping ", outputdir, " because some output script selected could not be found ",
+                "in scripts/output/single: ", name[! name %in% dir("scripts/output/single")])
```
LaviniaBaumstark, line 290: I like your proposal to run more than one R-scripts in one slurm job to avoid mess

LaviniaBaumstark (approved):
thanks for all the cleaning

modules/11_aerosols/exoGAINS/postsolve.gms
```diff
@@ -19,7 +19,7 @@ if((o_modelstat le 2),
 );
 
 *** Calculate AP emissions
-Execute "Rscript exoGAINSAirpollutants.R";
+Execute "Rscript --vanilla exoGAINSAirpollutants.R";
```
0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q, line 22: I'll state to obvious (at least obvious to anyone who read the `R` man page): 
this deactivates the renv, making consistent results from REMIND runs a gamble once the versions of the packages required by `ecoGAINSAirpollutants.R` diverge between the renv library and whatever happens to be in `R_LIBS_SITE`.

PR #1342, opened 2023-06-22
Add 45/NPi realization to calibrate on

orichters:
## Purpose of this PR
Switch to NPi calibration instead of baseline calibration. Therefore, the new realization `NPi` of the module `carbonprice` is developed. Use cm_startyear instead of cm_emiscen for if-conditions where needed. New input data version 6.542 include historic carbon tax info also for 2020 and REMIND is calibrated to the NPi scenario (new CES parameters for SSP2EU, SSP2EU-EU21, SSP1, SSP5, SDP_MC, SDP_EI, SDP_RC). Git managed scenario_config*.csv files are adjusted accordingly. In future, first a NPi scenario is run and based on this NDC, pkBudg* scenarios are generated.

## Type of change

- [x] Bug fix 
- [x] New feature 

## Checklist:

- [x] All automated model tests pass (`FAIL 0` in the output of `make test`), except for scenario_config_GCS.csv and scenario_config_coupled_GCS.csv, which need to be updated by @gabriel-abrahao 

* Test runs are here: 


robertpietzcker:
Just to understand, as I was to busy in the last week to follow it closely: is this already the step "making this default"?

if yes, could you add the path to the results and compScens for old/new calibration and policy results (also for EU21) so I can have a deeper look?

also, were there already any tests if this also works with differentiated ETS/ESR targets for the follow-up runs?

modules/37_industry/subsectors/bounds.gms
```diff
@@ -21,7 +21,7 @@ vm_cesIO.up(ttot,regi,"ue_steel_secondary")
       p37_cesIO_up_steel_secondary(ttot,regi,"%cm_GDPscen%")
     );
 $elseif.secondary_steel_bound "%cm_secondary_steel_bound%" == "scenario"
-if (1 eq cm_emiscen,
+if (cm_startyear eq 2005,
```
0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q, line 24: These changes are not equivalent.  Not all policy scenarios have to start later then 2005.  Not all scenarios that are fixed after 2005 are baseline scenarios.

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q (changes requested):

modules/37_industry/subsectors/bounds.gms
LaviniaBaumstark: right @robertpietzcker, it is more the other way around: from now on the first scenarios of a cascade will not be baselines anymore but can be a NPi scenario

robertpietzcker: right, my question was meant to be addressed to @0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q to make sure this is a piece of code he doesn't want executed in REMIND anymore (or only for very specific "how would a baseline look like" experiments)

gabriel-abrahao:
Just fixed the outdated switches in the GCS configs, `test_01-gms_readSettings.R` passes now.

modules/37_industry/subsectors/bounds.gms
0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q: https://github.com/remindmodel/remind/blob/af701bda24f0e0c1eafe5281a3fb299ebad2c4a8/modules/37_industry/subsectors/bounds.gms#L24-L28

If `NPi` is to be the new baseline and calibrated to `mrremind` output, then this upper bound should apply to it.

> I think the reason for this replacement was that in many places in the code, the "emiscen = 1" was used to say "the first REMIND run in a cascade", so it was widely replaced throughout the code.

Well, in this case, `cm_emiscen = 1` was used to say "if this is not a policy scenario".

robertpietzcker: ok, based on your first sentence the pull request is indeed right, as it makes NPi the scenario that is calibrated to the mrremind output

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q: > ok, based on your first sentence the pull request is indeed right

Incidentally so.  `cm_startyear eq 2005` is not the same as "if this is not a policy scenario".  I would suggest to avoid the
>  in many places in the code, the _"X"_ was used to say _"Y"_

situation and be explicit about what is meant.

Making `NPi` the "baseline" and `Base` a counter-factual renders `cm_emiscen` unsuited for this task.  But then we should have an explicit parameter to distinguish whether something is a baseline or climate policy scenario and which assumptions and parameters should apply to it.

robertpietzcker: "cm_calibrationRun"? "cm_firstRunInCascade"? "cm_takesFEpathsFromMrremind"? @LaviniaBaumstark

LaviniaBaumstark: "cm_firstRunInCascade" fits best what we mean with it as it is not only relevant for calibration scenarios

LaviniaBaumstark: but isn't this also the same as start_year = 2005 ?

LaviniaBaumstark: regarding adding a new switch called "cm_firstRunInCascade": if really necessary, sure; if not I would like to avoid adding a new switch would need to be added to each scenario_config file

modules/37_industry/subsectors/bounds.gms
```diff
@@ -35,12 +35,54 @@ if (cm_startyear eq 2005,
...
+      + 1
+      )
+    * pm_fedemand(t,regi,"ue_steel_secondary");
+$else.rcp_scen
   !! In policy scenarios, secondary steel production can be increased up to the
   !! limit of theoretical scrap availability.
   vm_cesIO.up(t,regi,"ue_steel_secondary")
     = p37_cesIO_up_steel_secondary(t,regi,"%cm_GDPscen%");
-);
+$endif.rcp_scen
```
0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q, line 85: Really?  Copying a block of code four times instead of writing an or-statement? Wow.

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q (approved):

modules/37_industry/subsectors/bounds.gms
LaviniaBaumstark: did not manage to get ifthen statements with more than one condition compiling, feel free to improve

robertpietzcker (approved):
looks good

PR #1350, opened 2023-07-11
initialize pm_pvpRegi = 0 instead of loading those values from the input.gdx file

LaviniaBaumstark:
## Purpose of this PR
Not load in pm_pvpRegi values from input.gdx avoids CO2 prices for time steps if they are not overwritten later in code (e.g. baseline scenarios after NPi caliration and using an input.gdx from an NPi calibration run)

## Type of change

- [x] Bug fix 

## Checklist:

* Test runs are here: /p/tmp/lavinia/MO/test_pm_pvpRegi/remind

### DIFF
--- a/core/preloop.gms
+++ b/core/preloop.gms
@@ -38,13 +38,13 @@ vm_co2eqMkt.l(ttot,regi,emiMkt) = 0;
 v_shfe.l(t,regi,enty,sector) = 0;
 v_shGasLiq_fe.l(t,regi,sector) = 0;  
 pm_share_CCS_CCO2(t,regi) = 0; 
-  
+pm_pvpRegi(ttot,regi,enty) = 0;
+
 *** overwrite default targets with gdx values if wanted
 Execute_Loadpoint 'input' p_emi_budget1_gdx = sm_budgetCO2eqGlob;
 Execute_Loadpoint 'input' vm_demPe.l = vm_demPe.l;
 Execute_Loadpoint 'input' q_balPe.m = q_balPe.m;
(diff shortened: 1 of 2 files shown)


core/preloop.gms
```diff
@@ -38,13 +38,13 @@ vm_co2eqMkt.l(ttot,regi,emiMkt) = 0;
 v_shfe.l(t,regi,enty,sector) = 0;
 v_shGasLiq_fe.l(t,regi,sector) = 0;  
 pm_share_CCS_CCO2(t,regi) = 0; 
-  
+pm_pvpRegi(ttot,regi,enty) = 0;
```
robertpietzcker, line 41: shouldn't this be only for t and not ttot?
so that if the run is fixed to another run, it will have the pm_pvpRegi values for the fixed time steps?

ah, likely I am wrong - this will anyway happen later in the read-in of input_ref, right?

robertpietzcker (approved):

core/preloop.gms
LaviniaBaumstark: this will be addressed in 
(cm_startyear gt 2005),
  Execute_Loadpoint 'input_ref' p_pvpRegiBeforeStartYear = pm_pvpRegi;
   pm_pvpRegi(ttot,regi,"perm")$((ttot.val gt 2005) AND (ttot.val lt cm_startyear)) =  p_pvpRegiBeforeStartYear(ttot,regi,"perm");
 );

orichters (approved):
With respect to `enty = perm`, this looks good. I cannot say anything how this affects the other `enty` options.

mleimbach (approved):

PR #1355, opened 2023-07-12
restore cm_emiscen=10, 45_carbonprice/Npi -> NPi in configs, minor bugfixes

orichters:
## Purpose of this PR

- cm_emiscen = 10 was lost [here](https://github.com/remindmodel/remind/pull/1177/files#diff-513d9cc5f1686e16c5eb20af6707897894d0995e75b961953aed75a81bddeeea). Documentation is added in this PR that touches main.gms anyway: https://github.com/remindmodel/remind/pull/1356
- 45_carbonprice/Npi -> NPi in configs
- with `cm_INCONV_PENALTY_FESwitch` = off, the `positive variables ;` lead to compilation error

## Type of change

- [x] Bug fix 

## Checklist:

- [x] My code follows the [coding etiquette](https://github.com/remindmodel/remind/blob/develop/main.gms#L80)
- [x] I performed a self-review of my own code
- [x] I explained my changes within the PR, particularly in hard-to-understand areas
- [x] I checked that the [in-code documentation](https://github.com/remindmodel/remind/blob/develop/main.gms#L120) is up-to-date
- [x] I adjusted the reporting in [`remind2`](https://github.com/pik-piam/remind2) where it was needed
- [x] I adjusted `forbiddenColumnNames` in [readCheckScenarioConfig.R](https://github.com/remindmodel/remind/blob/develop/scripts/start/readCheckScenarioConfig.R) in case the PR leads to deprecated switches
- [x] All automated model tests pass (`FAIL 0` in the output of `make test`)

PR #1356, opened 2023-07-12
reactive check_config, add regexp check for cfg$… values

orichters:
## Purpose of this PR

- using gms::check_config was disable while moving settings to main.gms. Reenable it as `checkFixCfg()`, with reading the config from main.gms, and adding all the other changes to cfg that were in `prepare.R`
- Add a regexp check for cfg$gms$ values: In the line of `main.gms` where the parameter or setglobal value is specified, add `regexp = whatever` which is used to check the value used for this switch. This allows to catch wrong settings for parameters such as `cm_emiscen` where no other such check existed, yet. There are certainly other lines in `main.gms` where this can be added, but it is a start…
- call `checkFixCfg()` in `start.R` and `start_bundle_coupled.R` such that errors are caught early, already while running in `--test` mode
- add to the tests that all scenario_config files are run with `start.R --test` such that errors are caught in the tests already. This costs less than 4 minutes, but we still might want to decide whether that needs to be part of the standard test
- Same with `scenario_config_coupled` files and `start_bundle_coupled.R --test` which is run using `make test-coupled`, adding < 2 minutes to the 75 it had before – that should be fine.
- Compile every scenario from the config files, but only in `make test-full` mode which takes hours anyway. Takes about 75 minutes, but I think that is worth a full test.
- Improve the tests that were calling scripts that potentially had `gms::getLine` such that they fail if that happens, instead of just waiting for nothing to happen.
- add `startgroup=*` such that `selectScenarios()` starts everything

In the process, the new tests lead to finding some bugs that I fixed in https://github.com/remindmodel/remind/pull/1355

## Type of change

- [x] Refactoring
- [x] New feature 

## Checklist:

- [x] My code follows the [coding etiquette](https://github.com/remindmodel/remind/blob/develop/main.gms#L80) and extends it
- [x] adjusted the reporting in [`remind2`](https://github.com/pik-piam/remind2) was not needed
- [x] All automated model tests pass (`FAIL 0` in the output of `make test`, `make test-coupled`, `make test-full`)


Makefile
```diff
@@ -59,6 +59,10 @@ test-coupled:    ## Test if the coupling with MAgPIE works. Takes significantly
 	$(info Coupling tests take around 75 minutes to run, please be patient)
 	@R_PROFILE_USER= TESTTHAT_RUN_SLOW=TRUE Rscript -e 'testthat::test_file("tests/testthat/test_20-coupled.R")'
 
+test-coupled-slurm: ## test-coupled, but on slurm
```
fbenke-pik, line 62: Shouldn't test-coupled be removed from make commands so you can only run coupled test with SLURM?

orichters: I don't think so. The model runs are on slurm anyway, the difference is whether also the test itself is on slurm. I think normally you should not do that (because it reserves a full CPU basically for waiting), but sometimes if the internet connection is unstable (train…) it is a helpful option because then the tests do not fail just because your console lost connection.

fbenke-pik (approved):

dklein-pik (approved):
Great, thanks for adding the comments. That helps. Could you add one ore two sentences at the top of the function that explain what the funtion does and what it is used for in general? Thank you.

PR #1371, opened 2023-08-02
suppress madrat standard config for start.R

orichters:
## Purpose of this PR

1. Adding a test that I forgot in #1370

2. The first scenario to be started with `./start.R` always looks like that:

```
testOneRegi-Base
   Configuring cfg for testOneRegi-Base
   SLURM option 8 selected: qos=priority nodes=1 tasks-per-node=1 mem=8000
   Run can be started using 0 specified gdx file(s).

Initialize madrat config with default settings..
    regionmapping = regionmappingH12.csv
    extramappings = NULL
    packages = madrat
    globalenv = FALSE
    verbosity = 1
    mainfolder = /p/projects/rd3mod/inputdata
    sourcefolder = NA
    cachefolder = NA
    mappingfolder = NA
    outputfolder = NA
    pucfolder = NA
    tmpfolder = NA
    nolabels = NULL
    forcecache = FALSE
    ignorecache = NULL
    cachecompression = gzip
    hash = xxhash32
    diagnostics = FALSE
    debug = FALSE
    maxLengthLogMessage = 200
..done!

   Writing cfg to file testOneRegi-Base.RData
   If this wasn't --test mode, I would submit testOneRegi-Base.
```

I don't think that the madrat stuff contains any useful information, given it just reproduces [these lines](https://github.com/pik-piam/madrat/blob/master/R/initializeConfig.R#L16-L35) from `madrat::initializeConfig()`.

This is avoided by just calling `madrat::getConfig(verbose = FALSE)` before starting any scenarios.

## Type of change

- [x] Cleanup

## Checklist:

- [x] I don't think that needs to be mentioned in the changelog.

### DIFF
--- a/scripts/start/readCheckScenarioConfig.R
+++ b/scripts/start/readCheckScenarioConfig.R
@@ -36,7 +36,7 @@ readCheckScenarioConfig <- function(filename, remindPath = ".", testmode = FALSE
   }
   nameisNA <- grepl("^NA$", rownames(scenConf))
   if (any(nameisNA)) {
-    warning("Don't use 'NA' as scenario name, you fool. Stopping now.")
+    warning("Do not use 'NA' as scenario name, you fool. Stopping now.")
   }
   illegalchars <- grepl("[^[:alnum:]_-]", rownames(scenConf))
   if (any(illegalchars)) {
--- a/start.R
+++ b/start.R
@@ -110,6 +110,9 @@ if (   'TRUE' != Sys.getenv('ignoreRenvUpdates')
   Sys.sleep(1)
 }
 
+# initialize madrat settings
+invisible(madrat::getConfig(verbose = FALSE))
+
 errorsfound <- 0 # counts ignored errors in --test mode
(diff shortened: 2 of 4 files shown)


scripts/start/readCheckScenarioConfig.R
```diff
@@ -36,7 +36,7 @@ readCheckScenarioConfig <- function(filename, remindPath = ".", testmode = FALSE
   }
   nameisNA <- grepl("^NA$", rownames(scenConf))
   if (any(nameisNA)) {
-    warning("Don't use 'NA' as scenario name, you fool. Stopping now.")
+    warning("Do not use 'NA' as scenario name, you fool. Stopping now.")
   }
```
0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q, line 40: Forgoing the contraction "don't", while still calling the user a fool, seems like an odd stylistic break. ;)

pfuehrlich-pik (approved):

PR #1382, opened 2023-08-21
in start_bundle_coupled fail on missing files/dirs

orichters:
## Purpose of this PR

- If you somehow misspecified the files / folders, `start_bundle_coupled` broke quite late with unclear message. Rather fail early.
- in `output.R`, if no folder contains gdx files (and mif if needed), fail and tell the user instead of printing this nonsense:

```
Please choose runs to be used for output generation:

1,a: all
2,p: Search pattern by regular expression...
3,f: Search by fixed pattern...
```

## Type of change

- [x] Bug fix 

## Checklist:

- [x] All automated model tests pass (`FAIL 0` in the output of `make test-coupled-slurm`)

### DIFF
--- a/start_bundle_coupled.R
+++ b/start_bundle_coupled.R
@@ -155,16 +155,6 @@ if (length(argv) > 0) {
   message("")
 }
 
-message("path_remind:           ", if (dir.exists(path_remind)) green else red, path_remind, NC)
-message("path_magpie:           ", if (dir.exists(path_magpie)) green else red, path_magpie, NC)
-message("path_settings_coupled: ", if (file.exists(path_settings_coupled)) green else red, path_settings_coupled, NC)
-message("path_settings_remind:  ", if (file.exists(path_settings_remind)) green else red, path_settings_remind, NC)
-message("path_remind_oldruns:   ", if (dir.exists(path_remind_oldruns)) green else red, path_remind_oldruns, NC)
-message("path_magpie_oldruns:   ", if (dir.exists(path_magpie_oldruns)) green else red, path_magpie_oldruns, NC)
-message("prefix_runname:        ", prefix_runname)
-message("n600_iterations:       ", n600_iterations)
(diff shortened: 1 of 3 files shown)


start_bundle_coupled.R
```diff
@@ -178,6 +168,21 @@ deletedFolders <- 0
...
+message("path_magpie_oldruns:   ", if (dir.exists(path_magpie_oldruns)) green else red, path_magpie_oldruns, NC)
+message("prefix_runname:        ", prefix_runname)
+message("n600_iterations:       ", n600_iterations)
+message("run_compareScenarios:  ", run_compareScenarios)
+
+if (any(! file.exists(c(path_settings_coupled, path_settings_remind))) ||
+    any(! dir.exists(c(path_remind, path_magpie, path_remind_oldruns, path_magpie_oldruns)))) {
+  stop("Missing files or directories, see in red above.")
+}
+
```
dklein-pik, line 185: Why did you move it downwards?

orichters: - It needed to be below the `dir.create("output")` to make sure `path_remind_oldruns` exists.
- I first wanted to do it via `errorsfound` and fail at the end of the file, but then I thought it is better to fail immediately. So yeah, one could move it back up some lines, but I think it does not matter.

dklein-pik: Ok, now one medal for me. It was kind of obvious that 'output' is part of ' path_remind_oldruns' .

dklein-pik (approved):

start_bundle_coupled.R
dklein-pik: Sorry, I know it is not important and we can leave it as is, but I simply can't shut up: if `output` has just been created the line above there will be nothing to be found in `path_remind_oldruns` anyway, is it?

orichters: Yes, so no run will be found which is ok, the script does not fail in this case.

PR #1394, opened 2023-09-06
Fix event for version increment action

fbenke-pik:
## Purpose of this PR

Sets event to trigger increment of dev version to `pull_request_target` instead of `pull_request` so that secrets in the REMIND repository can be accessed from within the workflow.

## Type of change

- [x] Bug fix 

## Checklist:


LaviniaBaumstark (approved):
nice

Issue #1399, opened 2023-09-06
Steel Primary and Secondary don't sum up to total for 2010 and 2015


orichters:
Shows up in checkSummation logs. How to reproduce
```
gdx <- "/p/projects/remind/modeltests/remind/output/SSP2EU-Base-AMT_2023-09-01_22.08.50/fulldata.gdx"
output <- remind2::reportFE(gdx)
prim <- output[,,"FE|Industry|Steel|++|Primary (EJ/yr)"]
sec <- output[,,"FE|Industry|Steel|++|Secondary (EJ/yr)"]
tot <- output[,,"FE|Industry|+++|Steel (EJ/yr)"]
```
Now, `abs(tot - prim - sec)` is smaller than `1e-14` except for 2010 and 2015:
```
(tot-prim-sec)[, c(2010, 2015), ]
```
```
        ttot
all_regi         y2010         y2015
     LAM -5.802134e-04  4.019396e-04
     OAS -4.167843e-05 -1.064893e-04
     SSA -2.356056e-05  6.938894e-18
     EUR -2.627722e-04  2.775558e-16
     NEU -1.409450e-05  7.801600e-06
     MEA -2.202686e-05 -1.179612e-16
     REF -1.119489e-03 -5.551115e-17
     CAZ -1.227636e-04 -6.938894e-17
     CHA -1.441973e-03 -8.960015e-05
     IND -3.563202e-05  2.665842e-05
     JPN  8.326673e-17  6.796558e-06
     USA -2.948931e-04  8.326673e-17
     GLO -3.959096e-03  2.471067e-04
```
Maybe this can be fixed. Tagging @0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q who should at least know who can fix that…

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
This is a reporting issue.  Please transfer to [pik-piam/remind2](https://github.com/pik-piam/remind2/).

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
Putting this into `remind2` might have been premature on my part.  Something is going on on the REMIND side.

orichters:
Closed, see also https://github.com/pik-piam/remind2/issues/455

PR #1404, opened 2023-09-11
make sure lower bound on vm_cesIO is at least the offset_quantity, to prevent negative net energy consumption

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
Net `vm_cesIO` (i.e. `vm_cesIO + pm_cesdata("offset_quantity")`, the latter being negative) was not limited at zero on the `pf` level, only on the `pf`-sum.  Specifically, while the sum of all hydrogen net consumption in industry would be 0 in 2010 (so equal production of 0), individual industry subsectors would consume more `feh2_` than indicated by their offset quantity (and others less), resulting in (very small) negative net consumption, which showed up when looking into a industry reporting issue (#1399).
I extended this to all lower bounds on `vm_cesIO`.  Other modules probably are not affected, but this was faster then checking all production factors individually.

- [x] Bug fix 

## Checklist:

* Test runs are here: `/p/tmp/pehl/Remind/output/SSP2EU-Base_2023-09-07_11.37.01`
* Comparison of results (what changes by this PR?): `/p/tmp/pehl/Remind/compScen-2023-09-11_12.42.07-H12.pdf`

### DIFF
--- a/modules/01_macro/singleSectorGr/bounds.gms
+++ b/modules/01_macro/singleSectorGr/bounds.gms
@@ -8,7 +8,8 @@
 
 *nb* lower bounds on CES values
 vm_cons.lo(t,regi)     = 1e-3;
-vm_cesIO.lo(t,regi,in)$( NOT in_industry_dyn37(in) ) = 1e-6;
+vm_cesIO.lo(t,regi,in)$( NOT in_industry_dyn37(in) ) 
+  = max(1e-6, abs(pm_cesdata(t,regi,in,"offset_quantity")));
 vm_cesIOdelta.lo(t,regi,in_putty) = 1e-6;
 
 *** fix energy inputs to CES structure in t0 to the parameter values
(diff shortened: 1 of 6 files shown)


robinhasse (approved):
Sounds and looks reasonable 👍

orichters (approved):
I have to admit I have not fully understood the role of the offset quantities, so I can just say that the changes look not implausible. Or rather: I trust you this fixes a problem while I can't see how this could create new ones.

mleimbach:
Makes sense for me as well.

mleimbach (approved):

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
> I have to admit I have not fully understood the role of the offset quantities

Growth of production factors is essentially multiplicative.  If a production factor is zero, it will stay zero, no matter how high one cranks the efficiency parameter of that factor.  So for factors that just are zero (hydrogen in 2005–15), we have the offset quantities, so that there is no production on the ESM side while there is some (virtual) consumption inside the CES function.

Issue #1413, opened 2023-09-18
runs restarted after preempt don't start subsequent runs


orichters:
- Happened for `/p/projects/remind/modeltests/remind/output/SSP2EU-NPi-AMT_2023-09-15_22.09.42`
- Reason: as `full.gms` already exists, [these lines in prepareAndRun.R](https://github.com/remindmodel/remind/blob/develop/scripts/start/prepareAndRun.R#L18-L29) sets `start_subsequent_runs <- FALSE`.
- this was introduced by @dklein-pik in [this commit](https://github.com/remindmodel/remind/commit/1c281a1112e0d0050cafc2387a63ada3db962bd2). I think as we now ask in `./start.R -r` whether subsequent runs can be started, we should rethink these lines again.

PR #1414, opened 2023-09-18
let preempted and resumed runs start their subsequent runs

orichters:
## Purpose of this PR

- see #1413
- it is meanwhile [asked in `start.R`](https://github.com/remindmodel/remind/blob/c2174b3b8831e087d26da0f32023850c1131b283/start.R#L151C31-L169) whether a restarted run should start its subsequent runs as well
- therefore, the automatic detection based on existence of `full.gms` is misleading, always leading to `FALSE` for preempted and resumed runs
- so always start subsequent runs here, expect for coupled runs and if it was selected in `start.R` not to
- make sure old log is saved also in case of preemption (I don't think that works the way I implemented it, as the file is probably already overwritten here)

## Type of change

- [x] Bug fix 

## Checklist:

### DIFF
--- a/scripts/start/prepareAndRun.R
+++ b/scripts/start/prepareAndRun.R
@@ -22,8 +22,6 @@ prepareAndRun <- function() {
   } else {
     # If "full.gms" exists, the script assumes that a full.gms has been generated before and you want
     # to restart REMIND in the same folder using the gdx that it eventually previously produced.
-    message("\nRestarting REMIND, find old log in 'log_beforeRestart.txt'.")
-    if (file.exists("log.txt")) file.copy("log.txt", "log_beforeRestart.txt", overwrite = TRUE)
     if (file.exists("fulldata.gdx")) file.copy("fulldata.gdx", "input.gdx", overwrite = TRUE)
   }
 
--- a/scripts/start/prepareAndRun.R
+++ b/scripts/start/prepareAndRun.R
@@ -19,17 +19,16 @@ prepareAndRun <- function() {
     # If no "full.gms" exists, the script assumes that REMIND did not run before and
     # prepares all inputs before starting the run.
     prepare()
-    start_subsequent_runs <- TRUE
   } else {
     # If "full.gms" exists, the script assumes that a full.gms has been generated before and you want
     # to restart REMIND in the same folder using the gdx that it eventually previously produced.
     message("\nRestarting REMIND, find old log in 'log_beforeRestart.txt'.")
-    if(file.exists("fulldata.gdx")) file.copy("fulldata.gdx", "input.gdx", overwrite = TRUE)
-    start_subsequent_runs <- FALSE
+    if (file.exists("log.txt")) file.copy("log.txt", "log_beforeRestart.txt", overwrite = TRUE)
+    if (file.exists("fulldata.gdx")) file.copy("fulldata.gdx", "input.gdx", overwrite = TRUE)
   }
 
   # Run REMIND, start subsequent runs (if applicable), and produce output.
-  run(start_subsequent_runs)
+  run()
 }
 
 # Only if this file is run directly via Rscript prepareAndRun.R, but not if this file
--- a/scripts/start/run.R
+++ b/scripts/start/run.R
@@ -5,7 +5,7 @@
 # |  REMIND License Exception, version 1.0 (see LICENSE file).
 # |  Contact: remind@pik-potsdam.de
 
-run <- function(start_subsequent_runs = TRUE) {
+run <- function() {
 
   load("config.Rdata")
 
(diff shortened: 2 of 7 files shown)


scripts/start/run.R
```diff
@@ -175,7 +175,7 @@ run <- function(start_subsequent_runs = TRUE) {
   # Use the name to check whether it is a coupled run (TRUE if the name ends with "-rem-xx")
   coupled_run <- grepl("-rem-[0-9]{1,2}$",cfg$title)
   # Don't start subsequent runs form here if REMIND runs coupled. They are started in start_coupled.R instead.
-  start_subsequent_runs <- (start_subsequent_runs | isTRUE(cfg$restart_subsequent_runs)) & !coupled_run
+  start_subsequent_runs <- ! isFALSE(cfg$restart_subsequent_runs) && ! coupled_run
```
dklein-pik, line 178: Do I understand it correctly that if the run is restarted after preemtion the `! ifFALSE`-expression evaluates to TRUE because `cfg$restart_subsequent_runs` is NULL because it has not been set by the user in `start.R`?

orichters: yes, if this run has not been restarted manually, it is just `NULL` and `isFALSE(NULL)` is `FALSE`, so subsequent standalone runs will be started

scripts/start/prepareAndRun.R
```diff
@@ -19,17 +19,16 @@ prepareAndRun <- function() {
...
     prepare()
-    start_subsequent_runs <- TRUE
   } else {
     # If "full.gms" exists, the script assumes that a full.gms has been generated before and you want
     # to restart REMIND in the same folder using the gdx that it eventually previously produced.
     message("\nRestarting REMIND, find old log in 'log_beforeRestart.txt'.")
-    if(file.exists("fulldata.gdx")) file.copy("fulldata.gdx", "input.gdx", overwrite = TRUE)
-    start_subsequent_runs <- FALSE
+    if (file.exists("log.txt")) file.copy("log.txt", "log_beforeRestart.txt", overwrite = TRUE)
+    if (file.exists("fulldata.gdx")) file.copy("fulldata.gdx", "input.gdx", overwrite = TRUE)
```
dklein-pik, line 27: I agree with what you write in the conversation tab: this copy command comes too late, because the `log.txt` has already been recreated in [`submit.R`](https://github.com/remindmodel/remind/blob/2758f32fe4907ad33f575f45aff497580f949012/scripts/start/submit.R#L133). Do you think it would be correct to move the copy command [here](https://github.com/remindmodel/remind/blob/2758f32fe4907ad33f575f45aff497580f949012/scripts/start/submit.R#L36)?

orichters:
I did it as suggested, and this seems to work. While requeued, `rs2` shows:
```
Folder           Runtime      inSlurm   RunType      RunStatus          Iter              Conv                   modelstat              Mif  AppResults
testOneRegi      pending      priority  quick EUR    Run in progress    1/1               NA                     NA                     no   no
```
Being `pending` while `Run in progress` is not completely counter-intuitive, but we might want to adapt that if needed.

The log states, see `less /p/tmp/oliverr/remind-smallfix/output/testOneRegi/log.txt`
```
Starting REMIND...
GAMS will provide logging in full.log.
slurmstepd: error: *** JOB 27909012 ON cs-f14c02b07 CANCELLED AT 2023-09-19T17:38:06 DUE TO JOB REQUEUE ***
Global .Rprofile loaded! (R version 4.1.2 (2021-11-01))
```

dklein-pik:
Wenn looking into the file `less /p/tmp/oliverr/remind-smallfix/output/testOneRegi/log.txt` I can't find the `slurmstep` line. Is this a bad sign?

scripts/start/submit.R
```diff
@@ -128,9 +128,13 @@ submit <- function(cfg, restart = FALSE, stopOnFolderCreateError = TRUE) {
   if (grepl("^direct", cfg$slurmConfig) || ! isSlurmAvailable()) {
     exitCode <- system("Rscript prepareAndRun.R")
   } else {
+    if (file.exists("log.txt")) {
+      message("\nRestarting REMIND, find old log in 'log_beforeRestart.txt'.")
+      file.rename("log.txt", "log_beforeRestart.R")
+    }
```
dklein-pik, line 134: I don't get why this is here: if restarted after preemption this will not be executed anyway (as you correctly corrected me ;-)), and if the run is restarted by the user the log and other files are copied [here](https://github.com/remindmodel/remind/blob/2758f32fe4907ad33f575f45aff497580f949012/start.R#L193), aren't they?

orichters: Yes, that is true.

orichters:
I added the append stuff to the coupling scripts as well. Hope I got all that need it.

orichters:
> Wenn looking into the file `less /p/tmp/oliverr/remind-smallfix/output/testOneRegi/log.txt` I can't find the `slurmstep` line. Is this a bad sign?

Can you look again? I restarted the run and therefore it went away, but now it should be back.

dklein-pik (approved):

orichters:
hab noch ein bisschen Kosmetik gemacht und die ohnehin gelöschte deepEL config bei den Tests aus den zu überspringenden rausgenommen. `make test` lief durch, also mergen wir mal

PR #1366, opened 2023-07-25
Check if input_bau.gdx is given or really needed; a couple of minor changes mostly related to MAgPIE coupling

dklein-pik:
## Purpose of this PR

This PR contains a couple of minor changes to R scripts mostly related to REMIND-MAgPIE coupling (code cleaning, adding comments, ...).

The most important change and not related to coupling is a new check in `checkFixCfg.R` that throws an error if an NDC is to be started but no `input_bau.gdx` was given. In case no NDC is to be calculated but an `input_bau.gdx` is provided it will be set to NA since it would otherwise introduce an unnecessary dependency to the preceeding BAU run. This is most relevant for the parallel mode of coupled runs, where in each iteration the unnecessary dependency may prevent the start of subseqent runs if the preceeding BAU run (that is actually not required in a non-NDC scenario) fails.

## Type of change

- [x] New feature 

## Checklist:

- [ ] My code follows the [coding etiquette](https://github.com/remindmodel/remind/blob/develop/main.gms#L80)
- [ ] I performed a self-review of my own code
- [ ] I explained my changes within the PR, particularly in hard-to-understand areas
- [ ] I checked that the [in-code documentation](https://github.com/remindmodel/remind/blob/develop/main.gms#L120) is up-to-date
- [ ] I adjusted the reporting in [`remind2`](https://github.com/pik-piam/remind2) where it was needed
- [ ] I adjusted `forbiddenColumnNames` in [readCheckScenarioConfig.R](https://github.com/remindmodel/remind/blob/develop/scripts/start/readCheckScenarioConfig.R) in case the PR leads to deprecated switches
- [x] All automated model tests pass (`FAIL 0` in the output of `make test`)
- [ ] The changelog `CHANGELOG.md` [has been updated correctly](https://gitlab.pik-potsdam.de/rse/rsewiki/-/wikis/Standards-for-Writing-a-Changelog)

PR #1419, opened 2023-09-22
fix path_gdx_bau checks, some more realizations need it

orichters:
## Purpose of this PR

- we forgot in #1366 that three other realizations actually read from the input_bau.gdx.
- I first implemented it in a straightforward manner, but then the NAVIGATE config failed because this file is needed only if cm_emiscen = 1, so it all became a bit more complicated.
- without this change, the start scripts would set `path_gdx_bau` to `NA` and then the realizations lack this file and fail

## Type of change

- [x] Bug fix 

## Checklist:


dklein-pik (approved):
Complicated but cool

PR #1432, opened 2023-10-17
Bugfix average CO2 tax calculation

Renato-Rodrigues:
## Purpose of this PR

Bugfix.

### Issue:

- Average CO2 emission market tax was wrongly calculated before because it was being calculated in the same loop statement that the CO2 emission market tax was being updated.

### Solution:

- Move CO2 tax definition for periods after the current target to an independent loop.

## Type of change

- [x] Bug fix 

## Checklist:

* Test runs are here: 
`/p/projects/ecemf/REMIND/2040_scenarios/v04_2023_06_02_WP5/output/new_04_Nzero_57_bio7p5_EU27_2023-10-16_15.02.05`

`/p/projects/ecemf/REMIND/2040_scenarios/v04_2023_06_02_WP5/compScen-NZero57_fixBug-2023-10-17_11.40.27-EU27.pdf`

### DIFF
--- a/modules/47_regipol/regiCarbonPrice/postsolve.gms
+++ b/modules/47_regipol/regiCarbonPrice/postsolve.gms
@@ -315,6 +315,13 @@ loop(ext_regi$regiEmiMktTarget(ext_regi),
   );
 );
 
+***  Assuming that other emissions outside the ESR and ETS see prices equal to the ESR prices
+loop((ttot,ttot2,ext_regi,emiMktExt,target_type_47,emi_type_47)$(pm_emiMktTarget(ttot,ttot2,ext_regi,"ESR",target_type_47,emi_type_47) or pm_emiMktTarget(ttot,ttot2,ext_regi,"all",target_type_47,emi_type_47)),
+  loop(regi$regi_groupExt(ext_regi,regi),
+    pm_taxemiMkt(t,regi,"other") = pm_taxemiMkt(t,regi,"ES");
+  );
+);
+
 *** updating periods after current target
 loop(ext_regi$regiEmiMktTarget(ext_regi),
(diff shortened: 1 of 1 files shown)


robertpietzcker (approved):
so if I understand correctly, the only change is that in the version before the PR, the years after the target period get their CO2 price right after the CO2 price adaption for the years of the target period itself,

 while after the PR, all targets first adjust the CO2 prices for the years during the target period, and only later the long-term prices are adjusted. 

do you understand why this makes a difference?

Renato-Rodrigues:
The loop `loop(emiMkt$emiMktGroup(emiMktExt,emiMkt),` first executes all code for the `ETS` item in emiMkt. When it finishes all calculations for the `ETS` item, it executes for the `ES` item, and finally executes for the `other` item.

If you have multiple emissions targets for the same region (e.g. one for 2020-2030, and another for 2035-2050), the code first runs the carbon price update for the first emission target period (e.g. 2020-2030), and use the average carbon price on 2030 to define the 2050 price in the given iteration.

In the previous formulation, the first run of the loop only updated the 2030 `pm_taxemiMkt` price for the ETS, and calculated the average price using the new ETS price, but using the old ESR and other prices. In the second run of the loop, the ESR price is  finally updated for 2030, and the average is recalculated again, but with a different value. Therefore, the 2050 ETS and ESR prices are using different average price values as reference, causing different `pm_taxemiMkt` prices for the period of 2035 to 2050.

In the new formulation, all prices for `pm_taxemiMkt` (ETS, ES and other) are updated first for 2030, and only after this finishes a single and consistent average price is calculated to define the average and, consequently, the `2050` price.

robertpietzcker:
thanks for the explanation!

PR #1434, opened 2023-10-19
cleaning and refactoring, use pm_taxCO2eqSum where possible, delete pm_pvpRegi

LaviniaBaumstark:
## Purpose of this PR
Cleaning and refactoring to ensure exact same values for year before startyear of ref secenario and current scenario. Use pm_taxCO2eqSum where possible and delete pm_pvpRegi.

## Type of change

- [x] Refactoring

## Checklist:

* Test runs are here: 

### DIFF
--- a/core/postsolve.gms
+++ b/core/postsolve.gms
@@ -19,7 +19,7 @@ $else.neg
     pm_taxCO2eqSum(ttot,regi) = abs( abs(q_co2eq.m(ttot,regi)) / (abs(qm_budget.m(ttot,regi))+ sm_eps) );
 $endif.neg 
    elseif (cm_emiscen eq 1),  !! even in a BAU scenario without other climate policies, the 2010/2015/2020 CO2 prices should be reported (that still needs to be fixed, I guess, maybe by adding the historic prices to the 45/carbonprice/off variation
-    pm_taxCO2eqSum(ttot,regi)$(ttot.val < 2025) = abs( pm_taxCO2eq(ttot,regi)); 
+    pm_taxCO2eqSum(ttot,regi)$(ttot.val < 2025) = pm_taxCO2eq(ttot,regi); 
 );
 
 if(cm_iterative_target_adj eq 4,
--- a/core/postsolve.gms
+++ b/core/postsolve.gms
@@ -12,7 +12,7 @@
 p_taxCO2eq_iteration(iteration,ttot,regi) = pm_taxCO2eq(ttot,regi);
 pm_taxemiMkt_iteration(iteration,ttot,regi,emiMkt) = pm_taxemiMkt(ttot,regi,emiMkt);
 
-if (cm_emiscen eq 6), 
+if( (cm_emiscen eq 6), 
 $ifthen.neg %optimization% == 'negishi'     
     pm_taxCO2eqSum(ttot,regi) = abs((abs(q_co2eq.m(ttot,regi)) / pm_ts(ttot)) / (pm_pvp(ttot,"good") + sm_eps));
 $else.neg
--- a/core/postsolve.gms
+++ b/core/postsolve.gms
@@ -14,7 +14,7 @@ pm_taxemiMkt_iteration(iteration,ttot,regi,emiMkt) = pm_taxemiMkt(ttot,regi,emiM
 
 *RP* added the historic 2010/2015 CO2 prices 
 if (cm_emiscen eq 9,
- pm_pvpRegi(ttot,regi,"perm") = (pm_taxCO2eq(ttot,regi) + pm_taxCO2eqRegi(ttot,regi) + pm_taxCO2eqSCC(ttot,regi))* pm_pvp(ttot,"good");
+ pm_pvpRegi(ttot,regi,"perm") = pm_taxCO2eqSum(ttot,regi)* pm_pvp(ttot,"good");
 elseif (cm_emiscen eq 6), !! the 2010/2015 CO2 prices do not need to be individually included, as they already influence the marginal of the q_co2eq equation (empirically tested) 
 
 $ifthen.neg %optimization% == 'negishi'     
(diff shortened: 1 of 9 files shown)


core/postsolve.gms
```diff
@@ -12,36 +12,6 @@
 p_taxCO2eq_iteration(iteration,ttot,regi) = pm_taxCO2eq(ttot,regi);
 pm_taxemiMkt_iteration(iteration,ttot,regi,emiMkt) = pm_taxemiMkt(ttot,regi,emiMkt);
 
-*RP* added the historic 2010/2015 CO2 prices
```
orichters, line 15: I'm absolutely not sure, but isn't the problem that now in cm_emiscen != 9, we don't have Price|Carbon correctly reported? I would think we have to put here:
```
if (cm_emiscen eq 6),
$ifthen.neg %optimization% == 'negishi'     
 pm_taxCO2eqSum(ttot, regi) = abs(q_co2eq.m(ttot,regi)) / pm_ts(ttot) / (pm_pvp(ttot,"good") + sm_eps);
$else.neg
 pm_taxCO2eqSum(ttot, regi) = abs(q_co2eq.m(ttot,regi)) / (abs(qm_budget.m(ttot,regi) )+ sm_eps); 
$endif.neg
```
etc., not sure about the other cases. No idea why we have the `abs` stuff here, that looks rather suspicious to me…

LaviniaBaumstark: well, pm_taxCO2eqSum is calculated for all emiscen cases. I think, the only problematic case is cm_emiscen eq 6. I will check, if this is still needed

orichters: Is it? Where? Because in the tax module, it seems it is simply an aggregation of `pm_taxCO2eq + pm_taxCO2eqRegi + pm_taxCO2eqSCC`, coming from the 45 module. But all these modules only work with emiscen = 9 (well, and 10 of course). So I guess with emiscen = 6, you will end up with `pm_taxCO2eqSum` = 0, if you don't overwrite it here.

LaviniaBaumstark: I agree, I hoped that we could delete (all) other emiscen scenarios. But seems we need to keep them, so added calculation of pm_taxCO2eqSum for more emiscen cases

core/presolve.gms
```diff
@@ -112,15 +112,7 @@ loop(regi,
           );
     );
 
-  if (cm_emiscen eq 9 or (cm_emiscen eq 10),
```
orichters, line 115: Sorry if I ask stupid questions, but for emiscen = 6, do we now have the right value here? Because we are in core/presolve, so the code in core/postsolve has not necessarily been executed, right? Or is it fine to just use the value saved in the last iteration?

core/postsolve.gms
```diff
@@ -12,34 +12,14 @@
...
-
-*LB* use the global permit price as regional permit price if no regional permit price is calculated
-loop(ttot$(ttot.val ge 2005),
-  loop(regi,
-    if(pm_pvpRegi(ttot,regi,"perm") eq NA,
-      pm_pvpRegi(ttot,regi,"perm") = pm_pvp(ttot,"perm");
-    );
-  );
+   elseif (cm_emiscen eq 1),  !! even in a BAU scenario without other climate policies, the 2010/2015/2020 CO2 prices should be reported (that still needs to be fixed, I guess, maybe by adding the historic prices to the 45/carbonprice/off variation
+    pm_taxCO2eqSum(ttot,regi)$(ttot.val < 2025) = abs( pm_taxCO2eq(ttot,regi));
```
orichters, line 22: why "abs"? I mean, that should never be negative, but if it is because of some error, we should maybe also report it…

core/postsolve.gms
```diff
@@ -12,34 +12,14 @@
...
 
-*RP* added the historic 2010/2015 CO2 prices 
-if (cm_emiscen eq 9,
- pm_pvpRegi(ttot,regi,"perm") = (pm_taxCO2eq(ttot,regi) + pm_taxCO2eqRegi(ttot,regi) + pm_taxCO2eqSCC(ttot,regi))* pm_pvp(ttot,"good");
-elseif (cm_emiscen eq 6), !! the 2010/2015 CO2 prices do not need to be individually included, as they already influence the marginal of the q_co2eq equation (empirically tested) 
-
+if( (cm_emiscen eq 6), 
 $ifthen.neg %optimization% == 'negishi'     
- pm_pvpRegi(ttot,regi,"perm") = abs(q_co2eq.m(ttot,regi)) / pm_ts(ttot) ;
+    pm_taxCO2eqSum(ttot,regi) = abs((abs(q_co2eq.m(ttot,regi)) / pm_ts(ttot)) / (pm_pvp(ttot,"good") + sm_eps));
```
orichters, line 17: also here, don't really understand all the `abs`. How can dividing a positive number by some positive number ever be negative? Doesn't matter, on the other hand :)

core/presolve.gms
```diff
@@ -112,15 +112,7 @@ loop(regi,
...
-*** TODO: take care, this means that the SCC are only priced into MAC-curve
-*** abatement if emiscen = 9 and for emiscen = 10 for CBA runs. Might want to change this.
-    p_priceCO2(ttot,regi) = (pm_taxCO2eq(ttot,regi) + pm_taxCO2eqRegi(ttot,regi) + pm_taxCO2eqSCC(ttot,regi) )* 1000;
-  else
-    p_priceCO2(ttot,regi) 
-    = abs(pm_pvpRegi(ttot,regi,"perm") / (pm_pvp(ttot,"good") + sm_eps))
-    * 1000;
-  );
+p_priceCO2(ttot,regi) = pm_taxCO2eqSum(ttot,regi) * 1000;
```
orichters, line 115: I hoped we could get rid of `p_priceCO2` also, but I see that would be a bigger issue…

core/postsolve.gms
LaviniaBaumstark: yes, I agree, let's use the "true" value without abs

core/presolve.gms
LaviniaBaumstark: in the very first Nash iteration we will have the wrong value but afterwards it will use the number from the last iteration which is o.k. I would say

orichters: fine!

orichters (approved):
Looks good from my side. As I never used the various `cm_emiscen` options, I'm not sure whether I fully understand whether everything is fine for `cm_emiscen = 4` or `6`, although it looks correct…

PR #1437, opened 2023-10-19
adjust geohdr costs to be consistent with region specific 2015-2040 values

pweigmann:
## Purpose of this PR
This fixes an issue with Capital Costs for Geothermal electricity falling in the near-term but then rising until 2070 by lowering the value they are converging towards: the `inco0` of `geohdr` in the `generisdata_tech.prn` files.

Due to very limited resource potentials, this will likely not have a strong effect on geothermal electricity generation.

## Type of change

- [x] Bug fix 

## Checklist:

## Further information:

* Test runs are here: 
* Comparison of results:

### DIFF
--- a/core/input/generisdata_tech.prn
+++ b/core/input/generisdata_tech.prn
@@ -57,7 +57,7 @@ lifetime              35         35          40          40          40         
 
 +                geohdr       geohe       hydro        wind         windoff     spv       solhe         csp
 tech_stat            1
-inco0               3600        1200        2700        2400         5000       5160        9999       10320
+inco0               1950        1200        2700        2400         5000       5160        9999       10320
 constrTme            4           1           5            1            4          1           1          3
 mix0
 eta                 1.00        1.00        1.00        1.00         1.00       1.00        1.00        1.00
--- a/core/input/generisdata_tech_SSP1.prn
+++ b/core/input/generisdata_tech_SSP1.prn
@@ -57,7 +57,7 @@ lifetime              35         35          40          40          40         
 
 +                geohdr       geohe       hydro        wind         windoff     spv       solhe         csp
 tech_stat            1
-inco0               3600        1200        2700        2400         5000       5160        9999       10320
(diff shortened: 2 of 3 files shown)


pweigmann:
@robertpietzcker I chose the minimum value across regions from 2040 as new value for inco0 in `generisdata_tech`, is that too optimistic?

Also, Nuclear and Hydro also seem to have an increase in Capital Cost between 2030 and 2070. While the former might make sense, should the latter be changed as well?

robertpietzcker:
thanks!
could you set up a check cascade NPI->NDC->PkBudg500 (SSP2EU)  to be sure that it really doesn't cause any major changes to the power sector results?

pweigmann:
> thanks! could you set up a check cascade NPI->NDC->PkBudg500 (SSP2EU) to be sure that it really doesn't cause any major changes to the power sector results?

will do! what about the hydro costs?

robertpietzcker:
> what about the hydro costs?

Actually, on second thought, I would take both hydro and geohdr out of `teRegTechCosts", so that it simply uses the generic values and doesn't have the regional differentiation. 

Regional differentiation is great if you also have spent some real time thinking about what limits/enables technology deployment in certain region, but we never did that, neither for hydro nor geohdr. 

So you can end up with a technology having on the paper very low investment costs in a region, such as hydro in the US which costs 1975/kW, the lowest value for ALL regions, but in fact it won't be expanded because of (bad resource potential/ getting people off the land that you want to use is expensive / environmental rules / xxxx).

so I would say with having these technologies have regionally differentiated costs creates the illusion that we actually have spent time thinking about it, which we didn't. 

currently the prices go up between 2020-2050 in some regions with low 2020 costs; setting the 2050 value lower will lead to larger deployment until 2050 although it already now seems quite high:
image

so I would prefer to simply take it out of the "regionally differentiated costs" set. 

or what do you think, @nicobauer ?

LaviniaBaumstark:
what is the status of this PR?

pweigmann:
@nicobauer what do you think?

Issue #1421, opened 2023-09-26
Input files required to run the REMIND model


dengzhong:
Hi,
And according to the README, the input data is not part of this project. But do we provide any example data for me to understand the structure and the format of the input data?
Many thanks!

LaviniaBaumstark:
For long time we where not allowed to distribute input data for REMIND. Therefore, we still have no process to do so but we are working on it. In the meantime we could provide you example data for the latest release version of REMiND. Please, send an e-mail to remind@pik-potsdam.de for further coordination.

LaviniaBaumstark:
we are now providign input data for our releases

Issue #299, opened 2021-01-13
Download and distribute input data error


fei-cheng:
Hi,

I downloaded the latest release and tried to run this model. It threw an error while trying to download and distribute input data. The error message looks like:
```
Load data..
Following files not found:
  rev5.961_690d3718e151be1b450b394c1064b1c5_remind.tgz
```
My understanding is it tries to download the above tgz file from somewhere. But where is the file stored?

And according to the README, the input data is not part of this project. But do we provide any example data for me to understand the structure and the format of the input data?

Many thanks!

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
Hi 

As noted in the [readme](https://github.com/remindmodel/remind/blob/develop/README.md#how-to-install) we can at present not share or publish the required input data for legal reasons.  Should you have licensed all the necessary data (the biggest item being the [IEA Energy Balances](http://data.iea.org/payment/products/117-world-energy-balances-2018-edition-coming-soon.aspx)), we can assist you in generating the data locally.  Best get in touch via remind@pik-potsdam.de to discuss details.

LaviniaBaumstark:
we are now providing input data for our releases

PR #1438, opened 2023-10-23
Added Issue Templates

whitehacker:
## Purpose of this PR

This pull request introduces standardized issue templates to streamline the issue creation and resolution process in our REMIND repository. The templates cover Feature Requests, Bug Reports and we can add more custom templates if needed. 

These are draft templates and feedback on the template format or content is highly welcomed.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug.yml
+++ b/.github/ISSUE_TEMPLATE/bug.yml
@@ -1,0 +1,78 @@
+name: Bug Report
+description: This template is for reporting bugs found in the REMIND model
+title: "[Bug]: "
+labels: ["bug", "remind"]
+assignees:
+  - whitehacker
+body:
+  - type: markdown
+    attributes:
+      value: |
+        Thanks for taking the time to fill out this bug report!
+  - type: input
+    id: contact
+    attributes:
+      label: Contact Details
+      description: How can we contact you if we need more info?
+      placeholder: ex. email@example.com
+    validations:
+      required: false
+  - type: checkboxes
+    id: checklist
+    attributes:
+      label: Please confirm the following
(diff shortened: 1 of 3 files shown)


0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q (commented):
Seems like an awful lot of work to make everybody who actually opens REMIND issues click once more often, seeing that the only issues we ever got from outside where about making input data available https://github.com/remindmodel/remind/issues/299, https://github.com/remindmodel/remind/issues/1421.

LaviniaBaumstark:
right, this is a preparation step for starting collaboration with external users to be prepared for their issues and to avoid wrong expectations.

.github/ISSUE_TEMPLATE/bug.yml
```diff
@@ -0,0 +1,78 @@
...
+      required: true
+  - type: textarea
+    id: logs
+    attributes:
+      label: Relevant log output
+      description: Please copy and paste any relevant log output. This will be automatically formatted into code, so no need for backticks.
+      render: shell
+  - type: checkboxes
+    id: terms
+    attributes:
```
LaviniaBaumstark, line 72: maybe we should add here that we are happy about any bug report and will try to fix bugs in REMIND soon. But that we con only provide very limited support for setting up and running REMIND outside PIK environment.

whitehacker: Sure, and added.

renv/settings.json
tonnrueter, line 1: I do not understand how the `renv` settings file is relevant to this PR. I suspect that the presence of this is due to an accidental commit

PR #1441, opened 2023-10-26
Nash convergence tracking

LaviniaBaumstark:
## Purpose of this PR
add and adjust parameters for tracking Nash convergence criteria over iterations

## Type of change

- [x] Refactoring
- [x] New feature 

## Checklist:

* Test runs are here: 


modules/80_optimization/nash/output.gms
```diff
@@ -24,7 +24,7 @@ putclose prices_NASH;
 *** The file is formated in columns as follows: | Experiment title | Region | Year | Iteration | Market | surplus | price |  ..
 file nash_info_convergence / "nash_info_convergence.csv" / ;
 put nash_info_convergence;
-put 'Scenario',',','Region',',','Year',',','Iteration',',',"Market",",","p80_surplus",",","p80_pvp_itr",",","p80_surplusMax",",","p80_surplusMaxRel",",":0 ;
+put 'Scenario',',','Region',',','Year',',','Iteration',',',"Market",",","p80_surplus",",","p80_pvp_itr",",","p80_surplusMax_iter",",","p80_surplusMaxRel",",":0 ;
```
0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q, line 27: This look painful.
https://www.gams.com/latest/docs/UG_Put.html#INDEX__2f_pc_22_suffixes_2d__21_put_2e_file

> 5: Formatted output; non-numeric output is quoted and each item is delimited with commas.

modules/80_optimization/nash/postsolve.gms
```diff
@@ -283,38 +280,36 @@ loop(trade$(NOT tradeSe(trade)),
...
+         p80_messageShow("infes") = YES;
+     );
 *** critertion "nonopt": The next lines are a workaround for the status 7
 *** problem. If the objective value does not differ too much from the last known
 *** optimal solution, accept this solution as if it were optimal. 
-  if (1 le iteration.val,
-    !! no last iteration if this is the first; NA value in p80_repyLastOptim is
-    !! sticky, so test this separately
-    if (   p80_repy(regi,'modelstat') eq 7
+     p80_convNashObjVal_iter(iteration,regi) = abs(p80_repy(regi,'objval') - p80_repyLastOptim(regi,'objval'));
```
Renato-Rodrigues, line 290: I would save the `p80_convNashObjVal_iter` parameter without the `abs` transformation, so we can keep the deviation signal in the parameter for further analysis if necessary.

modules/80_optimization/nash/declarations.gms
```diff
@@ -92,11 +92,20 @@ p80_count                                  "count regions with feasible solution
 p80_eoWeights_fix(all_regi)                "default and fallback weighting factors"
 
 p80_SolNonOpt(all_regi)                    "solve status"
-p80_taxrev_dev(ttot,all_regi)              "deviation of tax revenue normalized by GDP if taxes did not converge"
 
 pm_fuExtrForeign(ttot,all_regi,all_enty,rlf) "foreign fuel extraction"
-p80_taxrev_dev(ttot,all_regi)                "deviation of tax revenue in percent GDP"
-;
+
+p80_convNashTaxrev_iter(iteration,ttot,all_regi) "deviation of tax revenue in percent GDP per iteration"
```
Renato-Rodrigues, line 98: as some other convergence deviation parameters include the `_dev` suffix, maybe we should adopt this on the naming convention for these convergence deviation parameters also.

modules/80_optimization/nash/postsolve.gms
```diff
@@ -323,26 +318,25 @@ if(p80_DevPriceAnticipGlobAllMax("2100") gt 0.1 * p80_surplusMaxTolerance("good"
...
+loop(regi,
     loop(t,
-      if( abs(vm_taxrev.l(t,regi)) / vm_cesIO.l(t,regi,"inco") gt 1E-4,
-        p80_taxrev_dev(t,regi) = abs(vm_taxrev.l(t,regi)) / vm_cesIO.l(t,regi,"inco");
-        if (t.val lt 2100,
-          s80_bool = 0;
-          p80_messageShow("taxconv") = YES;
-        );
-      );
+         p80_convNashTaxrev_iter(iteration,t,regi) = abs(vm_taxrev.l(t,regi)) / vm_cesIO.l(t,regi,"inco");
```
Renato-Rodrigues, line 324: I would save this parameter without the `abs` transformation to save the sign information for further debugging if necessary.

modules/80_optimization/nash/postsolve.gms
```diff
@@ -323,26 +318,25 @@ if(p80_DevPriceAnticipGlobAllMax("2100") gt 0.1 * p80_surplusMaxTolerance("good"
...
-  );
 );
 
 *** additional criterion: Were regional climate targets reached? 
+p80_emiMktTarget_dev_iter(iteration,ttot,ttot2,ext_regi,emiMktExt) = abs(pm_emiMktTarget_dev(ttot,ttot2,ext_regi,emiMktExt));
 $ifthen.emiMkt not "%cm_emiMktTarget%" == "off" 
 loop((ttot,ttot2,ext_regi,emiMktExt)$pm_emiMktTarget_dev(ttot,ttot2,ext_regi,emiMktExt),
 *** regipol targets must be met within 1% of target deviation, deviation for budget targets is measured relative to target value, while for year targets it is relative to 2015 emissions
-  if((abs(pm_emiMktTarget_dev(ttot,ttot2,ext_regi,emiMktExt)) gt cm_emiMktTarget_tolerance),
+  if((p80_emiMktTarget_dev_iter(iteration,ttot,ttot2,ext_regi,emiMktExt) gt cm_emiMktTarget_tolerance),
```
Renato-Rodrigues, line 339: `pm_emiMktTarget_dev_iter` is actually defined and used in module 47. Its name shouldn't change to use `p80` without further refactoring.

Renato-Rodrigues (changes requested):

modules/80_optimization/nash/declarations.gms
LaviniaBaumstark: I think, here it is more important that it it a Nash con. criterion, so that it includes `convNash`

modules/80_optimization/nash/postsolve.gms
LaviniaBaumstark: `pm_emiMktTarget_dev` is defined in module 47, `p80_emiMktTarget_dev_iter` is a new parameter intrduced for the Nash convergance analysis

modules/80_optimization/nash/postsolve.gms
LaviniaBaumstark: changed it, it also had a bug

modules/80_optimization/nash/postsolve.gms
Renato-Rodrigues: The `p80_emiMktTarget_dev_iter` parameter is duplicating the values of the already existent `pm_emiMktTarget_dev_iter` parameter.
The only difference is that it is applying the absolute transformation to the values, which should be removed in my opinion as it removes information from the parameter that could be used for debugging purposes.

modules/80_optimization/nash/postsolve.gms
LaviniaBaumstark: also before we did only use the abs info here

modules/80_optimization/nash/declarations.gms
Renato-Rodrigues: I didn't ask for changing the `convNash` part of the parameter name. My request was to add the _dev suffix (e.g. p80_convNashTaxrev_dev_iter) like it is used in other existent convergence check parameters (e.g. `p80_emiMktTarget_dev_iter`, `p80_implicitQttyTarget_dev_iter`, `p80_globalBudget_dev_iter`).

modules/80_optimization/nash/postsolve.gms
```diff
@@ -283,38 +280,36 @@ loop(trade$(NOT tradeSe(trade)),
...
-      display "Not all regions were status 2 in the last iteration. The deviation of the objective function from the last optimal solution is too large to be accepted:";
-      s80_dummy
-      = p80_repy(regi,'objval')
-      - p80_repyLastOptim(regi,'objval');
-      display s80_dummy;
-    );
-  );
-);
+           !! the solution differs over iteration when status 7 occurs. 
+           AND p80_convNashObjVal_iter(iteration,regi) lt - 1e-4,
```
Renato-Rodrigues, line 297: shouldn't this be 
```
AND abs(p80_convNashObjVal_iter(iteration,regi)) lt 1e-4,
```

Renato-Rodrigues (changes requested):

modules/80_optimization/nash/postsolve.gms
Renato-Rodrigues: but it would be better to save the info without the absolute value, like we do for other tests, and only use the absolute transformation in the conditional checks.

modules/80_optimization/nash/postsolve.gms
LaviniaBaumstark: ah, wasn't aware of `pm_emiMktTarget_dev_iter`, added it to all realizations to use it here

modules/80_optimization/nash/postsolve.gms
LaviniaBaumstark: no, it was a mistake we introduced with the refactoring. We want to only accept `7` if the obj. value is not too much smaler than the obj. value from last iteration. So, the direction matters!

modules/47_regipol/none/datainput.gms
```diff
@@ -9,6 +9,7 @@
 
 * initialize regipol target deviation parameter
 pm_emiMktTarget_dev(ttot,ttot2,ext_regi,emiMktExt) = 0;
+pm_emiMktTarget_dev_iter(iteration, ttot,ttot2,ext_regi,emiMktExt) = 0;
```
Renato-Rodrigues, line 12: There is no need to initialize this parameter on the `none` realization, as the parameter should never be used in the model in this case because `cm_emiMktTarget` would be `off`. 
If you set a scenario with `cm_emiMktTarget` values, the realization `regiCarbonPrice` is required to be active.

Renato-Rodrigues (changes requested):

modules/47_regipol/none/datainput.gms
LaviniaBaumstark: yes, but we would likt to always have all parameters available whish we need to the nash conv. visualization

modules/80_optimization/nash/postsolve.gms
Renato-Rodrigues: Interesting, I find it odd that if the current iteration has a very different, but higher, value for `p80_repy(regi,'objval')` versus the previous iteration, we would accept the model to be stable enough to accept a non optimal solution.

fbenke-pik: Right, but this is how it is implemented on the main develop branch right now, isn't it? If this should be changed, now would be a good time.

Renato-Rodrigues: Yes, this is the way it is now.
Maybe create an issue about this and assign probably Marian so he can give his opinion about changing this?

modules/80_optimization/nash/output.gms
LaviniaBaumstark: but we are anyway looking at other parameter with the same dimensions

modules/80_optimization/nash/postsolve.gms
LaviniaBaumstark: but here the abs() only holds for `vm_taxrev` and not the whole term

fbenke-pik (approved):

modules/80_optimization/nash/postsolve.gms
Renato-Rodrigues: `vm_cesIO` is a quantity and declared as a positive variable, so it is always positive. Therefore, applying the abs() to the numerator or to both the numerator and denominator makes no difference.

Renato-Rodrigues (approved):

PR #1448, opened 2023-11-01
fix automatic path_gdx_bau correction

orichters:
## Purpose of this PR
- the inequality module does not need `path_gdx_bau` anymore, the last occurence was removed by @piontek, except for one that was commented out
- adjust second check in `checkFixConfig` that might remove the path_gdx_bau despite it being needed

## Type of change

- [x] Bug fix 

## Checklist:

### DIFF
--- a/scripts/start/readCheckScenarioConfig.R
+++ b/scripts/start/readCheckScenarioConfig.R
@@ -83,6 +83,7 @@ readCheckScenarioConfig <- function(filename, remindPath = ".", testmode = FALSE
   }
   if ("path_gdx_bau" %in% names(scenConf)) {
     # fix if bau given despite not needed. needBau is defined in needBau.R
+    # initialize vector with FALSE everywhere and turn elements to TRUE if a scenario config row setting matches a needBau element
     scenNeedsBau <- rep(FALSE, nrow(scenConf))
     for (n in intersect(names(needBau), names(scenConf))) {
       scenNeedsBau <- scenNeedsBau | scenConf[[n]] %in% needBau[[n]]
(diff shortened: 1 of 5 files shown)


scripts/start/readCheckScenarioConfig.R
```diff
@@ -82,20 +82,10 @@ readCheckScenarioConfig <- function(filename, remindPath = ".", testmode = FALSE
...
-    for (n in names(needBau)) {
-      if (n %in% names(scenConf)) {
-        if (n == "welfare" && "cm_emiscen" %in% names(scenConf)) {
-          scenNeedsBAU <- scenNeedsBau | (scenConf[[n]] %in% needBau[[n]] & scenConf[["cm_emiscen"]] != 1)
-        } else {
-          scenNeedsBau <- scenNeedsBau | scenConf[[n]] %in% needBau[[n]]
-        }
-      }
+    for (n in intersect(names(needBau), names(scenConf))) {
+      scenNeedsBau <- scenNeedsBau | scenConf[[n]] %in% needBau[[n]]
```
dklein-pik, line 88: In L86 `scenNeedsBau` has been defined as a vector. Still needed?

orichters: Yes, it is still a vector, as `scenConf[[n]]` is also vector (for example, the `carbonprice` column). So `scenNeedsBau` starts as being `FALSE` everywhere, and once any of these columns has an element that needs a `path_gdx_bau`, it is turned to `TRUE` but only in the entry corresponding to that row.

dklein-pik: Oh sorry, I didn't notice, that `scenConf[[n]]` is also vector.

dklein-pik (approved):

PR #1450, opened 2023-11-01
adjust test-code workflow to avoid timeout when installing gms

fbenke-pik:
## Purpose of this PR

- With the current settings, installation of gms using pak was hanging until aborted (new issue since 30th October)
- I removed "https://cloud.r-project.org" from extra-repositiories, as for gms and its dependencies CRAN and the pik repo should suffice
- I removed `gms` from `extra-packages`, as gms should be installed as specified in the DESCRIPTION file 
- I added `any::` as a prefix to gms, since this seems to be the right way to do it
- I updated the checkout version to v3, since it is also used in current examples
- [Reference](https://github.com/r-lib/actions/tree/v2/setup-r-dependencies)

## Type of change

- [x] Bug fix

### DIFF
--- a/.github/workflows/test-code.yaml
+++ b/.github/workflows/test-code.yaml
@@ -18,20 +18,19 @@ jobs:
     env:
       GITHUB_PAT: ${{ secrets.GITHUB_TOKEN }}
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
 
       - name: Remove .Rprofile
         run: rm .Rprofile
 
       - uses: r-lib/actions/setup-r@v2
         with:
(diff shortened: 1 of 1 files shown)


orichters (approved):
looks good: the check does not fail anymore in your case…

pfuehrlich-pik (approved):

pfuehrlich-pik:
Not quite clear to me why this fixes things, is it the `any::`?

fbenke-pik:
I think it is removing the extra repository and extra package argument did the trick, as these reduce complexity for the pak dependency solver.  However, I have not found the reason why it started taking so long that the job had to be cancelled.

fbenke-pik:
`::any` was probably even not used correctly, as it is required for `extra-packages`. i sneaked in a fix removing the prefix in a subsequent PR, and it still works.

PR #1455, opened 2023-11-03
For renv integration: change how first run in a cascade is detected

### LINKED ISSUE #1409: avoid future renv issues for coupled runs
while #1407 is fixed now, we should try to avoid such complications in the future

- REMIND-MAgPIE runs should behave in the same way as REMIND standalone with respect to renv
- [This line](https://github.com/remindmodel/remind/blob/712dec7ec3903272f46a1d97c7fc1c01e9d2f3af/scripts/start/submit.R#L52) seems to evaluate to `TRUE` for every REMIND-MAgPIE run, which I think it shouldn't. I guess it is because coupled runs have the REMIND main folder as working directory.
- This results in the `renv` code below being executed on the compute nodes which failed, while it works on the login nodes. Not sure whether that also means that `renv` does not work as intended at the moment.

suggestions:
- move the renv checks from [submit.R](https://github.com/remindmodel/remind/blob/develop/scripts/start/submit.R#L46-L96) to an own function in `scripts/start/`
- only call the checks, snapshot generation etc. it in `start.R` and `start_bundle_coupled.R`, not in `submit()` to avoid failing runs
- idea: for subsequent runs in a cascade (for both standalone and REMIND-MAgPIE), copy the renv stuff not from the main folder, but from the folder that starts the run, to guarantee that all the cascade runs on the same renv

dklein-pik:
## Purpose of this PR

In a cascade of runs only the first run in a cascade should create a `renv.lock` by generating a `renv::snaphot()`, while all subsequnt runs should use this `renv.lock` from the first run to execute `renv::restore()` and not create their own snapshot. The detection of whether the current run to be started is the first run in a cascade didn't work properly for coupled runs, as explained here #1409 . 

In this PR I make a suggestion how [the line](https://github.com/remindmodel/remind/blob/712dec7ec3903272f46a1d97c7fc1c01e9d2f3af/scripts/start/submit.R#L51C1-L51C1)

`firstRunInCascade <- normalizePath(renv::project()) == normalizePath(".")`

mentioned by @orichters that tries to detect the first run in a cascade (and thus needs to create a renv snapthot) could evaluate correctly for standalone *and* coupled runs. The idea is to add a new entry `cfg$renvLockFromPreceedingRun` to the cfg only for subseqnent runs. For the first run in a cascade this entry is NULL and a snaphot is generated. If it exists it contains the path to the `renv.lock` of the preceeding run, that should be used by `renv:restore()` for the subsequent run.

Fixes #1409 

## Type of change

- [x] Bug fix 
- [ ] Refactoring
- [ ] New feature 
- [ ] Fundamental change

## Checklist:

* Test runs are here: /p/tmp/dklein/remind-renvtest


start_bundle_coupled.R
```diff
@@ -650,6 +650,7 @@ for (scen in common) {
 
 if (! "--test" %in% flags && ! "--gamscompile" %in% flags) {
```
orichters, line 651: That should be dropped completely, I think. You should always use `renv`, I think.

scripts/start/run.R
```diff
@@ -207,6 +207,9 @@ run <- function() {
       }
       message("In ", RData_file, ", use current fulldata.gdx path for ", paste(needfulldatagdx, collapse = ", "), ".")
       cfg$files2export$start[needfulldatagdx] <- fulldatapath
+      # let the subsequent run use the renv.lock of this run
+      message("In ", RData_file, ", use current renv.lock for subsequent run ", run, ".")
+      cfg$renvLockFromPreceedingRun <- file.path(cfg_main$remind_folder,cfg_main$results_folder,"renv.lock")
```
orichters, line 212: you might need to add that to `remindextras` in `scripts/start/checkFixCfg.R`

orichters (commented):
I just had a very short look, but I think you chose a good concept. Haven't checked your implementation in detail, yet.

An alternatve would just to have `cfg$renvlock <- "renv.lock"` or whatever in `default.cfg`, and while starting subsequent runs, it just replaces it. `start.R` and `start_bundle_coupled.R` would write this file, `submit` would only copy. But no idea whether this is has any (dis)advantages compared to the way you implemented it.

start_bundle_coupled.R
dklein-pik: Done

scripts/start/run.R
dklein-pik: Done

scripts/start/checkFixCfg.R
```diff
@@ -15,7 +15,7 @@ checkFixCfg <- function(cfg, remindPath = ".", testmode = FALSE) {
   refcfg <- gms::readDefaultConfig(remindPath)
   remindextras <- c("backup", "remind_folder", "pathToMagpieReport", "cm_nash_autoconverge_lastrun",
                                "gms$c_expname", "restart_subsequent_runs", "gms$c_GDPpcScen",
-                               "gms$cm_CES_configuration", "gms$c_description", "model")
+                               "gms$cm_CES_configuration", "gms$c_description", "model", "renvLockFromPreceedingRun")
```
pfuehrlich-pik, line 18:
```suggestion
                               "gms$cm_CES_configuration", "gms$c_description", "model", "renvLockFromPrecedingRun")
```
Although I prefer `previous` anyway :wink:

scripts/start/run.R
pfuehrlich-pik, line 212:
```suggestion
      cfg$renvLockFromPreceedingRun <- file.path(cfg_main$remind_folder, cfg_main$results_folder, "renv.lock")
```

scripts/start/submit.R
```diff
@@ -75,8 +76,8 @@ submit <- function(cfg, restart = FALSE, stopOnFolderCreateError = TRUE) {
         message("done.")
       } else {
         # a run renv is loaded, we are presumably starting new run in a cascade
-        message("   Copying lockfile into '", cfg$results_folder, "'")
-        file.copy(renv::paths$lockfile(), file.path(cfg$results_folder, "_renv.lock"))
+        message("   Copying lockfile '",cfg$renvLockFromPreceedingRun,"' into '", cfg$results_folder, "'")
```
pfuehrlich-pik, line 79:
```suggestion
        message("   Copying lockfile '", cfg$renvLockFromPreceedingRun, "' into '", cfg$results_folder, "'")
```

pfuehrlich-pik:
Looks good to me, although I didn't test anything

dklein-pik:
Probably because I added a few commits that fix typos I need your review again, unfortunately.

orichters (approved):
I also did not test it (but I assume David did).

I checked that code and that looks convincing.

orichters:
> Probably because I added a few commits that fix typos I need your review again, unfortunately.

No, I think the PR was never formally approved, but we just added comments until now.

dklein-pik:
> I also did not test it (but I assume David did).
> 
> I checked that code and that looks convincing.

Yes, I did.

orichters:
Thanks, David, for fixing that!

Issue #1479, opened 2023-11-28
MR #1154 broke REMIND EU21


0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
Due to hard-coded H12 regions in `./modules/33_CDR/portfolio/input/p33_transport_costs.inc`
```
$ sed -n 's/.*"\([A-Z]\{3\}\)".*/\1/p' ./modules/33_CDR/portfolio/input/p33_transport_costs.inc | uniq 
CAZ
CHA
EUR
IND
JPN
LAM
MEA
NEU
OAS
REF
SSA
USA
```
That file has to be generated dynamically from `mrremind` to provide the correct regional data.
REMIND EU21 is inoperable until either that is done or this change is reverted, as are the mandatory model tests.

*****

Also, MRs #1154, #1477, #1478 cannot possibly have had valid passing model tests
> * [x]  All automated model tests pass (`FAIL 0` in the output of `make test`)

@orichters @robertpietzcker

robertpietzcker:
> Also, MRs https://github.com/remindmodel/remind/pull/1154, https://github.com/remindmodel/remind/pull/1477, https://github.com/remindmodel/remind/pull/1478 cannot possibly have had valid passing model tests

        All automated model tests pass (FAIL 0 in the output of make test)

@orichters @robertpietzcker

hm, how can that be? I executed "make test", and I got "fail 0":

![image](https://github.com/remindmodel/remind/assets/61154412/5a822708-0996-4548-8522-feeeca11f247)

any ideas, Michaja?

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
> any ideas, Michaja?

Sure.
The commit you tested
```
/p/projects/remind/users/robertp/savePR/2023-11-24_deltacapFixCCS/new_onlyNoCCSdeltacapFx (deltacapFixCCS|?? M)$ git log -n 1 --format="%h"
52e2532
```
is waaay down there in the commit history of your branch at the point of merging.
```
$ git log --graph --format="%h | %ch | %s" 52e2532^..96d5a0b
*   96d5a0b | Mon 13:57 | Merge pull request #1478 from robertpietzcker/deltacapFixCCS
|\  
| *   56242a0 | Mon 10:53 | Merge branch 'deltacapFixCCS' of https://github.com/robertpietzcker/remind into deltacapFixCCS
| |\  
[...]

PR #1482, opened 2023-11-28
Set dummy EW data to fix EU21

katarkow:
## Purpose of this PR
Here's a temporary fix for #1479 before I migrate EW transport costs to mrremind.

## Type of change

- [x] Bug fix


LaviniaBaumstark:
did you try to run `make test`?

Issue #302, opened 2021-01-19
reduce github repository size


giannou:

Renato-Rodrigues:
Adding below some context and a suggestion for an approach to handle this issue.

**Description of the problem:**

The increased remind repository size was caused by updating binary files that are part of our version control list. They are represented by the gdx files used as input data in REMIND. 

Every time that any binary file is updated in a commit, git saves a copy of the old file to be able to restore the model to the previous state. As gdx files are already compressed formats and have in between 10-20 mb in file size in our case, every time they are updated the model size will increase by approximately the size of the replaced or deleted gdx file.   

You can check that only 18 gdx files, most of them that are not part of remind anymore, correspond to 90% of the recent git folder size increase by using the commands below:
```
git verify-pack -v .git/objects/pack/pack-a8707650a83bdb954302efdf1a999c297de61825.idx \
| sort -k 3 -n \
| tail -100

git rev-list --objects --all | grep  0f78fcba8ee03ea1e8d8407d4f3da5dd1a084e5f
```

**Why is this important:**

- Every time a change is made to the CES tree structure in REMIND, the input gdx file needs to be mandatorily updated for every region mapping used in the model.  
- Every time any meaningful changes are introduced at calibration level (ex: adjusting FE baseline trajectories, GDP levels,...) it is highly recommended to update the input.gdx file as this will not only speed up the first REMIND iteration (which takes much more time to solve than the other iterations), but also make sure that the model searches for a solution in the vicinity of the already tested calibration process.
- Alternative FE module realizations usually require specific input data gdxs.

So, it is right to assume that we will eventually need to update binary gdx files in the model and the repository will keep growing in size as time passes.

**Proposal for a solution:**

1. From now on, gdx commits should only be made if extremely necessary to avoid escalating the issue.
2. An alternative repository and version control system should be set in place to handle these binary files, and once this is done, all gdx files should be removed from the remind repository to this alternative repository. (ex. https://docs.gitlab.com/ee/topics/git/lfs/)
3. For the next full version of REMIND, gdx files should not be part of the repository and the git history should be edited to remove all gdx packed files (https://rtyley.github.io/bfg-repo-cleaner/ , https://github.com/newren/git-filter-repo). To keep backwards compatibility for older publications, the current repository should be copied with the git history intact to an alternative repository.

orichters:
Remind has surpassed 1GB. Maybe we should do said cleanup now?

giannou:
[...]

PR #1491, opened 2023-12-01
Add file size check

fbenke-pik:
## Purpose of this PR
Following up on [this discussion](https://github.com/remindmodel/remind/issues/302#issuecomment-1277403120), this PR introduces a simple file size limit check for all files currently under version control. It is part of the workflow test-code that is run for any PRs and commits on main, master and develop branches. 

The 600 kB limit is chosen to exceed the current largest file 
`modules/31_fossil/exogenous/input/p31_fix_fuelex.put (570.67kB)`

## Type of change
- [x] New feature 

## Checklist:


fbenke-pik:
> Also, could this not be put under the .github/ directory? The Remind directory is cluttered as it is …
Yes, happy to do that

> I feel it should be easy to move some stuff to mrremind and go to maybe 250 KB.
True, we can look into moving stuff. For some of the listed files, I have ideas. For others, not. 
@LaviniaBaumstark should this be done before or after merging this PR?

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q (changes requested):
Having looked into this a bit, I would call this a workflow or something.  It is not a git hook, and should not be confused with one.

> - Do we want to restrict the check to new files only? In that case, we need to come up with a clever way to define "new files". Maybe new files in comparison to develop?

`git ls-tree -r HEAD --name-only` "lists the contents of a given tree object", i.e. the files currently existing in the repository.
`git diff --cached --name-only` lists "the changes you staged for the next commit" i.e. the new files.
So drop the first to just get changed files.

Better yet, use `substring(grep("^[AM]", system("git status --short --porcelain", intern = TRUE), value = TRUE), first = 4)` to get only files that have been either added or modified, and not rely on the fact that deleted files do not exist and do not have a file size (`out <- out[!is.na(out$size), ]`) ;)

> - If we want to reduce file size limit, but not restrict it to new files, we must come up with a way to deal with files currently exceeding the limit: either fixing the files or defining hard-coded exceptions comes to mind.

I do not see the point in checking the size of all files for every single merge request.  Looks like some sort of code atavism to me.

fbenke-pik:
From what I can see, the problem with both your approaches is that the only work for the latest commit. 
However, the actual check is run when you try to merge into develop or push to the develop branch of your fork in the github repo and that could include more than one commit. So any solution looking at only new files must then identify all relevant changes spanning possibly multiple commits.  Or am I missing sth here, Michaja?


There is one alternative, to make this check part of [these checks in pre-commit.ci](https://results.pre-commit.ci/run/github/226360184/1701255062.IFyTS72sSy6ooEcx0mYL6w) using [the hook "check-added-large-files"](https://github.com/pre-commit/pre-commit-hooks#check-added-large-files). If this works as I hope it would, there should be a check of file size when opening a PR.

fbenke-pik:
Looks like pre-commit check won't work as expected: https://results.pre-commit.ci/run/github/226360184/1701792228.w1y28gzhSL-kQ9aVwps_zw

It is just fine with me adding a 14 MB mif on a test branch.

LaviniaBaumstark:
> p31_fix_fuelex



> Looking at the large files in the repository
> 
> ```
> $ git ls-tree -r HEAD --name-only | xargs ls -sh | sort -k 1hr,1 | head -n 20
> 572K modules/31_fossil/exogenous/input/p31_fix_fuelex.put
> 216K core/input/grid/EDGAR_CO2_2010_excl_agr_and_intl_transport.mz
> 216K core/input/grid/EDGAR_CO2.mz
> 212K core/input/grid/EDGAR_SO2_2005_excl_agr_and_intl_transport.mz
> 212K core/input/grid/EDGAR_SO2.mz
> 204K config/scenario_config_21_EU11_Fit_for_55_sensitivity.csv
> 200K tutorials/figures/git-7-pull-request-github-1.PNG
> 172K tutorials/figures/git-8-pull-request-github-2.PNG
> 148K tutorials/figures/appResults_window.png
> 120K core/sets.gms
> 112K scripts/output/comparison/notebook_templates/AriadneComparison.Rmd
> 108K main.gms
> ```
> 
> I feel it should be easy to move some stuff to `mrremind` and go to maybe 250 KB.
> 
> Also, could this not be put under the `.github/` directory? The Remind directory is cluttered as it is …

regarding the biggest file: I would suggest to delete the whole realization https://github.com/remindmodel/remind/tree/develop/modules/31_fossil/exogenous @nicobauer? 
regarding the EDGE-T data: I will talk to the transport people

fbenke-pik:
Ok, since this is a running solution, I'd say we merge this for now and remove the threshold later as we delete/move files. 

I am open to more elegant solutions replacing this later, but feel like this simple solution works well for now.

LaviniaBaumstark:
> Ok, since this is a working solution, I'd say we merge this for now and lower the threshold later as we delete/move files.
> 
> I am open to more elegant solutions replacing this later, but feel like this simple solution works well for now.

I agree

LaviniaBaumstark (approved):

fbenke-pik:
I see, the workflow is triggered by [push](https://docs.github.com/en/webhooks/webhook-events-and-payloads#push) and [pull request](https://docs.github.com/en/webhooks/webhook-events-and-payloads#pull_request) events. 

So I'd say, it would not be a proper commit hook in any case with the current solution.

PR #1457, opened 2023-11-09
enable management of local calibration results

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
For production runs it is generally advisable to calibrate first, since establishing what exactly went into a given calibration results is an inherently manual process that is not guaranteed to yield definitive results.
This merge request adds a `set-local-calibration` target to the Makefile that implements [this tutorial](https://gitlab.pik-potsdam.de/pehl/tuts/-/blob/master/Using_local_REMIND_calibrations.md) (with some improvements).
Users will get a `calibration_results/` directory with a `collect_calibration` script that can be used to create `.tgz` files from calibration results (just like at `/p/projects/remind/inputdata/CESparametersAndGDX/`) and will have the directory included as a `remind_repos` option in their R session.  Use of local calibration results in REMIND from there on is identical to that of provided calibration results – simply set the `CESandGDXversion` accordingly.

- [x] New feature 

## Checklist:

- [x] My code follows the [coding etiquette](https://github.com/remindmodel/remind/blob/develop/main.gms#L80)
- [ ] I performed a self-review of my own code
- [x] I explained my changes within the PR, particularly in hard-to-understand areas
- [x] I checked that the [in-code documentation](https://github.com/remindmodel/remind/blob/develop/main.gms#L120) is up-to-date
- [x] I adjusted the reporting in [`remind2`](https://github.com/pik-piam/remind2) where it was needed
- [x] I adjusted `forbiddenColumnNames` in [readCheckScenarioConfig.R](https://github.com/remindmodel/remind/blob/develop/scripts/start/readCheckScenarioConfig.R) in case the PR leads to deprecated switches
- [x] All automated model tests pass (`FAIL 0` in the output of `make test`)
- [x] The changelog `CHANGELOG.md` [has been updated correctly](https://gitlab.pik-potsdam.de/rse/rsewiki/-/wikis/Standards-for-Writing-a-Changelog)

PR #1505, opened 2023-12-08
First draft of updated Running REMIND locally tutorial

tonnrueter:
## Purpose of this PR

Extends tutorial Section 2a: Running REMIND locally

### DIFF
--- a/tutorials/02a_RunningREMINDLocally.md
+++ b/tutorials/02a_RunningREMINDLocally.md
@@ -5,7 +5,7 @@ Mika Pflüger (mika.pflueger@pik-potsdam.de), Tonn Rüter (tonn.rueter@pik-potsd
 
 ## System Requirements
 
-- Installed python version must be `3.10.X` or older. To check your python version run `python --version` in a terminal of your choice
+- Installed python version must be `< 3.11`. We recommend a python version `3.10.X` as it is closest to being up-to-date while remaining REMIND compatible. To check your python version run `python --version` in a terminal of your choice
 - GAMS must be installed and the path to GAMS applications needs to be in your PATH environment variable
 - R version must be at least `4.2.X`. To check your R version run `R --version` in a terminal of your choice
 - On Windows we recommend to use PowerShell rather than the standard command prompt
@@ -139,25 +139,25 @@ then restart your run.
 In case your REMIND run fails and you wish to be supported by REMIND RSE in getting it to run properly, please help us by recording the logging output of the run. On linux, start your run via
 
 ```bash
-Rscript start.R &>> ~/failed_REMIND_run.log
+Rscript start.R &>> failed_REMIND_run.log
 ```
 
 or on Windows via
 
 ```PowerShell
-Rscript.exe .\start.R *>> ~/failed_REMIND_run.log
+Rscript.exe .\start.R *>> failed_REMIND_run.log
 ```
 
-This writes *all* output from your run to a file in your home directory called `failed_REMIND_run.log`. After hitting enter in your terminal session, none of the usual output will show up since it is being redirected into the aforementioned file. You can still follow the progress of your run by opening a second terminal window and type in
+This writes *all* output from your run to a file in your `<REMIND_DIR>` directory called `failed_REMIND_run.log`. After hitting enter in your terminal session, none of the usual output will show up since it is being redirected into the aforementioned file. You can still follow the progress of your run by opening a second terminal window and type in
 
 ```bash
-tail -n 25 -f ~/failed_REMIND_run.log
+tail -n 25 -f <REMIND_DIR>/failed_REMIND_run.log
 ```
 
 on Linux or
 
 ```PowerShell
-Get-Content ~/failed_REMIND_run.log -Tail 10 -Wait
+Get-Content <REMIND_DIR>/failed_REMIND_run.log -Tail 10 -Wait
 ```
 
 on Windows. In case you need to interact with the run script during execution (ie. when selecting the run mode) you can still do so in the terminal session from which you started the run.
--- a/tutorials/02a_RunningREMINDLocally.md
+++ b/tutorials/02a_RunningREMINDLocally.md
@@ -5,10 +5,13 @@ Mika Pflüger (mika.pflueger@pik-potsdam.de), Tonn Rüter (tonn.rueter@pik-potsd
 
 ## System Requirements
 
+Check the general system requirements in [Section 1: Getting REMIND](01_GettingREMIND.md).
+
 - Installed python version must be `< 3.11`. We recommend a python version `3.10.X` as it is closest to being up-to-date while remaining REMIND compatible. To check your python version run `python --version` in a terminal of your choice
-- GAMS must be installed and the path to GAMS applications needs to be in your PATH environment variable
-- R version must be at least `4.2.X`. To check your R version run `R --version` in a terminal of your choice
-- On Windows we recommend to use PowerShell rather than the standard command prompt
+- There are some additional recommendations for running REMIND locally on a **Windows systems**:  
+  - To avoid lengthy compilation times during preparation for your remind run, we recommend to use a version of R for which most R-packages that REMIND depends on, are available as binary packages. This is typically the case for recent versions of R, currently `4.2.X` or `4.3.X`. To check your R version run `R --version` in a terminal of your choice
+  - As mentioned in the general system requirements, GAMS must be installed. On Windows, make sure that the path to GAMS applications is in your PATH environment variable
+  - We recommend to use PowerShell rather than the standard command prompt
 
 ## Getting Input Data
 
--- a/tutorials/02a_RunningREMINDLocally.md
+++ b/tutorials/02a_RunningREMINDLocally.md
@@ -74,6 +74,32 @@ REMIND_repos_scp_key="/home/myusername/.ssh/id_ed25519"
 
 Make sure to use your username on the cluster and the correct path to your private ssh key (might also be named `id_rsa` or something similar starting with `id_`).
 
+### Tell REMIND where to find MAGICC
+
+REMIND needs to be told where to find the configuration files of the 'Model for the Assessment of Greenhouse Gas Induced Climate Change' ([MAGICC](https://magicc.org/)). The config files all come with REMIND, so we just need to set the appropriate location in the REMIND default config file at `<REMIND_DIR>/config/default.cfg`. Open the file with a text editor of your choice and locate the variable `magicc_template`. In the `default.cfg` file, the corresponding section will look like
+
+```R
+cfg$magicc_template <- "/p/projects/rd3mod/magicc/"
+```
+
+Change it to the fully qualified file path (i.e. a file path without double points `..` or the tilde `~`) of the MAGICC configuration files. These are located at `<REMIND_DIR>/core/magicc`. The resulting line should look like:
+
+```R
+cfg$magicc_template <- "<REMIND_DIR>/core/magicc/"
+```
+
+On Linux the fully qualified file path will look similar to:
+
+```R
+cfg$magicc_template <- "/home/<YOUR USERNAME>/REMIND/core/magicc"
+```
+
+On Windows remember to also escape the backslashes:
+
+```R
+cfg$magicc_template <- "C:\\Users\\<YOUR USERNAME>\\REMIND\\core\\magicc"
+```
+
 ## Start a run
 
 In a terminal session change directory to the location where REMIND was cloned via `cd <REMIND_DIR>` and start a run as you would do on the cluster. As a linux user, type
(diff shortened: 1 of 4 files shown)


tutorials/02a_RunningREMINDLocally.md
```diff
@@ -1,10 +1,146 @@
 # Running REMIND locally
-Mika Pflüger (mika.pflueger@pik-potsdam.de)
+Mika Pflüger (mika.pflueger@pik-potsdam.de), Tonn Rüter (tonn.rueter@pik-potsdam.de)
 
-*Note:* This tutorial is a work in progress, not all steps needed are described, yet.
+*Note:* This approach of running REMIND is a work in progress, however successful REMIND runs on local machines have been conducted using this tutorial.
+
+## System Requirements
+
+- Installed python version must be `3.10.X` or older. To check your python version run `python --version` in a terminal of your choice
```
0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q, line 8: Based on what?
Until now, I was only aware on a [3.7 <= $x$ < 3.11 dependency](https://github.com/remindmodel/remind/pull/1466), inherited from `climate-assessment` → `pyam-iamc`.
So both restrictions (here and in the `Makefile`) might contract to just 3.10

tutorials/02a_RunningREMINDLocally.md
```diff
@@ -1,10 +1,146 @@
...
+- GAMS must be installed and the path to GAMS applications needs to be in your PATH environment variable
+- R version must be at least `4.2.X`. To check your R version run `R --version` in a terminal of your choice
```
0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q, line 10: Where does that come from?
We should update [the `DESCRIPTION` file](https://github.com/remindmodel/remind/blob/9f6a2cb01f19b783dd8823ae0e6616c0c658962d/DESCRIPTION#L12) then.

tutorials/02a_RunningREMINDLocally.md
```diff
@@ -1,10 +1,146 @@
...
+
+```bash
+# REMIND data setup
+# If the input data is spread over multiple directories,
+# separate them with a : on linux, a ; on windows
+REMIND_repos_dirs="/my/first/path:/my/second/path"
+# On windows, it would look something like
+# REMIND_repos_dirs="C:\Users\<your windows username>\REMIND;D:\REMIND"
+# depending on your used file paths
+```
```
0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q, line 37: As of #1457, user could also just run `make set-local-calibration` and then drop their files (both input data and CES gdxes) into `./calibration_results/`, which is included by default via [`.Rprofile`](https://github.com/remindmodel/remind/blob/9f6a2cb01f19b783dd8823ae0e6616c0c658962d/.Rprofile#L117-L123).
This might be easier if setting up just a single directory to run REMIND in.  For multiple directories a central location would be preferable, of course.

tutorials/02a_RunningREMINDLocally.md
```diff
@@ -1,10 +1,146 @@
...
+
+then restart your run.
+
+### Support RSE with Extensive Log Files
+
+In case your REMIND run fails and you wish to be supported by REMIND RSE in getting it to run properly, please help us by recording the logging output of the run. On linux, start your run via
+
+```bash
+Rscript start.R &>> ~/failed_REMIND_run.log
+```
```
0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q, line 110: Why not to a file in the local directory?

tutorials/02a_RunningREMINDLocally.md
```diff
@@ -1,10 +1,146 @@
...
+
+```PowerShell
+Rscript.exe .\start.R *>> ~/failed_REMIND_run.log
+```
+
+This writes *all* output from your run to a file in your home directory called `failed_REMIND_run.log`. After hitting enter in your terminal session, none of the usual output will show up since it is being redirected into the aforementioned file. You can still follow the progress of your run by opening a second terminal window and type in
+
+```bash
+tail -n 25 -f ~/failed_REMIND_run.log
+```
```
0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q, line 122: Or
```
Rscript start.R 2>&1 | tee ~/failed_REMIND_run.log
```

tutorials/02a_RunningREMINDLocally.md
```diff
@@ -1,10 +1,146 @@
...
+
+As long as we do not provide a curated collection of input data please obtain these files from the cluster:
+
+```bash
+rev6.606_62eff8f7_remind.tgz
+rev6.606_62eff8f7_validationremind.tgz
+CESparametersAndGDX_878ac5d69254efb4eba5c1fa39aba64000307bb1.tgz
+```
+
+The contents of these tar-archives are necessary to run REMIND locally. Since REMIND input data is constantly updated the up-to-date file names may vary, but will exhibit a similar pattern. The first two files can be found on the cluster at `/p/projects/rd3mod/inputdata/output`, the latter file in `/p/projects/remind/inputdata/CESparametersAndGDX`.
```
fbenke-pik, line 25: Mention also: Make sure that the configuration parameters `inputRevision` and `CESandGDXversion`  in your `config/default.cfg` match the file names.

tutorials/02a_RunningREMINDLocally.md
```diff
@@ -1,10 +1,146 @@
...
+*Note:* This approach of running REMIND is a work in progress, however successful REMIND runs on local machines have been conducted using this tutorial.
+
+## System Requirements
+
+- Installed python version must be `3.10.X` or older. To check your python version run `python --version` in a terminal of your choice
+- GAMS must be installed and the path to GAMS applications needs to be in your PATH environment variable
+- R version must be at least `4.2.X`. To check your R version run `R --version` in a terminal of your choice
+- On Windows we recommend to use PowerShell rather than the standard command prompt
+
+## Use Local Input Data
```
fbenke-pik, line 13: Should be structured as follows

## Getting Input Data

### Use Local Input Data
...

### Download Input Data Automatically
....

tutorials/02a_RunningREMINDLocally.md
fbenke-pik: This would be a third option, then. Tonn, can you try this approach and see if it works for you?  You can also check out [chapter 12 ](https://github.com/remindmodel/remind/blob/develop/tutorials/12_Calibrating_CES_Parameters.md#results)  to get a more detailed description of this approach.

tutorials/02a_RunningREMINDLocally.md
```diff
@@ -1,10 +1,146 @@
...
+
+or as a Windows user, type
+
+```PowerShell
+Rscript.exe .\start.R
+```
+
+## Known Issues & Further Debugging
+
+### MAGCFG_STORE File Path Resolution Fails in `prepare.R`
```
fbenke-pik, line 57: This point should be solved by setting
`cfg$magicc_template` in `default.cfg` to  <- "[PATH TO REMIND]/core/magicc/"

tutorials/02a_RunningREMINDLocally.md
tonnrueter: Based on the `climate-assesment` / `pyam-iamc` dependency. I think python `3.10` would be a reasonable choice

tutorials/02a_RunningREMINDLocally.md
tonnrueter: make is not available on Windows systems

tutorials/02a_RunningREMINDLocally.md
tonnrueter: The log-file does not belong to the REMIND code base, nor is it relevant for REMIND execution. It is a file specific to the REMIND user. Rather than locating it amongst the REMIND files, I therefore propose to place it in their home directory

tutorials/02a_RunningREMINDLocally.md
tonnrueter: Would this allow for user interaction with the REMIND/Rscript process?

tutorials/02a_RunningREMINDLocally.md
0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q: Points well taken.

The thing is: this is a replication and as such a source of inconsistency.  An undiscerning user might look at this tutorial, and never the `DESCRIPTION` file, which is supposed to contain the "real" dependencies.
I would argue for telling people to run `make ensure-reqs` (or `Rscript -e 'source("scripts/start/ensureRequirementsInstalled.R"); ensureRequirementsInstalled(rerunPrompt="make ensure-reqs")'` if they do not have that) and note here that things are easier with `R 4.2` on Windows (on Linux, everything is installed from source anyway).

I wonder: who is the target audience for this tutorial, i.e. who wants to run REMIND locally (and on Windows)? @LaviniaBaumstark 

If that is "REMIND team members working on their laptop instead of the cluster", we could reasonably expect them to get a work environment that includes `make` (and support them in doing so), since the tests they are expected to perform in the course of their work depend on it.

If that is "REMIND team members" plus "outside interested parties", we might have a "default" description for a full work environment and a striped-down version for those who just want to dabble.

-----

> Updating the DESCRIPTION file to R (> 4.1.3) would therefore break the REMIND cluster deployment

Nothing that could not be overcome with one edit by RSE …
```
$ module --terse available 2>&1 | grep 'R/4.2'
R/4.2.0/gcc-mkl
```
;)

tutorials/02a_RunningREMINDLocally.md
0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q: Sure does.

tutorials/02a_RunningREMINDLocally.md
0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q: There can be all kind of junk in the REMIND directory not required (anymore).  That is why `.log` files are ignored by git.
I think it is weird to put a log about something somebody is working on in one directory into another directory.

tutorials/02a_RunningREMINDLocally.md
tonnrueter: Done

tutorials/02a_RunningREMINDLocally.md
fbenke-pik: From https://github.com/remindmodel/remind/blob/develop/tutorials/11_ManagingRenv.md
> make is pre-installed except on Windows, run winget install gnuwin32.make to install it there

Does that work for you?

fbenke-pik: If we add it as an option, it would be a third option (not replacing this option), and it can be described by referencing the description in the Results Section of the calibration chapter.

tutorials/02a_RunningREMINDLocally.md
tonnrueter: I refer to a scenario in which the user has to select the run type by putting in a number. `tee` would not relay the user input to `Rscript` and the run would get stuck. But that scenario might not even happen with running REMIND on a local machine, so I'm happy with the `tee` based solution as well

tutorials/02a_RunningREMINDLocally.md
fbenke-pik: I agree with Michaja, it is common practice to dump all the logs into the REMIND folder, so let's stick to that.

tutorials/02a_RunningREMINDLocally.md
tonnrueter: Good point. I'll change it to 

Installed python version must be `< 3.11`. We recommend a python version `3.10.X` as it is closest to being up-to-date while remaining REMIND compatible.

tutorials/02a_RunningREMINDLocally.md
0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q: Why should `tee` relay the input?  It has not been redirected.

tutorials/02a_RunningREMINDLocally.md
tonnrueter: Done

tutorials/02a_RunningREMINDLocally.md
fbenke-pik: Maybe just create a separate section with additional remarks for Windows users. There, you can mention that one should better use an R version which still offers binaries, as you probably don't want to compile the R libs on your Windows machine.

Then, you can simply refer to the requirements in Chapter 01 Getting REMIND.

fbenke-pik: > I wonder: who is the target audience for this tutorial, i.e. who wants to run REMIND locally (and on Windows)? @LaviniaBaumstark

We do have a request for this.
I agree, Windows should not be the recommended OS to use. This is why I would move all the Windows-specific recommendations to a separate section at the bottom, starting with sth along the lines of "We do not recommend it, but if you must.."

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q: > you can mention that one should better use an R version which still offers binaries, as you probably don't want to compile the R libs on your Windows machine.

Amazing that we assume people wanting to run REMIND cannot be troubled to wait half an hour to set up the required environment.

tutorials/02a_RunningREMINDLocally.md
tonnrueter: Exactly. Thus I recommend to use two separate terminal sessions instead of piping the output to `tee`

tutorials/02a_RunningREMINDLocally.md
tonnrueter: I updated the whole section to emphasize that most of the System Requirements only concern those willing to run REMIND locally *under Windows*. Is this fine with you, @0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q ?

tutorials/02a_RunningREMINDLocally.md
tonnrueter: Works 👍 

I removed this from the debugging section and added another in which I explain how to adapt the `default.cfg` file

fbenke-pik (approved):

PR #1515, opened 2024-01-02
code clean on switches

cchrisgong:
## Purpose of this PR
update documentations according to https://pad.systemli.org/p/remind_documentation-keep
mostly only doc string changes, minimal code changes

code changes:
1) bug fix on deepEl switch, putting back missing cm_VRE_supply_assumptions = 2 and cm_VRE_supply_assumptions = 3 options
2) bug fix / reduce confusion on line [core/datainput.gms](https://github.com/remindmodel/remind/pull/1515/files#diff-328de481dd31565bdd16ff7586e7d89a354d406b3a43cdde4df0f6613e9439d1). So now contradictory codes which were previously overwritten by later line is no longer in contradiction. 
```
$ifthen.REG_techcosts "%cm_techcosts%" == "REG"   !! cm_techcosts
*** for those technologies, for which differentiated costs are available for
*** 2015-2020, use those
loop(te$( teNoLearn(te) AND teRegTechCosts(te) ),
  !! no value after 2020 is currently used (see convergence below)
  pm_inco0_t(ttot,regi,te)$( ttot.val ge 2015 AND ttot.val lt 2040)
  = p_inco0(ttot,regi,te);
```
is overwritten by the 2025-2040 values of global convergence in prices later, so making this period less than 2040 or 2025 is the same
3) remove obsolete switch c_solscen

## Type of change

- [ ] Bug fix 
- [ ] Refactoring
- [ ] New feature 
- [X ] Minor change (default scenarios show only small differences)
- [ ] Fundamental change

## Checklist:

- [X ] My code follows the [coding etiquette](https://github.com/remindmodel/remind/blob/develop/main.gms#L80)
- [X ] I performed a self-review of my own code
- X ] I explained my changes within the PR, particularly in hard-to-understand areas
- [X ] I checked that the [in-code documentation](https://github.com/remindmodel/remind/blob/develop/main.gms#L120) is up-to-date
- [X ] I adjusted the reporting in [`remind2`](https://github.com/pik-piam/remind2) where it was needed
- [X ] I adjusted `forbiddenColumnNames` in [readCheckScenarioConfig.R](https://github.com/remindmodel/remind/blob/develop/scripts/start/readCheckScenarioConfig.R) in case the PR leads to deprecated switches
- [X ] All automated model tests pass (`FAIL 0` in the output of `make test`)
- [X ] The changelog `CHANGELOG.md` [has been updated correctly](https://gitlab.pik-potsdam.de/rse/rsewiki/-/wikis/Standards-for-Writing-a-Changelog)

* Test runs are here: 


core/bounds.gms
```diff
@@ -217,7 +217,7 @@ if (cm_nucscen eq 5,
 );
 
 *** -------------------------------------------------------------
-*** *DK* Phaseout of 1st generation biofuel technologies
+*** *DK* Force no new capacities of 1st generation biofuel technologies to be installed after 2030, allowing more cost-efficient new generation of biofuel technologies free entrance to the market
```
merfort, line 220:
```suggestion
*** *DK* Force no new capacities of 1st generation biofuel technologies to be
*** installed after 2030, allowing more cost-efficient and more sustainable new
*** generation of biofuel technologies free entrance to the market
```

main.gms
```diff
@@ -260,7 +260,7 @@ $setGlobal climate  off               !! def = off
 *'---------------------    16_downscaleTemperature    --------------------------
 *'
 *' * (off)
-*' * (CMIP5): downscale GMT to regional temperature based on CMIP5 data (between iterations, no runtime impact). Requires climate= 'off' and cm_rcp_scen=none"iterative_target_adj" = 9] curved convergence of CO2 prices between regions until cm_CO2priceRegConvEndYr; developed countries have linear path from 0 in 2010 through cm_co2_tax_2020 in 2020;
+*' * (CMIP5): downscale GMT to regional temperature based on CMIP5 data (between iterations, no runtime impact). [Requires climate= off, cm_rcp_scen=none, iterative_target_adj = 9] curved convergence of CO2 prices between regions until cm_CO2priceRegConvEndYr; developed countries have linear path from 0 in 2010 through cm_co2_tax_2020 in 2020;
```
merfort, line 263:
```suggestion
*' * (CMIP5): downscale GMT to regional temperature based on CMIP5 data (between iterations, no runtime impact). [Requires climate = off, cm_rcp_scen = none, iterative_target_adj = 9] curved convergence of CO2 prices between regions until cm_CO2priceRegConvEndYr; developed countries have linear path from 0 in 2010 through cm_co2_tax_2020 in 2020;
```
```suggestion
*' * (CMIP5): downscale GMT to regional temperature based on CMIP5 data (between iterations, no runtime impact). [Requires climate= off, cm_rcp_scen=none, iterative_target_adj = 9] curved convergence of CO2 prices between regions until cm_CO2priceRegConvEndYr; developed countries have linear path from 0 in 2010 through cm_co2_tax_2020 in 2020;
```

main.gms
```diff
@@ -633,11 +632,11 @@ parameter
...
 *'
 parameter
-  cm_1stgen_phaseout        "choose if 1st generation biofuels should phase out after 2030 (vm_deltaCap equals 0)"
+  cm_1stgen_phaseout        "scenario setting for phase-out of 1st generation biofuels"
 ;
   cm_1stgen_phaseout  = 0;         !! def = 0  !! regexp = 0|1
-*' *  (0): 1st generation biofuels after 2020 are fixed at upper limit of resource potential (maxprod)
-*' *  (1): no new capacities for 1st generation biofuel technologies may be built after 2030 -> phaseout until ~2060
+*' *  (0): no phase-out. Production of 1st generation biofuels after 2045 is bound from below by 90% of maximum resource potential ("maxprod")
+*' *  (1): phase-out. No new capacities for 1st generation biofuel technologies are built after 2030 (i.e. added capacity vm_deltaCap equals 0), in practice this means a slow phaseout of 1st generation biofuel due to lack of economic competitiveness. Bioenergy production is bound from below by 90% of maximum biomass resource potential in 2045
```
merfort, line 639: I am not sure about that. @dklein-pik do you remember, what we discussed there? In the phase-out scenario (1) the description of the second part seems to contradict the phase-out. We can't have a phase-out and a lower bound (90% of potential) at the same time, right?

main.gms
```diff
@@ -814,17 +813,17 @@ parameter
...
-*' * (6): iterative adjustment VI for both budget and tax runs based on peak CO2 emissions budget, without changing temporal profile (i.e. with overshoot), see core/postsolve.gms for direct algorithms of adjustment
-*' * (7): iterative adjustment VII for tax runs based on peak CO2 emissions, with change of temporal tax profile after time of peak budget, aiming for net-zero thereafter, see core/postsolve.gms for direct algorithms of adjustment
-*' * (9): [REMIND 2.1 default for validation peakBudget runs, in combination with carbonprice = none; after the peaking year annual increase by c_taxCO2inc_after_peakBudgYr. Automatically shifts c_peakBudgYr to find the correct peaking year for a given .
+  cm_iterative_target_adj = 0;      !! def = 0  !! regexp = 0|2|3|4|5|6|7|9
+*' * (0): no iterative adjustment of CO2 price (terminology: CO2 price and CO2 tax in REMIND is used interchangeably)
+*' * (2): iterative adjustment of CO2 tax based on climate forcing calculated by climate model magicc (for runs with budget and/or CO2 tax constraints), see ./modules/15_climate/magicc/postsolve.gms for direct algorithm
```
strefler, line 820: suggest to change to "iterative adjustment of CO2 tax or emission budget based on climate forcing calculated by climate model magicc, see ./modules/15_climate/magicc/postsolve.gms for direct algorithmfor runs with budget or CO2 tax constraints"
would use the first part for the following lines as well, as we always constrain either budget or tax, but cannot constrain both at the same time

main.gms
```diff
@@ -814,17 +813,17 @@ parameter
...
-*' * (5): iterative adjustment V for both budget and tax runs based on CO2 emissions 2020-2100, see core/postsolve.gms for direct algorithms of adjustment
-*' * (6): iterative adjustment VI for both budget and tax runs based on peak CO2 emissions budget, without changing temporal profile (i.e. with overshoot), see core/postsolve.gms for direct algorithms of adjustment
-*' * (7): iterative adjustment VII for tax runs based on peak CO2 emissions, with change of temporal tax profile after time of peak budget, aiming for net-zero thereafter, see core/postsolve.gms for direct algorithms of adjustment
-*' * (9): [REMIND 2.1 default for validation peakBudget runs, in combination with carbonprice = none; after the peaking year annual increase by c_taxCO2inc_after_peakBudgYr. Automatically shifts c_peakBudgYr to find the correct peaking year for a given .
+  cm_iterative_target_adj = 0;      !! def = 0  !! regexp = 0|2|3|4|5|6|7|9
+*' * (0): no iterative adjustment of CO2 price (terminology: CO2 price and CO2 tax in REMIND is used interchangeably)
```
strefler, line 819: suggest to change to CO2 tax - I think we mainly use tax instead of price in in-code comments

main.gms
```diff
@@ -814,17 +813,17 @@ parameter
...
+*' * (2): iterative adjustment of CO2 tax based on climate forcing calculated by climate model magicc (for runs with budget and/or CO2 tax constraints), see ./modules/15_climate/magicc/postsolve.gms for direct algorithm
+*' * (3): [requires 45_carbonprice = NDC and emiscen = 9] iterative adjustment of CO2 tax based on 2025 or 2030 regionally differentiated emissions, see ./modules/45_carbonprice/NDC/postsolve.gms for direct algorithm
+*' * (4): iterative adjustment of CO2 tax based on CO2 FF&I emission budget (i.e. without landuse emissions) (2020-2100), see core/postsolve.gms for direct algorithms
+*' * (5): iterative adjustment of CO2 tax based on economy-wide CO2 emissions (2020-2100), see core/postsolve.gms for direct algorithms
+*' * (6): iterative adjustment of CO2 tax based on economy-wide CO2 emission peak budget, features: without changing temporal profile (i.e. with overshoot), see core/postsolve.gms for direct algorithms
```
strefler, line 824: What does "without changing temporal profile (i.e. with overshoot)" mean? We never change the temporal profile of the CO2 tax, that's valid for all realizations. As this is a peak budget realization, there is no overshoot.

main.gms
```diff
@@ -814,17 +813,17 @@ parameter
...
+*' * (7): iterative adjustment of CO2 tax based on economy-wide CO2 emission peak budget, features: with a change of temporal tax profile after time of peak budget, aiming for net-zero thereafter, see core/postsolve.gms for direct algorithms
+*' * (9): [require carbonprice = diffCurvePhaseIn2Lin] iteratively obtain endogenous global CO2 price under peak budget constraint, option IX. Features: 1) after the year when budget peaks, CO2 tax has an annual increase by c_taxCO2inc_after_peakBudgYr, 2) automatically shifts c_peakBudgYr to find the correct year of budget peaking for a given budget. For REMIND version v2.1 or above.
```
strefler, line 826: I'm not sure this is true. The comments in the code say that it works with several CO2 price path realizations of module 45
suggest to delete "option IX"
the "global CO2 price" is confusing, as the regional CO2 prices might differ. Suggest to use the same formulation as in the other realizations: iterative adjustment of CO2 tax based on economy-wide CO2 emission peak budget"

core/postsolve.gms
```diff
@@ -23,7 +23,7 @@ $endif.neg
 );
 
 if(cm_iterative_target_adj eq 4,
-*JeS* Update tax levels/ multigasbudget values to reach the CO2 FF&I budget (s_actualbudgetco2 runs from 2020-2100)
+*JeS* Update tax levels / cumulative GHG emission level (CO2+CH4+N2O, in CO2eq units) to reach the CO2 FF&I budget (s_actualbudgetco2 runs from 2020-2100, not peak budget)
```
strefler, line 26: We update either the emission tax or the emission budget, which can be a CO2 only or a multigas budget, depending on cm_multigasscen. Therefore, I would not specify here, but simply say "Update emission tax levels / budgets to reach the CO2 FF&I budget". Same for the following lines

strefler (changes requested):

main.gms
cchrisgong: yeah I am also a bit confused by this, but maybe Felix(?) @fschreyer implemented this bound because it conflicted with EDGE-T? https://github.com/remindmodel/remind/blob/develop/modules/30_biomass/magpie_40/bounds.gms#L24

main.gms
cchrisgong: thanks! fixed in the new commit

main.gms
cchrisgong: thanks! fixed in the new commit

main.gms
cchrisgong: thanks! fixed in the new commit

core/postsolve.gms
cchrisgong: thanks! fixed in the new commit

I use cumulative emission level to mean what the model produces in-iteration, and reserve the word "budget" for the parameter which is the cumulative emission target/emission budget

main.gms
cchrisgong: hmmm, I simply copied over the original text from main.gms https://github.com/remindmodel/remind/blob/develop/main.gms#L825C1-L825C16, I also don't know what the temporal profile or overshoot comments are referring to... do you know who wrote this originally? Even though I don't know what it means, it seems to be important, as it is the only thing that differs from setting=7 (I believe?)

how should I best describe 6 or 7?

core/postsolve.gms
```diff
@@ -23,7 +23,7 @@ $endif.neg
 );
 
 if(cm_iterative_target_adj eq 4,
-*JeS* Update tax levels/ multigasbudget values to reach the CO2 FF&I budget (s_actualbudgetco2 runs from 2020-2100)
+*JeS* Update tax levels / cumulative emission level to reach the FF&I budget target (s_actualbudgetco2 runs from 2020-2100, not peak budget)
```
strefler, line 26: I understand the idea of having different words for the current budget (=cumulative emission level) and the target, but it doesn't make sense to me to "update cumulative emission levels" as we can only update the target. That's why I suggested tax levels / budgets. Also, I would keep the "CO2" in "CO2 FF&I budget"

core/postsolve.gms
```diff
@@ -54,7 +54,7 @@ display s_actualbudgetco2;
 );
 
 if(cm_iterative_target_adj eq 5,
-*JeS* Update tax levels/ multigasbudget values to reach the CO2 budget (s_actualbudgetco2 runs from 2020-2100)
+*JeS* Update CO2 tax trajectories / cumulative emission level to reach the budget target (s_actualbudgetco2 runs from 2020-2100, not peak budget)
```
strefler, line 57: would stick with the same wording for all realizations, i.e. tax levels / budgets

strefler (changes requested):

core/postsolve.gms
cchrisgong: okay, then I think I don't quite understand how this works...  as I understand, the algorithm raises the in-iteration CO2 tax with some function, and then with this new CO2 price, the model finds an optimum and in postsolve, we can calculate the global cumulative emission level from the energy system under the new CO2 price trajectories (for all regions), and the price traj. can be regionally differentiated. And then the algo compares this cumulative level to the target, which I call the budget (like a household budget, which is fixed, as I assume my income is fixed, so the budget is not something I can adjust(?) at least intuitively I understand this as such). And then depends on the difference between cum. emission level and the target, the algo changes the CO2 price traj. again, and so on

is this roughly how it works? or do we update the budget? i.e if we put c_budgetCO2from2020 = 500, we do not adjust this number, right?

on CO2 FF&I budget -> is this because FF&I target is defined as having no methane, etc? sorry I am not familiar with the definition here, if it is UNFCCC/IPCC definition, maybe we can even link this here?

main.gms
fschreyer: Well, I honestly do not remember the exact context in which those bounds were necessary. I think, I only [shifted them](https://github.com/remindmodel/remind/commit/42f09d2f7398cf9ac95cc3f1572f955aedeb80a9). Currently, it seems the vm_deltaCap is forced to zero from 2035 and vm_fuExtr is between 90-100% of the potential in 2045 for  both 0 and 1 values of the switch (and the lower bound continues after 2045 for the non-phaseout case). For the phase-out case, the 2045 90% lower bound does not make sense to me.

merfort: Ok, I think, this switch needs some polishing content-wise. I just did two test runs (Baseline) to check, what it is doing and apparently the "phaseout = off" case actually leads to more bioenergy in the short term.
pe_biomass_1st_generation

cchrisgong: @merfort the last comment, you mean phaseout = On (green line) right?

I can leave the comment string as is, as it describes the current state of the code. When you do a patch, can tag me and I can take a look at the description again, if you want :)

merfort: Ok, if `vm_deltaCap` is forced to be zero from 2035 on, that explains why the ramp up needs to be so quick in the phase-out scenario, when there is still that lower bound in 2045.

I will prepare a patch.

core/bounds.gms
```diff
@@ -217,7 +217,9 @@ if (cm_nucscen eq 5,
 );
 
 *** -------------------------------------------------------------
-*** *DK* Phaseout of 1st generation biofuel technologies
+*** *DK* Force no new capacities of 1st generation biofuel technologies to be
```
LaviniaBaumstark, line 220: if we delete *DK*, I think this could go into iur in code docu, right?

core/bounds.gms
```diff
@@ -229,7 +231,7 @@ if(cm_1stgen_phaseout=1,
 *** -----------------------------------------------------------
 
 if (cm_startyear gt 2005,
-  if (c_solscen eq 3,
+  if (c_solscen eq 2,
```
LaviniaBaumstark, line 234: was this a bug? Or did you change the meaning of a switch number?

core/datainput.gms
```diff
@@ -1209,10 +1203,10 @@ display pm_inco0_t;
 
 $ifthen.REG_techcosts "%cm_techcosts%" == "REG"   !! cm_techcosts
 *** for those technologies, for which differentiated costs are available for
-*** 2015-2040, use those
+*** 2015-2020, use those
 loop(te$( teNoLearn(te) AND teRegTechCosts(te) ),
   !! no value after 2020 is currently used (see convergence below)
-  pm_inco0_t(ttot,regi,te)$( ttot.val ge 2015 AND ttot.val lt 2040 )
+  pm_inco0_t(ttot,regi,te)$( ttot.val ge 2015 AND ttot.val lt 2025)
```
LaviniaBaumstark, line 1209: why did you change this year? I have no idea if this is o.k.

core/declarations.gms
```diff
@@ -48,6 +48,7 @@ f_lab(tall,all_regi,all_POPscen)                     "labour data for all possib
 pm_lab(tall,all_regi)                                "data for labour [bn people]"
 pm_esCapCost(tall,all_regi,all_teEs)                 "Capital energy cost per unit of consumption for end-use capital (energy service layer)"
 pm_cesdata_sigma(ttot,all_in)                        "elasticities of substitution"
+*' If elasticities of substitution (sigma) is below 1, the smaller it is the less the substitution replacement effect between different CES nodes. The products become more and more complementary in the production, meaning that the more one product is used, the more the other one is demanded as well. If sigma is larger than 1, the more one product is used, the less the others are used, i.e. the products are substitutes
```
LaviniaBaumstark, line 51: this comment should go next to an equation or somewhere else and not be in declarations.gms

main.gms
```diff
@@ -585,10 +586,9 @@ parameter
 parameter
   c_solscen                 "solar option choice"
 ;
-  c_solscen        = 1;        !! def = 1  !! regexp = [1-3]
+  c_solscen        = 1;        !! def = 1  !! regexp = [1-2]
 *' *  (1): yes
-*' *  (2): no solar
-*' *  (3): fix at bau level
+*' *  (2): fix at bau level
```
LaviniaBaumstark, line 591: can be very dangerous, if someone is using old scenario_config files. Keep the number for the same meaning. If you delete a number, we can get gaps but that is o.k.

LaviniaBaumstark (changes requested):
Did you generate an in-code docu based on your changes and checked, that everything looks o.k.?

core/bounds.gms
cchrisgong: yes, should I delete "*DK*"? I think it meant David wrote this?

cchrisgong: In theory the programmer responsible for the code should be tracked by git blame now, but the old signatures, I am not sure if we should keep them (they should in theory be traceable to the old version control system)

core/datainput.gms
cchrisgong: This is to reduce confusion with the code below this which overwrites the 2025-2040 values, so making this period less than 2040 or 2025 is equivalent

core/declarations.gms
cchrisgong: I see, is there a reason we cannot put goxygen in declarations.gms? it might be helpful to further explain the declared variables.. I can also include it in the definition..

main.gms
cchrisgong: forgot to mention... solscen might be deprecated, no one seems to use it in the current csv configs..

LaviniaBaumstark (approved):

Renato-Rodrigues:
before you merge this, could you add to the PR description a list of the changes that are not documentation improvements. Like for example switches or options that were deprecated.

strefler (approved):

PR #1523, opened 2024-01-17
Change price anticipation convergence criterion

robertpietzcker:
## Purpose of this PR

Change the price anticipation convergence criterion. 
Before, it only looked at the prefactor of the price anticipation, `sm_fadeoutPriceAnticip`. 
Now it checks if `p80_DevPriceAnticipGlobAllMax` (the maximum sum of the absolute of the price change due to the price anticipation effect times the imports/exports over all regions and energy carriers) is smaller than 0.1 times the goods trade imbalance threshold (`p80_surplusMaxTolerance`).
It should not have an effect on the results, but hopefully make convergence in earlier iterations possible.

## Type of change

## Checklist:

* Test runs are here: 

### DIFF
--- a/modules/80_optimization/nash/postsolve.gms
+++ b/modules/80_optimization/nash/postsolve.gms
@@ -304,16 +304,17 @@ loop(regi,
      );
 ); !!regi
 
-***additional criterion: are the anticipation terms sufficienctly small?
+*** criterion only for checking, not applied anymore: are the anticipation terms sufficienctly small?
 p80_fadeoutPriceAnticip_iter(iteration) = sm_fadeoutPriceAnticip;
 if(sm_fadeoutPriceAnticip gt cm_maxFadeOutPriceAnticip, 
-  s80_bool = 0;
+***  s80_bool = 0; !! not an active convergence criterion anymore 
   p80_messageShow("anticip") = YES;
 );
+
 *' criterion "Deviation due to price anticipation": are the resulting deviations sufficiently small?
 *' compare to 1/10th of the cutoff for goods imbalance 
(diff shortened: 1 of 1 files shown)


modules/80_optimization/nash/postsolve.gms
```diff
@@ -304,16 +304,17 @@ loop(regi,
...
 p80_fadeoutPriceAnticip_iter(iteration) = sm_fadeoutPriceAnticip;
 if(sm_fadeoutPriceAnticip gt cm_maxFadeOutPriceAnticip, 
-  s80_bool = 0;
+***  s80_bool = 0; !! not an active convergence criterion anymore 
   p80_messageShow("anticip") = YES;
 );
+
 *' criterion "Deviation due to price anticipation": are the resulting deviations sufficiently small?
 *' compare to 1/10th of the cutoff for goods imbalance 
 if(p80_DevPriceAnticipGlobAllMax("2100") gt 0.1 * p80_surplusMaxTolerance("good"),
```
LaviniaBaumstark, line 316: for the plot we need a parameter which depends on iterations adn we should use the same parameter for teh plot which is used in the criterion

mleimbach: Lavinia, there is a corresponding parameter with an iteration index: p80_DevPriceAnticipGlobAllMax2100Iter
right?
Looks fine for me.

LaviniaBaumstark: ifso, we should use `p80_DevPriceAnticipGlobAllMax2100Iter` here for the criterion

robertpietzcker: done

LaviniaBaumstark (approved):

mleimbach (approved):

PR #1522, opened 2024-01-17
only solve regions that do not have a valid solution for this nash iteration

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
If one or more regions fail to solve, not all regions are passed to
CONOPT again, potentially speeding up model runtimes (especially for 
EU21) and allowing for higher values of cm_solver_try_max, as truely
infeasible regions will return from CONOPT rather fast.

Also, `resusd` in `p80_repy_iteration` now shows total time used for all `sol_itr`, not only the last.

## Type of change
- [x] New feature 

## Checklist:

- [x] My code follows the [coding etiquette](https://github.com/remindmodel/remind/blob/develop/main.gms#L80)
- [ ] I performed a self-review of my own code
- [x] I explained my changes within the PR, particularly in hard-to-understand areas
- [x] I checked that the [in-code documentation](https://github.com/remindmodel/remind/blob/develop/maxn.gms#L120) is up-to-date
- [x] I adjusted the reporting in [`remind2`](https://github.com/pik-piam/remind2) where it was needed
- [x] I adjusted `forbiddenColumnNames` in [readCheckScenarioConfig.R](https://github.com/remindmodel/remind/blob/develop/scripts/start/readCheckScenarioConfig.R) in case the PR leads to deprecated switches
- [x] All automated model tests pass (`FAIL 0` in the output of `make test`)
- [x] The changelog `CHANGELOG.md` [has been updated correctly](https://gitlab.pik-potsdam.de/rse/rsewiki/-/wikis/Standards-for-Writing-a-Changelog)

PR #1531, opened 2024-01-30
Updating AMT scenario config

Renato-Rodrigues:
## Purpose of this PR

Update main scenario config file:

1. ~~Removed Base scenarios as we are currently calibrating to NPi assumptions.~~ please check https://github.com/remindmodel/development_issues/issues/242 
2. Fixed mistake in budget definition for EU21 scenarios: `SSP2EU-EU21-PkBudg650` was using `500` as budget, meanwhile `SSP2EU-EU21-PkBudg500`was using `650` as budget.
3. ~~Changed `slurmConfig` option on `SSP2EU-EU21-NPi-calibrate` to follow option that is used in other calibration scenarios.~~ Change postponed until runtime with different cores numbers are checked again.

## Type of change

- [ x ] Bug fix 

## Checklist:

- [ x ] My code follows the [coding etiquette](https://github.com/remindmodel/remind/blob/develop/main.gms#L80)
- [ x ] I performed a self-review of my own code


config/scenario_config.csv
```diff
@@ -2,55 +2,47 @@ title;start;CES_parameters;optimization;slurmConfig;regionmapping;extramappings_
...
 # EU21 SSP2EU;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
-SSP2EU-EU21-NPi-calibrate;calibrate;calibrate;;15;./config/regionmapping_21_EU11.csv;./config/extramapping_EU27.csv;rcp45;;;0;NPi;1;2100;;9;;;;;;;;;;;;NPi2018;regiCarbonPrice;;;on;on;EUR_regi, NEU_regi;;;;;;;;;;3;;;;;;Mix1;2005;;;;;SSP2EU-EU21-calibration: This baseline scenario follows the Shared Socioeconomic Pathways 2 called Middle of the Road.
-SSP2EU-EU21-Base;1,AMT;;;;./config/regionmapping_21_EU11.csv;./config/extramapping_EU27.csv;;;;0;;;2100;;;;off;;;;;;;;;;;;;;;;EUR_regi, NEU_regi;;;;;;;;;;;;;;;;Mix1;2005;;;;;SSP2EU-EU21-Base: This baseline calibration scenario follows the Shared Socioeconomic Pathways 2 called Middle of the Road.
+SSP2EU-EU21-NPi-calibrate;calibrate;calibrate;;14;./config/regionmapping_21_EU11.csv;./config/extramapping_EU27.csv;rcp45;;;0;NPi;1;2100;;9;;;;;;;;;;;;NPi2018;regiCarbonPrice;;;on;on;EUR_regi, NEU_regi;;;;;;;;;;3;;;;;;Mix1;2005;;;;;SSP2EU-EU21-calibration: This baseline scenario follows the Shared Socioeconomic Pathways 2 called Middle of the Road.
```
0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q, line 11: Why run `SSP2EU-EU21-NPi-calibrate` on twelve instead of 16 cores?

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q (changes requested):

config/scenario_config.csv
Renato-Rodrigues: I did a presentation in a remind meeting in 2019, when we merged REMIND-EU, that showed that increasing the number of cores from 12 to 16 had no gains for the total model solution time. 

At the time, the determinant for the time taken at each iteration was the solution time of the "most complex" regions, regions that had substantial amounts of primary energy exports or demand. The number of cores had not much influence because other regions solved much quicker in sequence as the other cores became available.

Unless someone did the work of evaluating this again and found a distinct behavior, changing to use 16 cores can be simply a waste of resources.

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q: Might be worth looking into that again with #1522.
I would say postpone this, to see if it is still true.  And if it is, change `scripts/start/choose_slurmConfig.R`, too, as it is misleading REMIND users.

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q (approved):

config/scenario_config.csv
Renato-Rodrigues: Reverted for now as suggested by the reviewer.

LaviniaBaumstark:
Yes, I agree, we decided to keep all bse scenarios

Renato-Rodrigues:
> I think we had decided to keep the Base run, so that we have a continuous tracking of how that set-up evolves over model versions.

As soon as you change the base calibration assumption, you cannot compare Baseline runs neither for results or for runtime statistics. 
I am not sure what benefit do we have from comparing apples to oranges here from my point of view.

LaviniaBaumstark:
Thanks for updating EU21-AMT-settings

LaviniaBaumstark:
> > I think we had decided to keep the Base run, so that we have a continuous tracking of how that set-up evolves over model versions.
> 
> As soon as you change the base calibration assumption, you cannot compare Baseline runs neither for results or for runtime statistics. I am not sure what benefit do we have from comparing apples to oranges here from my point of view.

I personally have no strong opinion on this but that is what we decided. Maybe just for ensuring, that base runs are in principle possible.

Renato-Rodrigues:
> Yes, I agree, we decided to keep all bse scenarios

Could you cite in more detail the use case for keeping running the Base runs or the reason to refute my comment above?

robertpietzcker:
> As soon as you change the base calibration assumption, you cannot compare Baseline runs neither for results or for runtime statistics.

I don't understand this comment - isn't it the same for Base as it is for eg PkBudg650? They both start from an NPi; the one adds policies, the other takes some away. Both can be compared over model versions, no?

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
> I don't understand this comment - isn't it the same for Base as it is for eg PkBudg650? They both start from an NPi; the one adds policies, the other takes some away. Both can be compared over model versions, no?

The question is "what is the Baseline run supposed to be?"  We have not answered that.  But during validation the question of whether Baseline is a viable counter-factual for "no climate policy at all" was met by inquisitive looks by the sector people.

> I think we had decided to keep the Base run, so that we have a continuous tracking of how that set-up evolves over model versions.

What for?  We are almost certainly never going to use these kind of runs.

We are calibrating to NPi since July.  Has _anybody_ looked at the Baseline scenarios since then?  That are five runs every week we might as well just not do, 135 so far.

Renato-Rodrigues:
> I don't understand this comment - isn't it the same for Base as it is for eg PkBudg650? They both start from an NPi; the one adds policies, the other takes some away. Both can be compared over model versions, no?

As far as understand:

- In theory, two runs of the model made under two different calibrations assumptions would never be perfectly comparable.
- This happens because during the calibration process we are fitting the indifference curves, by adjusting the CES parameters, to a different set of assumptions.
- However, the less price/cost related changes are included in the scenario, the greater is the effect of the calibration change in the optimal result found. That is because the optimal solution will rely more into moving the solution point in the indifference curve, which is highly connected to the calibration, instead of switching indifference curve levels.
-  That is why, at least in my opinion, Base runs will have much more changes dictated by the calibration change itself, than the policy runs, in which moving in the CES curve is much less important than relative price changes.

Renato-Rodrigues:
I am removing the Base change from this PR and I created an issue here https://github.com/remindmodel/development_issues/issues/242 so we can have listed the reasons to keep the Base scenarios and make a definitive decision before the next release.

robertpietzcker (approved):

PR #1549, opened 2024-02-12
Removing German specific nuclear capacity bounds

Renato-Rodrigues:
## Purpose of this PR

- A recent update to the historical data made the previous German capacity fixing located in the regipol module incompatible with the historical data.

- REMIND equation `q_PE_histCap` enforces nuclear capacity in 2010, 2015 and 2020 to be at least 90% of the historical data input calculated at `mrremind` `calcCapacity` function.
- The regipol module had extra code to enforce 2015 and 2020 German nuclear capacities to be `10.8/1000` and `7.8/1000`, respectively.

- `q_PE_histCap` sets a lower bound equal to `90% of 8.93/1000 = 8.037/1000` which is lower than the previous enforced capacity value in the regipol module `7.8/1000` for `vm_cap.fx("2020","DEU","tnrs","1")`.

- This PR removes the German specific code fixing the nuclear capacitites, and relies on the core equation to reflect the historical behavior instead.

## Type of change

- [x] Bug fix 

## Checklist:

### DIFF
--- a/modules/47_regipol/regiCarbonPrice/bounds.gms
+++ b/modules/47_regipol/regiCarbonPrice/bounds.gms
@@ -18,10 +18,6 @@
 *'
 *' ###### Bounds for Historic and Near-term Alignment
 
-*' These bounds account for historic nuclear power development.
-vm_cap.fx("2015",regi,"tnrs","1")$((cm_startyear le 2015) and (sameas(regi,"DEU"))) = 10.8/1000; 
-vm_cap.fx("2020",regi,"tnrs","1")$((cm_startyear le 2020) and (sameas(regi,"DEU"))) = 7.8/1000;
-
 *' This limits wind and solar PV capacity additions for 2025 in light of recent slow developments as of 2023.
 *' Upper bound is double the historic maximum capacity addition in 2011-2020.
 loop(regi$(sameAs(regi,"DEU")),


Renato-Rodrigues:
@fschreyer and @robertpietzcker , are you ok with this change? or would you prefer instead to update the nuclear capacity fixings and enforce an exact capacity number?
I imagine that in the context of REMIND enforcing the lower bound of 90% of the historical capacity is good enough, but maybe in the context of ARIADNE you want to hit exact numbers here. 

I am also not sure if the model would want to go even higher for nuclear capacity if we just set a lower bound here. If this is the case we would need to add an additional upper bound limit to avoid strange results.

robertpietzcker:
do you have an idea why the pe bound input has a different value? 

I think it should also be ok for ARIADNE, but if Felix disagrees we can add it again later.

robertpietzcker (approved):

Renato-Rodrigues:
The nuclear capacity values come from the `calcCapacity` function, and I think they are an average of the `Ember` capacity numbers.
Most probably the `Ember` historical capacity data was updated recently to include more recent years data. 
@fbenke-pik do you know if this was the case?

Most probably, the previous REMIND code and historical capacity values were calculated using an average of three years (2018 to 2020 maybe?), and the current one uses the average of four years instead (2018 to 2021), as far as I can guess without digging more.

fschreyer:
No objection. It is anyways better if this is done via the general formulation. 

I wonder, though, why these lines had an ``le`` and not ``l``, which was probably not intended, and maybe led to a different vintage structure as desired / as we have now. Anyways, we should also fix nuclear to 0 in 2025, but we can have a closer look once it becomes relevant in Ariadne.

fschreyer (approved):

PR #1551, opened 2024-02-13
fix bug in reportCEScalib.R

fbenke-pik:
## Purpose of this PR

Fix a small bug in reportCEScalib.R script and apply autoformat.


LaviniaBaumstark (approved):
thanks

PR #1556, opened 2024-02-16
adjust tax conv. criterion and activate it

LaviniaBaumstark:
## Purpose of this PR
adjust and activate tax criterion as Nash convergance criterion

## Type of change

## Checklist:

* Test runs are here: 


mleimbach:
Did we decided for 10% tolerance? That`s certainly too much. If this should be a real check, 1% or even less should be applied.

mleimbach (approved):

PR #1565, opened 2024-02-20
fix on reference run: landuse in REMIND standalone runs + MAGICC6

orichters:
## Purpose of this PR

- Hopefully fixes https://github.com/remindmodel/development_issues/issues/221#issuecomment-1953940386
- for `Costs|Biomass|Price integral presolve`: For `cm_startyear` > 2005, load the value from reference run and overwrite only for `t`, not `ttot`. As this parameter is used only in [this single equation](https://github.com/remindmodel/remind/blob/ef8a82fb4a0d92cbd7a324f82f400c3f78ae5c0f/modules/30_biomass/magpie_40/equations.gms#L28) for ttot > cm_startyear, this should have no effect on REMIND, but fixes the reporting [here in reportCosts](https://github.com/pik-piam/remind2/blob/f977fd850d796e2d3a1cb485f2b263ec9011c59e/R/reportCosts.R#L88).
- for `Net Taxes|GHG emissions|w/o CO2 LUC` and `Net Taxes|GHG emissions|CO2 LUC`: The problem was that [this equation](https://github.com/remindmodel/remind/blob/79d4579de65f73a86a626b7d58052f0da41265c0/modules/21_tax/on/postsolve.gms#L26) returned different results if `cm_multigasscen` differed between scenarios. Load `pm_taxrevCO2LUC0` from `input_ref.gdx` and overwrite it only for `t`, not `ttot`
- for `Net Taxes|SO2`. The effect was very small, caused by the fact that `pm_gdp_gdx` can slightly change between iterations. In the reference run, the value for `p21_tau_so2_tax` was not recalculated in postsolve, but in the next runs in the cascade it was indeed calculated based on the adjusted value. As this was a rough approximation anyhow, use `pm_gdp` which does not change. @strefler is fine with that.
- simply fix MAGICC6 variables in post-processing. We will not (be able to) fix that at the source, and the error is on the order of 0.1%, so I think that is the easiest (say: only) option. @gabriel-abrahao: maybe you can have a short look?

## Type of change

- [x] Bug fix 

## Checklist:

- [x] All automated model tests pass (`FAIL 0` in the output of `make test`): `[ FAIL 0 | WARN 0 | SKIP 6 | PASS 81 ]`

* Test run that fixes the errors as intended, see also [this development issue](https://github.com/remindmodel/development_issues/issues/220#issuecomment-1957152387): `/p/tmp/oliverr/remind-smallfix/output/SSP2EU-PkBudg1050_2024-02-21_09.47.34`


dklein-pik (approved):
The solution for the correct fixing (overwrite only for `t` and not `ttot`) look good. Thank you!

PR #1567, opened 2024-02-20
increase memory on single node runs

orichters:
## Purpose of this PR

- or reporting.R fails on oom event while running reportFE.

## Type of change

- [x] Bug fix 

## Checklist:

- [x] All automated model tests pass (`FAIL 0` in the output of `make test`, including when I [do not skip the output tests](https://github.com/remindmodel/remind/blob/679f292c69a4e3e660ef7100b52e1d087e198700/tests/testthat/test_06-output.R#L7)): `[ FAIL 0 | WARN 0 | SKIP 5 | PASS 92 ]`


orichters:
Probably rather a more substantive problem, see mattermost decision.

orichters:
The problem does not seem simple to solve, see https://github.com/pik-piam/remind2/issues/549

To avoid having a broken model, merge that, undo it later

Issue #1561, opened 2024-02-19
negative fossil emissions


0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
As of `/p/projects/remind/modeltests/remind/output/SSP2EU-NPi-AMT_2024-01-05_22.08.21/`, REMIND reports negative fossil emissions in the industry sector, e.g.
```
$ awk -F ';' '1 == NR || $0 ~ /CAZ;Emi\|CO2\|Energy\|Demand\|Industry\|Chemicals\|Solids\|\+\|Fossil/' /p/projects/remind/modeltests/remind/output/SSP2EU-NPi-AMT_2024-01-05_22.08.21/REMIND_generic_SSP2EU-NPi-AMT.mif | column -ts ';' | cut -c -145
Model   Scenario        Region  Variable                                                  Unit       2005       2010       2015       2020       
REMIND  SSP2EU-NPi-AMT  CAZ     Emi|CO2|Energy|Demand|Industry|Chemicals|Solids|+|Fossil  Mt CO2/yr  1.2903585  4.2679138  4.3356236  -3.3315255 
```
see also https://github.com/pik-piam/remind2/issues/520

This error tracks through `remind2::reportEmi()` via `EmiIndSubSec < 0` ← `o37_deFeIndSubSec_woNonEn < 0` ← `o37_demFeIndSubSec < vm_demFeNonEnergySector` to https://github.com/remindmodel/remind/blob/3341f41e94961fb22a748006d90a86caa6be979d/modules/37_industry/subsectors/equations.gms#L215-L231 `vm_demFeNonEnergySector` taking `p37_chemicals_feedstock_share` (here 70 %) of `vm_cesIO(feso_chemicals)`, although 68 % of `fesos` (across all industry subsectors) is provided by `sesobio`:

variable/parameter | value
---|---
`vm_cesIO(feso_chemicals)` | 3.05e-3
`vm_demFeNonEnergySector(sesofos,fesos,indst,ETS)` | 2.14e-3
`vm_demFeNonEnergySector(sesobio,fesos,indst,ETS)` | 0
`vm_demFeSector_afterTax(sesofos,fesos,indst,ETS)` | 3.37e-3
`vm_demFeSector_afterTax(sesobio,fesos,indst,ETS)` | 7.26e-3

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
[...]

PR #1576, opened 2024-02-23
fix negative fossil emissions in industry

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
This does ~two things~:
1. ~Subtract `vm_demFENonEnergySector` for `entySeBio` and `entySeSyn` from `vm_emiTeDetailMkt` in `q_emiTeDetailMkt`, as they will generate negative emissions when plastics are landfilled.~  (Already accounted for elsewhere.)
2. Enforce identical shares between `entySeFos`, `entySeBio` and `entySeSyn` in `vm_demFENonEnergySector` and `vm_demFEsector_afterTax`.

- [x] Bug fix 

## Checklist:

### DIFF
--- a/modules/37_industry/subsectors/equations.gms
+++ b/modules/37_industry/subsectors/equations.gms
@@ -242,6 +242,21 @@ q37_feedstocksLimit(ttot,regi,entySE,entyFE,emiMkt)$(
   vm_demFENonEnergySector(ttot,regi,entySE,entyFE,"indst",emiMkt)
 ;
 
+*' Feedstocks have identical fossil/biomass/synfuel shares as industry FE
+q37_feedstocksShares(t,regi,entySE,entyFE,emiMkt)$(
+                                          sum(te, se2fe(entySE,entyFE,te)) ) ..
+    vm_demFEsector_afterTax(t,regi,entySE,entyFE,"indst",emiMkt)
+  * sum(se2fe(entySE2,entyFE,te),
+      vm_demFENonEnergySector(t,regi,entySE2,entyFE,"indst",emiMkt)
+    )
+  =e=
+    vm_demFENonEnergySector(t,regi,entySE,entyFE,"indst",emiMkt)
+  * sum(se2fe2(entySE2,entyFE,te),
+      vm_demFEsector_afterTax(t,regi,entySE2,entyFE,"indst",emiMkt)
+    )
+;
+
+
 *' Calculate mass of carbon contained in chemical feedstocks
 q37_FeedstocksCarbon(ttot,regi,sefe(entySE,entyFE),emiMkt)$(
                          entyFE2sector2emiMkt_NonEn(entyFE,"indst",emiMkt) ) ..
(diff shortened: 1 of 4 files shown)


0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
Test runs: NPi/PkBudg1050/PkBudg500, latest AMT and this fix.  Chemicals is not a net-negative subsector with magical negative fossil emissions anymore.
image
image
`/p/tmp/pehl/Remind/compScen-2024-02-28_08.37.02-H12.pdf`

modules/37_industry/subsectors/equations.gms
```diff
@@ -242,6 +242,21 @@ q37_feedstocksLimit(ttot,regi,entySE,entyFE,emiMkt)$(
...
+    vm_demFEsector_afterTax(t,regi,entySE,entyFE,"indst",emiMkt)
+  * sum(se2fe(entySE2,entyFE,te),
+      vm_demFENonEnergySector(t,regi,entySE2,entyFE,"indst",emiMkt)
+    )
+  =e=
+    vm_demFENonEnergySector(t,regi,entySE,entyFE,"indst",emiMkt)
+  * sum(se2fe2(entySE2,entyFE,te),
+      vm_demFEsector_afterTax(t,regi,entySE2,entyFE,"indst",emiMkt)
+    )
+;
```
0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q, line 257: @Renato-Rodrigues what is your take on this formulation, w.r.t. CONOPT solvability?

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
close #1561

modules/37_industry/subsectors/equations.gms
Renato-Rodrigues: > @Renato-Rodrigues what is your take on this formulation, w.r.t. CONOPT solvability?

I would say it is not ideal, as it enforces strong circularities in two levels: 
(1) the circularity between an specific FE carrier in industry (`vm_demFEsector(entyFE)`) and a sub-product of it (`vm_demFENonEnergySector(entyFE)`), as the endogenous share calculated at `q37_feedstocksShares` will influence the value of the first variable marginals.
(2) the dependency of an specific FE carrier in industry (`vm_demFEsector(entyFE)`) and the total FE industry. Changing any FE carrier marginal, changes all other FE carrier marginal values.

This is taken into account in the Jacobian matrix, but it demands at least two additional second order terms that could be quite complicated depending of the case.

I am not sure although much complex this is to the model to solve without testing it.

Ideally, feedstocks demand should be treated independently of the total industry final energy demand to avoid this. I imagine this could remove the need for this circularity, but this would require reworking more equations, so if the solution time is acceptable with this solution, I guess we would have to live with that for now.

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
@mellamoSimon, @fschreyer bump

mellamoSimon:
> if the solution time is acceptable with this solution, I guess we would have to live with that for now.

did you check if there's an impact on the solution time? Otherwise, it looks good to me :)

mellamoSimon:
>So we are interested in correct results only if they do not take too much time?

I think that, to a certain extent, that is true... but I assume that Renato's concern is that it doesn't blow our ability to use the model. How long did the runs take?

mellamoSimon (approved):

PR #1583, opened 2024-02-28
fix reference scenario gdx to use the NPi run

Renato-Rodrigues:
## Purpose of this PR

- fix reference scenario gdx to use the NPi run so all scenarios use the same 2025/2030 "comparison points", avoiding different adjustment costs in different policy scenarios. 

## Type of change

- [ x ] Bug fix


robertpietzcker (approved):
Thanks!

PR #1588, opened 2024-02-29
some cleaning of default.cfg

LaviniaBaumstark:
## Purpose of this PR
cleaning

## Type of change

- [x] Refactoring

## Checklist:

* Test runs are here: 


fbenke-pik:
I would not run the nash analysis script in the current format after each run.  The script is large and inefficient and usually requires more than the standard allocated memory to succeed.

In this format, it is suitable for exploration as part of the TF, but not a general-purpose script. If you want to run it always, it would have to be trimmed and improved.

fbenke-pik (approved):

PR #1592, opened 2024-03-01
preparation of ccsinje cost increase

tabeado:
## Purpose of this PR
This PR prepares the increase of cost for ccsinje. 
- Introduce **construction time** (from 0 to 3 years) which leads to regionally slightly differentiated cost due to different financing cost markups (through p_tkpremused)
- Change **tech_stat** from 2 to 0. There are many counteracting factors influencing the technical cost of ccsinje over time. Our best guess is to leave cost constant.
- Switch **cm_ccsinjeCost** to choose between low, medium, and high cost:

_low_ =  old estimate before the change; i.e. main estimate as of 2035 is ~7.5 USD/tCO2. This also changes tech_stat back to 2 and (for now) constrTme back to 0. It is kept as default for now until changes are discussed at the next remind meeting
_med_ = new main estimate, stays constant at 12 USD/tCO2 over time. As a crude reference, taking the average of all on & offshore DOG fields & saline formations in Budinis et al (2017), yields ~11.4 USD/tCO2. 
_high_ = upper end of storage cost + somewhat higher transport cost. 

Effects from the higher cost on default runs will be discussed at the next REMIND meeting & then following PR that changes the default cost, i.e. adjusts switch default from "low" to "med"

## Type of change

- [ ] Bug fix 
- [ ] Refactoring
- [ ] New feature 
- [ ] Fundamental change

## Checklist:


core/input/generisdata_tech.prn
```diff
@@ -121,12 +121,13 @@ lifetime                30       30          45          45          45
 *lifetime              40          40          40          40
 
 +                 ccsinje
-tech_stat               2
-inco0                 220
```
LaviniaBaumstark, line 125: do you need to also adjsut it in generisdata_tech_SSP1.prn and generisdata_tech_SSP5.prn?

main.gms
```diff
@@ -1474,6 +1474,8 @@ $setglobal cm_inco0RegiFactor  off  !! def = off
 *** cm_CCS_markup "multiplicative factor for CSS cost markup"
 ***   def <- "off" = use default CCS pm_inco0_t values.
 ***   or number (ex. 0.66), multiply by 0.66 the CSS cost markup
+$setglobal cm_ccsinjeCost low !! def = low !! regexp = med|low|high
```
LaviniaBaumstark, line 1477: why are you using a compiler flag instead of a parameter?

LaviniaBaumstark (commented):
please specify the type of your change

core/input/generisdata_tech.prn
tabeado: yes, you are right, I forgot that!

main.gms
tabeado: The idea was to have a name as we wanted to give a low, medium & high estimate.
Also, for the low estimate, which is the original one, it is necessary to change more than just inco0. But I guess I could also accomodate that with a parameter, and just suggest a value for the high estimate in main.gms, if you prefer that

LaviniaBaumstark: ah, o.k. I am fine with a compiler flag if there are reasons for that

tabeado:
The code compiles, however, when running make test the testOneRegi run fails. I updated renv to rule that out.  /p/tmp/tabeado/remind_clean_new/remind/output/testOneRegi
This is the emerging issue:
image
@LaviniaBaumstark  I cannot see the connection to my changes, maybe you can?

Tagging @mellamoSimon as the error that comes up is connected to your feedstock implementation

mellamoSimon:
oh no... there's a fix that [might be complicated](https://github.com/remindmodel/remind/pull/1576/files#r1505776572) for the solver. @Renato-Rodrigues will probably be able to confirm if the error you are getting could be explained by the way that equation is defined. Did you get successful runs though? I mean not for testOneREgi.. maybe the `gdx` file for tesOneRegi needs to be updated and that might help?

Renato-Rodrigues:
@mellamoSimon and @0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q you most probably would only have to limit the equation `q37_feedstocksShares` to work only in the `sameas(emiMkt,"ETS")` like you probably do for other feedstocks FE equations.
This will most probably remove the above reported issue.

mellamoSimon:
thank you! I can test that tomorrow

PR #1593, opened 2024-03-04
New ccs near-term bounds based on project announcements

amerfort:
## Purpose of this PR
I updated the near-term projections of geologic CO2 storage according to the IEA CCUS project database. 
Therefore, I newly introduce a regional lower bound for 2025 and 2030 based on storage projects that are either operational or under construction. I update the upper bound for both years including also announced projects (often times the final investment decision is still pending). This assumption follows the reasoning that geologic storage requires extensive geologic exploration and lead times for CCS projects are at least 5 years, but likely even longer (expert judgement, personal communication, CDRterra). 

## Type of change
Update near-term CCS bounds

- [ ] Bug fix 
- [ ] Refactoring
- [ ] New feature 
- [ ] Fundamental change

## Checklist:

* Test runs are here: 

### DIFF
--- a/config/default.cfg
+++ b/config/default.cfg
@@ -27,7 +27,7 @@ cfg$regionmapping <- "config/regionmappingH12.csv"
 ### Additional (optional) region mapping, so that those validation data can be loaded that contain the corresponding additional regions.
 cfg$extramappings_historic <- ""
 #### Current input data revision (<mainrevision>.<subrevision>) ####
-cfg$inputRevision <- "6.70"
+cfg$inputRevision <- "6.74"
 
 #### Current CES parameter and GDX revision (commit hash) ####
 cfg$CESandGDXversion <- "cdddb54b54a8586b4fef00eb60a3be6cfa23ca55"
--- a/core/bounds.gms
+++ b/core/bounds.gms
@@ -391,20 +391,28 @@ $endif
 
 
 *** -------------------------------------------------------------------------------------------------------------
-*RP* Upper limit on CCS deployment in 2020-2030
+*AM* No geological storage before 2025 (omitting approx. 40 MtCCS/yr globally in 2020 for Enhanced Oil Recovery)
+*AM* Lower limit for 2025 and 2030 is capacities of all projects that are operational and under construction from project data base
+*AM* Upper limit for 2025 and 2030 additionally includes announced/planned projects from project data base
+*AM* In nash-mode regions cannot easily share ressources, therefore CCS potentials are redistributed in Europe: 
+*AM* Potential of EU27 regions is pooled and redistributed according to GDP (Only upper limit for 2030)
+*AM* Norway and UK announced to store CO2 for EU27 countries. So 50% of Norway and UK potential in 2030 is attributed to EU27-Pool
 *LP* if c_ccsinjecratescen=0 --> no CCS at all and vm_co2CCS is fixed to 0 before, therefore the upper bound is only set if there should be CCS!
 *** -------------------------------------------------------------------------------------------------------------
 
 if ( c_ccsinjecratescen gt 0,
-        vm_co2CCS.up(ttot,regi,"cco2","ico2","ccsinje","1")$(ttot.val ge 2005 AND ttot.val lt 2020) = 0;
-	vm_co2CCS.up("2020",regi,"cco2","ico2","ccsinje","1") = pm_boundCapCCS(regi);
-	vm_co2CCS.up("2025",regi,"cco2","ico2","ccsinje","1") = pm_boundCapCCS(regi);
+  vm_co2CCS.up(ttot,regi,"cco2","ico2","ccsinje","1")$(ttot.val ge 2005 AND ttot.val lt 2025) = 0;
+	vm_co2CCS.up("2025",regi,"cco2","ico2","ccsinje","1") = pm_boundCapCCS("2025",regi,"up")/(1000*11/3);
+	vm_co2CCS.up("2030",regi,"cco2","ico2","ccsinje","1") = pm_boundCapCCS("2030",regi,"up")/(1000*11/3);
+	vm_co2CCS.lo("2025",regi,"cco2","ico2","ccsinje","1") = pm_boundCapCCS("2025",regi,"low")/(1000*11/3);
+	vm_co2CCS.lo("2030",regi,"cco2","ico2","ccsinje","1") = pm_boundCapCCS("2030",regi,"low")/(1000*11/3);
 );
 
 loop(regi,
-  if( (pm_boundCapCCS(regi) eq 0),
-    vm_cap.fx("2020",regi,teCCS,rlf) = 0;
-	vm_cap.fx("2025",regi,teCCS,rlf) = 0;
+  loop(t$(t.val le 2030),
+    if( ( pm_boundCapCCS(t,regi,"up") eq 0),
+      vm_cap.fx(t,regi,teCCS,rlf) = 0;
+    );
   );
 );
 
--- a/core/datainput.gms
+++ b/core/datainput.gms
@@ -823,8 +823,8 @@ if(pm_NuclearConstraint("2020",regi,"tnrs")<0,
 );
 );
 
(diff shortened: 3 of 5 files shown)


amerfort:
Dear all, is that an acceptable change? Is it too restrictive to set lower and upper bounds on geologic storage (sometimes leading to fixed values, when upper and lower bounds are identical?). 
And is it ok to use a single cs4r file to set a lower and an upper bound? And as we are in the validation rush, can we use it as it is for the validation and improve coding etiquette afterwards?

Renato-Rodrigues:
> Is it too restrictive to set lower and upper bounds on geologic storage (sometimes leading to fixed values, when upper and lower bounds are identical?).

I consider fixings for values in future periods of time in REMIND as a not good practice.
It is very debatable from a pure theoretical point of view, as it assumes we know the future, but also it leaves no room for the model to deviate from your fixings. The later always has as a bad side effect of hiding the information if the model wants to get ride or wants to get more of the fixed quantity that you are bounding. Fixings also remove marginal information.
 
Allowing a deviation range of at least 5% to 10% in between the upper and lower bounds is a better approach in my opinion.

Also, if you want to fix something, avoid setting a upper and a lower bound to the same value in different statements, and use the .fx statement instead. The code is much clearer for what you are doing this way.

Renato-Rodrigues:
> And is it ok to use a single cs4r file to set a lower and an upper bound? And as we are in the validation rush, can we use it as it is for the validation and improve coding etiquette afterwards?

I would say it is perfectly fine to do that.
Just as suggestions for future improvements I would include to the input data (`pm_boundCapCCS`):
(1) an additional dimension to designate if the value is an upper bound or an lower bound. In case you chose to just define a median value instead in the file and set an allowed deviation in code as I suggested in my previous message, this would not be necessary.
(2) and additional dimension to reflect the data with a period (year) dimension, so we can extend current bounds to cover different values for 2020, 2025, and 2030 in order to reflect more correctly cases like historical oil recovery CDR in USA that are not depreciable and could be fostered by current legislations.

On another note, I believe this code should be moved from the core files to the CDR module files.

Renato-Rodrigues:
Thanks Anne.
Please disregard my second comment then.

I had no access to the new input file, so I made the assumption that you did not have the period dimension because you use multiple lines to define the upper and lower bounds for 2025 and 2030. 
In the case you have the years in the input file, I guess you can rewrite lines 405-408 to use the period information from the input file instead of repeating multiple lines for each year that you have data from. This way you would have a generalized solution that could also work if new data is added without changes to the code.

On the code locations, my bad, it is surely not the same as CDR!
Maybe we should think in a way to reorganize CCS specific code also, so it can belong to an specific place instead of pilling up in the core generic code files. This is not related with your PR although.

robertpietzcker:
Thanks for the work, Anne!
1. it is currently still a draft PR - is this on purpose?
2. could you maybe put a link to the new input file in the PR? then it is easier for a reviewer to understand what this means. 
3. Potentially some of the code can be made nicer, but it doesn't look wrong to me, so I will accept it now in order to not slow down the validation process. 
3a. I would define and use a conversion parameter (is this MtCO2 to GtC?) instead of (1000*11/3);
4. could you please make a compScen of an NPi and a PeakBudg650 with and without your changes and link them in the PR? (I think you can still do that after the PR has been accepted and merged, if I am not mistaken. Otherwise, simply as a comment down here)

robertpietzcker:
(ah, I just accidently changed it from "draft PR" to "PR" myself...

robertpietzcker:
wait - I just realized that you increased the input data revision - did anything else change in this input data revision? so do we need to calibrate REMIND for this?

robertpietzcker (approved):
if the input data increase doesn't bring any other changes to the input data and thus requires calibration, I think it can be merged.

LaviniaBaumstark:
this input data change should not need new clibration runs. Anyway, for the final validation runs I will run a new calibration.

Renato-Rodrigues:
Were the points I raised in my first message covered in any way that is not evident to me before this PR was merged?
https://github.com/remindmodel/remind/pull/1593#issuecomment-1977038553

If yes, please disregard this message. In this case I would just add a recommendation to include to the PR before you merge it an explanation on how the raised issue was fixed without changing the files included in the PR. 

If not, I would like to ask whenever any reviewer brings points to be addressed in any PR, that the PR should not be merged before either:
- the reviewer retracts his requests, or
- the author contacts the reviewer and explains why the reasons to not add the requested change to the PR. 

This was a topic already discussed extensively in a REMIND meeting, and this is a fundamental principle in software development and peer reviewed processes.

PS: Anne, please don't get this message wrong. 
**You did an amazing job both on the content side, and in your prompt and relevant replies in this PR.** 
I just consider that the point I raised about adding a slack instead for the fixing of exact values to future variables is relevant enough that should be fixed before the merge. Doing this should not imply more than 5 minutes of additional work also, so I see no reason why any urgency reason would matter to ignore it. 
As far as I understood, your direct reply was positive to my request. 
So, unless I am missing something, I would like my review to be considered before merging the PR.

robertpietzcker:
Hi Renato, 

I guess my approving the PR without reacting to your comments, and saying "I guess it can be merged" gave the clear signal to Anne "go ahead and merge". 
I should have reacted more extensively to your first comment before approving the PR - I think I (mis)read them more as "we should discuss this" rather than "they are clear roadblocks to merging". 

If I understand correctly, the three points you raise are a) "should we have hard bounds for the future in REMIND?", b) "it would be good to add a 10% range left free", and c) "don't use lower/upper bounds when you actually fix a value", right?

on a), I would say we had such hard bounds in the past, and while in principal I think our adjustment costs approach is much better in creating realistic deployment pathways, it is very challenging to really represent the early upscaling phase, where different countries may have strongly differing initiative, and it will be impossible to represent this with general adjustment costs. Given the argument "you can't do CCS in <7 years from concept to operating" to fix 2030 upper/lower bounds thus seems reasonable in this case. 

b) I agree that it would be good to leave the model some range to make decisions, as it will allow us to see if the model is on the upper or lower bound - this would speak for a 5-10% increase of the upper bound. At the same time, we have many regions where the bound is at 0, so adding 5-10% won't make a difference. Still, this doesn't speak against multiplying the upper bound by 1.1 to give the model some slack for all regions where it is >0.

c) I think your comment is correct when describing pure fixings; however here a fixing is only a byproduct of the fact that some regions don't have any new CCS plans, so the upper and lower bound fall together on the same numerical value. 
I am not sure if for that case there is really an advantage in first setting the upper and lower bounds for all regions (because for most upper bound > lower bound), and then asking if upper bound = lower bound and then setting a `.fx` after that? 



PS: how do you retract requests / how can one see requests by different reviewers? 
Or did you mean "that I didn't write 'please go ahead with a merge'?

Renato-Rodrigues:
>I guess my approving the PR without reacting to your comments, and saying "I guess it can be merged" gave the clear signal to Anne "go ahead and merge".

I understand that Anne focused more on your review orientations, but I think that this does not retract from the fact that there were other reviewers involved in the PR review.

> I should have reacted more extensively to your first comment before approving the PR - I think I (mis)read them more as "we should discuss this" rather than "they are clear roadblocks to merging".

I don't think a reviewer cannot approve the PR as you did if you agree with the current state of the code. You are not obliged to react to other reviewer's comments.
 
However, I strongly believe that it is the responsibility of the PR creator to only merge the PR if all issues raised in the PR by all reviewers are either addressed or include a justification why they will not be addressed.
It is the author responsibility and prerogative to reply to the issues raised. It is not a responsibility of other reviewers to react to the points raised other than their owns.

About your points, I don't disagree with anything that you wrote on (a). 
On (b) I would just like to additionally underline that allowing the slack removes rigidity from the model, possibly avoiding infeasibilities and/or adding debugging information to future model runs.
(c) is a lesser issue, but it is also an immediate fix. There are infinite ways of writing the same code, but that does not mean that we should not abide by the simplicity principle in software engineering.

> PS: how do you retract requests / how can one see requests by different reviewers?
Or did you mean "that I didn't write 'please go ahead with a merge'?

You should be able anyway to see any public comment made in the PR by any other reviewer. 
Again, I don't think it is a reviewer job to reply to other reviewers. 
I was always referring to the fact that the PR author should provide feedback or justification about the issues raised, and not necessarily the other reviewers.

PR #1604, opened 2024-03-11
harden nash against spurious consecutive failures

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
- `p80_trackConsecFail` did check solvestat/modelstat only against 1/2 and 4/7
- if some regions where solved in a sol_itr CONOPT iteration other then one, solvestat/modelstat for other regions would be 0/0 in `p80_repy`, and `p80_trackConsecFail` would be erroneously increased

- checking `p80_repy_iteration` instead

- [X] Bug fix 

## Checklist:

### DIFF
--- a/modules/80_optimization/nash/postsolve.gms
+++ b/modules/80_optimization/nash/postsolve.gms
@@ -640,29 +640,28 @@ if(s80_bool eq 1,
 
 );
 
-
-*** check if any region has failed to solve consecutively for a certain number of times
-if(cm_abortOnConsecFail, !! execute only if consecutive failures switch is non-zero
-    loop(regi,
-        if(((p80_repy(regi,"solvestat") eq 1) and (p80_repy(regi,"modelstat") eq 2))
-        or ((p80_repy(regi,"solvestat") eq 4) and (p80_repy(regi,"modelstat") eq 7)), !! region was solved successfully
-            p80_trackConsecFail(regi) = 0;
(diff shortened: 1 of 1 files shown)


robertpietzcker:
I actually notice the same thing last week and changed the code a bit more fundamentally, also adding some output of time used during the various loops (https://github.com/robertpietzcker/remind/tree/202403_improveNashOutput).
If it is ok with you, I would rather prefer NOT to merge your version but rather go the full step directly (I will prepare the PR in an hour or so, I wanted to test it over the weekend before making the PR) so I don't have to change my code to include your changes  (which are part of my changes, but in a slightly different way)?

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
I would opt for separating bug fixes (this) and new features (printing timestamps).

robertpietzcker:
ok, no problem, I can take split my PR and take the "printing time steps" out of the first PR; 
it is just that I added a split of p80_repy into "what happened in the current loop of solitr" and "what happened until now in this solitr loop", which both solves the wrong code (that you saw and also fixed) and allows more precise information on what is happening in the model.

robertpietzcker:
sorry, alles zu voll. 
mach erstmal so, und ich pack dann meine Änderungen danach dazu

robertpietzcker (approved):

PR #1624, opened 2024-03-19
set LFITER to the highest permissible number so CONOPT stops fussing about it

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:


orichters (approved):
Assuming that fixes
```
 ** Warning **  The value of LFITER is out of range.
                LFITER is decreased from 2147483647 to 1000000000.
```

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
Yes.  But it does not.  Stupid CONOPT.

PR #1626, opened 2024-03-19
improving comment for cm_implicitQttyTarget convergence

Renato-Rodrigues:
## Purpose of this PR

- Improving comment for non converged quantity targets.

## Type of change


orichters (approved):
Thanks!

PR #1634, opened 2024-03-25
fix cm_repeatNonOpt regexp

JakobBD:
Also change non-'off' value from 'yes' to more intuitive 'on' (has no effect, since is only compared to 'off' string in code)

## Purpose of this PR

bugfix: The lack of regexp throws an error when running REMIND with `scenario_config.csv`, as it misses the `cm_repeatNonOpt` switch in the config file. 

## Type of change

- [x] Bug fix 

## Checklist:

* Test runs are here: 

### DIFF
--- a/main.gms
+++ b/main.gms
@@ -216,9 +216,9 @@ $offdigit
 *** turn profiling off (0) or on (1-3, different levels of detail)
 option profile = 0;
 
-file foo_msg;     !! This creates a dummy output file with a well-defined output format:  
+file foo_msg;     !! This creates a dummy output file with a well-defined output format:
 foo_msg.nr = 1;   !! namely F-format (decimal) (and not E-format = scientific notation)
-*** The file can throughout the code be activated with `putclose foo_msg;` and used in the form `put_utility foo_msg "msg" / "xxxx"` to print out xxxx to full.lst 
+*** The file can throughout the code be activated with `putclose foo_msg;` and used in the form `put_utility foo_msg "msg" / "xxxx"` to print out xxxx to full.lst
 *** and be sure that the numeric format is F-format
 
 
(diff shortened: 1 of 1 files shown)


0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q (approved):

orichters:
Did REMIND really crash for you without the `regexp`? This should really not be the case! Can you provide an example where that happens, @JakobBD?

JakobBD:
`/p/tmp/jakobdu/remind_pbs/output/SSP2EU-NPi_2024-03-20_17.17.02/log.txt`
I changed the code in the repo since then, but the log file lists the commit hash, if you'd like to reproduce the error.

orichters:
Honestly, that is a complete mystery for me. Are you sure you did not adjust the `main.gms` somewhat between the creation of the rdata file and the moment the model run was prepared? Because if I check out the commit hash and run `gms::readDefaultConfig(".")$gms$cm_repeatNonOpt` returns `"off"` as it should, while if you load `config.Rdata`, this switch is indeed missing in `cfg$gms`. But basically, the `cfg` is read twice using the same code as far as I see, and I really wonder how this might have happened that the switch is present only once without any change in `main.gms`.

Might it be that between you send the run to slurm and the Rdata file was written, and the run actually started, you pulled from develop, caught [this commit](https://github.com/JakobBD/remind/commit/56d1adebb2e322e2c07aa4b8591cff49259005bd) and therefore the switch was missing in your cfg? Just a guess, but I think that is more probable than the code being wrong that works for everyone else.

Strong support for this theory: `SSP2EU-NPi_2024-03-20_17.17.02` states the run was started at 17.17, while your latest commit according to the logfile is `2024-03-20 17:22:55`. So I think that is no mystery anymore :)

JakobBD:
Ah, interesting. Then I just got misled by the `Either adapt the parameter choice or the regexp in main.gms` error message.

PR #1649, opened 2024-04-02
adjust checkFixCfg error message

orichters:
## Purpose of this PR

- remove confusing text, see #1634 
- I think the following error message that will be printed in case of regexp mismatches anyway is sufficient:
```
Parameter cfg$gms$whatever = off does not fit this regular expression: yes|no
```

## Type of change

- [x] just some error message change


JakobBD (approved):

JakobBD:
Is the regexp always given in the main.gms? Then the message `does not fit this regular expression` could be extended by `in main.gms` to indicate where to look.

PR #1475, opened 2023-11-23
AR6 Climate assessment on reporting

gabriel-abrahao:
## Purpose of this PR
This PR adds the option to run AR6's [climate-assessment](https://github.com/iiasa/climate-assessment) as part of the normal `output.R` procedure, under the option `ar6Climate`. If one wants it to be run automatically, it can be added to `cfg$output`. This includes the probabilistic assessment of temperature change, compatible with the AR6 standards. 

By default, the whole 600 member MAGICC v7.5.3 ensemble is run, which can take a while. Only certain percentiles are output, with each percentile in a different variable. Although the original output is yearly, which can be found inside a subfolder `climate-temp`, only data for the REMIND timesteps are added to the standard MIF (`REMIND_generic***`). All new variables have the `AR6 climate diagnostics|` prefix.  The original temperature variable, `Temperature|Global Mean`, is still calculated with the old (v6) MAGICC. The comparable variable in the new assessment is  `AR6 climate diagnostics|Surface Temperature (GSAT)|MAGICCv7.5.3|XXth percentile`, with separate variables for 10 different percentiles.

The subfolder `climate-temp` includes the full output of `climate-assessment`, including files describing temperature exceedence probabilities and the consequent climate classification of the scenario (C1, C2, etc). 

## Type of change
- [x] New feature 

## Checklist:

- [x] My code follows the [coding etiquette](https://github.com/remindmodel/remind/blob/develop/main.gms#L80)
- [x] I performed a self-review of my own code
- [x] I explained my changes within the PR, particularly in hard-to-understand areas
- [x] I checked that the [in-code documentation](https://github.com/remindmodel/remind/blob/develop/main.gms#L120) is up-to-date
- [x] I adjusted the reporting in [`remind2`](https://github.com/pik-piam/remind2) where it was needed
- [x] I adjusted `forbiddenColumnNames` in [readCheckScenarioConfig.R](https://github.com/remindmodel/remind/blob/develop/scripts/start/readCheckScenarioConfig.R) in case the PR leads to deprecated switches
- [ ] All automated model tests pass (`FAIL 0` in the output of `make test`)
- [x] The changelog `CHANGELOG.md` [has been updated correctly](https://gitlab.pik-potsdam.de/rse/rsewiki/-/wikis/Standards-for-Writing-a-Changelog)

PR #1653, opened 2024-04-04
Integrated MAGICC7/AR6 climate assessments between REMIND iterations

tonnrueter:
## Purpose of this PR

The `15_climate` REMIND module now offers the `ar6` option. This runs a single scenario MAGICC7 climate emulation via IIASA's `climate-assessment` to obtain the global mean temperature based on REMIND generated emission data. Previously we used the same simple climate model integration for [AR6 Climate assessment on reporting](https://github.com/remindmodel/remind/pull/1475)

## Type of change

- [ ] Bug fix 
- [ ] Refactoring
- [x] New feature 
- [ ] Fundamental change

## Checklist:

* Test runs are here: 


core/loop.gms
```diff
@@ -149,6 +149,16 @@ o_negitr_disc_cons_drInt_reg(iteration,regi) =
     + sum(ttot$(ttot.val eq 2100), vm_cons.l(ttot,regi) * qm_budget.m(ttot,regi)/ (qm_budget.m("2005",regi) + 1.e-8) * ( pm_ttot_val(ttot)- pm_ttot_val(ttot-1) ) * 0.5 )
 ;
 
+***--------------------------------------------------------------------------
+***         Save GDX with solution before postsolve
+***--------------------------------------------------------------------------
+execute_unload 'fulldata_prepostsolve';
```
orichters, line 155: prepostsolve is a bit like Ostwestfalen.

modules/15_climate/ar6/declarations.gms
```diff
@@ -0,0 +1,47 @@
+*** |  (C) 2006-2023 Potsdam Institute for Climate Impact Research (PIK)
```
orichters, line 1: I would actually be in favor of having MAGICC7 somewhere in the module name, because the AR6 used different climate models. Why not MAGICC7_AR6?

orichters (commented):
I just had a very short look, that is way to much to do that for a moment.

modules/15_climate/ar6/declarations.gms
tonnrueter: That's fine by me! The other `15_climate` module option is `magicc`, so `magicc7_ar6`  would match 👍

core/loop.gms
tonnrueter: And much like Ostwestfalen contains Bielefeld - an imaginary place people still claim to have visited - the current magicc6 integration contains only nebulous statements with regards to how it obtains it's remind emission data

While the `_prepostsolve` name can be improved upon, the more pressing issue - that i just now realize - is that we're doing a gdx dump *every iteration*, regardless of the `15_climate` module configuration. This needs to be limited, right? However I do not feel proficient enough in GAMS to do this myself. Please help @gabriel-abrahao @LaviniaBaumstark

core/loop.gms
```diff
@@ -149,6 +149,16 @@ o_negitr_disc_cons_drInt_reg(iteration,regi) =
...
+!! retain gdxes of intermediate iterations by copying them using shell
+!! commands
+* if (c_keep_iteration_gdxes eq 1,
+*   put_utility logfile, "shell" /
+*     "cp fulldata_prepostsolve.gdx fulldata_prepostsolve_" iteration.val:0:0 ".gdx";
```
0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q, line 160: What's up with this?

modules/15_climate/ar6/postsolve.gms
```diff
@@ -0,0 +1,163 @@
...
+* $include "./core/magicc.gms";
+* *** execute MAGICC (this is cheap enough, ~2s)
+* Execute "Rscript run_magicc.R";
+* *** read in results
+* Execute "Rscript read_DAT_TOTAL_ANTHRO_RF.R";
+* Execute_Loadpoint 'p15_forc_magicc'  p15_forc_magicc;
+* Execute "Rscript read_DAT_SURFACE_TEMP.R";
+* Execute_Loadpoint 'p15_magicc_temp' pm_globalMeanTemperature = pm_globalMeanTemperature;
+* *** MAGICC only reports unitl 2300:
+* pm_globalMeanTemperature(tall)$(tall.val gt 2300) = 0;
```
0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q, line 29: So this runs MAGICC6, then MAGICC7?

core/loop.gms
tonnrueter: Sidenote: Thanks to @0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q who explained me that the emission data dump happens in [`./core/magicc.c`](https://github.com/remindmodel/remind/blob/develop/core/magicc.gms)

orichters: MAGICC6 gets its data [from the file written here](https://github.com/remindmodel/remind/blob/develop/core/magicc.gms#L121-L191). Agree that you best put that behind some kind of if-clause like 
```
if(sameas("%climate%","magicc7_ar7"), p46_startInIteration = 0);
```
Of course, replace the p46 stuff by the code you want to execute. But there is also the
```
$Ifi "%carbonprice%" == "temperatureNotToExceed" $include "./modules/45_carbonprice/temperatureNotToExceed/realization.gms"
```
construction and the
```
$if %carbonprice% == "off"      pm_macSwitch(emiMacMagpie) = 0;
```
way and I never really understood what to use.

scripts/input/climate_assessment_run.R
```diff
@@ -0,0 +1,355 @@
...
+# Character vector of all required MAGICC7 environment variables
+magiccEnvs <- c(
+    "MAGICC_EXECUTABLE_7"    = magiccBinFile,    # Specifies the path to the MAGICC executable
+    "MAGICC_WORKER_ROOT_DIR" = magiccWorkersDir, # Directory of magicc workers
+    "MAGICC_WORKER_NUMBER"   = 1                 # TODO: Get this from slurm or nproc
+)
+
+gamsEnvs <- c(
+    "R_GAMS_SYSDIR" = gamsRDir
+)
```
0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q, line 180: If this is for the benefit of `gdxrrw`, it is quite pointless.
1. This is run inside a REMIND run, which will not even start unless `gams` is on the PATH, where `gdxrrw` will pick it up.
2. `gdxrrw` will actually be initialised when the `remind2` package is loaded, because it has `gdx` as a dependency, which will do that explicitly.  So for `R_GAMS_SYSDIR` to have any influence on `gdxrrw` behaviour, this would have to be done before `require(remind2)`.

core/loop.gms
0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q: > While the `_prepostsolve` name can be improved upon, the more pressing issue - that i just now realize - is that we're doing a gdx dump _every iteration_, regardless of the `15_climate` module configuration. This needs to be limited, right?

Just put it inside `./modules/15_climate/magicc7_ar6_whatever/postsolve.gms`.

LaviniaBaumstark: in generall, I agre wit Michaja. If we need the gdx dump in many realizations and wtiht the same information (at the same time during solution) we might do it once at a central place

config/scenario_config_magicc.csv
```diff
@@ -0,0 +1,9 @@
+title;start;CES_parameters;optimization;slurmConfig;climate;regionmapping;extramappings_historic;cm_rcp_scen;cm_iterative_target_adj;subsidizeLearning;c_budgetCO2from2020;carbonprice;cm_co2_tax_2020;c_peakBudgYr;cm_CO2priceRegConvEndYr;cm_emiscen;c_regi_earlyreti_rate;c_tech_earlyreti_rate;cm_fetaxscen;cm_co2_tax_growth;cm_bioenergy_SustTax;cm_maxProdBiolc;c_ccsinjecratescen;c_ccscapratescen;cm_CESMkup_build;cm_CESMkup_ind;techpol;regipol;cm_implicitQttyTarget;cm_emiMktTarget;cm_NucRegiPol;cm_CoalRegiPol;cm_altFeEmiFac;cm_POPscen;cm_GDPscen;cm_demScen;cm_oil_scen;cm_gas_scen;cm_coal_scen;c_techAssumptScen;cm_nucscen;cm_so2tax_scen;cm_multigasscen;cm_LU_emi_scen;cm_tradecostBio;cm_1stgen_phaseout;c_SSP_forcing_adjust;cm_APscen;water;cm_EDGEtr_scen;cm_startyear;path_gdx;path_gdx_ref;path_gdx_bau;path_gdx_refpolicycost;description
```
orichters, line 1: This config is outdated, can you please copy the descriptions from the `scenario_config.csv` file. I corrected many spelling errors and content mistakes recently.

modules/15_climate/ar6/declarations.gms
```diff
@@ -0,0 +1,47 @@
...
+  s15_forcing_budgetiterationoffset "offset for the calculation of iteratively adjusted budget"
+*** to reach specified forcing level in 2100 in W/m^2 -- best guess is 2010 
+*** value of respective forcing category minus 0.7
+  s15_forcing_budgetiterationoffset_tax "offset for the calculation of iteratively adjusted budget"
+*** to reach specified forcing level in 2100 in W/m^2 -- best guess taken from analysis of ERL paper results
+s15_tempOffset2010 "mean temperature in 2010 from MAGICC"
+;
+
+parameters
+p15_oghgf_pfc    	"exogenous forcings from RCP all in W/m^2: PFCs",
```
orichters, line 19: does "all in W/m^2" apply to the other parameters? Because ideally you would add the unit to all parameters separately, which is helpful if you read them from the gdx and don't have them together like they are here.

gabriel-abrahao (commented):
I disagree with this one. The only thing that is really MAGICC7 specific in this realization is the parameter sets we currently use, but even that would take just new files, no changes in the interface, to use another OpenSCM-available model such as FAIR. RIght now, using FAIR instead of MAGICC would just be a matter of adding a flag to one of the Python calls. So I really think MAGICC7 should not be in the name of the new realization.

tonnrueter:
> I disagree with this one. The only thing that is really MAGICC7 specific in this realization is the parameter sets we currently use, but even that would take just new files, no changes in the interface, to use another OpenSCM-available model such as FAIR. RIght now, using FAIR instead of MAGICC would just be a matter of adding a flag to one of the Python calls. So I really think MAGICC7 should not be in the name of the new realization.

I think it'd be beneficial to have separate module realizations for different SCMs. I see your point here, however currently MAGICC is the de-facto standard for SCM integration in REMIND. Also there's more to come regarding climate assessment integration: I'm working on a [PIAM pacakage `remindClimateAssessment`](https://github.com/pik-piam/remindClimateAssessment) that will interface REMIND with SCMs more broadly. Once we have that adding further realizations, e.g. with FAIR as SCM should be easy, yet it makes sense to me to keep them in dedicated climate module realization e.g. `modules/15_climate/FAIR`

config/scenario_config_magicc.csv
LaviniaBaumstark: could this file be deleted? Is it still necessary?

LaviniaBaumstark:
> > I disagree with this one. The only thing that is really MAGICC7 specific in this realization is the parameter sets we currently use, but even that would take just new files, no changes in the interface, to use another OpenSCM-available model such as FAIR. RIght now, using FAIR instead of MAGICC would just be a matter of adding a flag to one of the Python calls. So I really think MAGICC7 should not be in the name of the new realization.
> 
> I think it'd be beneficial to have separate module realizations for different SCMs. I see your point here, however currently MAGICC is the de-facto standard for SCM integration in REMIND. Also there's more to come regarding climate assessment integration: I'm working on a [PIAM pacakage `remindClimateAssessment`](https://github.com/pik-piam/remindClimateAssessment) that will interface REMIND with SCMs more broadly. Once we have that adding further realizations, e.g. with FAIR as SCM should be easy, yet it makes sense to me to keep them in dedicated climate module realization e.g. `modules/15_climate/FAIR`

I see Gabriel's point. But that would also mean, that `ar6` is no better name. Than the moduel should be caled `climate_assessmant` or ` openSCM`, etc.

config/scenario_config_magicc.csv
orichters: was deleted already

gabriel-abrahao:
I agree, I'm in favor of `openSCM`.

gabriel-abrahao (approved):
There was a mistake on the call to the TIRF script, which I fixed but is untested since the cluster is down. As the old implementation is very likely to stop working on the new cluster, I'm approving this PR anyway so the people who need the TIRF (@fpiontek @orichters ?) can at least try. Me and Tonn will be testing this as soon as the cluster is back.

PR #1658, opened 2024-04-17
Extend cs2 historical timespan

lecfab:
## Purpose of this PR

In CompareScenario2, when available, historical/projection data will be yearly until 2025 (instead of 2020), and then shown every 5 years. In the image below, there will simply be more little red crosses between 2020 and 2025.

## Type of change
- [x] Minor change 

## Further
- update [default profile generator](https://github.com/pik-piam/remind2/blob/master/inst/markdown/compareScenarios2/cs2_main.Rmd#L26-L28), see remind2 PR https://github.com/pik-piam/remind2/pull/588
- do you think we should put `"yearsBarPlot": "c(2020, 2030, 2050)"` (with 2020 instead of 2010) as it seems more relevant for "current value"?


orichters (approved):
approve and would be in favor to plot 2020 values in the bar plots.

PR #1660, opened 2024-04-22
set cm_wasteIncinerationCCSshare to off in all scenario configurations until remindmodel/development_issues#274 is addressed

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
- [x] Bug fix


robertpietzcker:
I commented in the issue that I think this is on purpose, not a bug - but let's wait for comments from the others before going forward here

robertpietzcker:
(and I am sorry for not reacting earlier - I am getting tagged in many of the industry changes, and as I don't feel responsible for all the industry stuff I only read up on the git messages concerning industry stuff irregularly. It would be great if already the header of the git message would include a "you are ASSIGNED" information... then I would have reacted faster)

JakobBD:
I agree with Robert here, unabatable emissions aren't an improvement over the status quo. See #1662 for an alternative suggestion.

PR #1668, opened 2024-04-30
Exposing SE electricity tax rate tech specific ramp up logistic function parameters as a switch

Renato-Rodrigues:
## Purpose of this PR

- Expose the exponential curve parameter a and b to be changed directly in the scenario config.

## Checklist:

* Test runs are here: 

- p21_tau_SE_tax disabled
`/p/projects/ecemf/REMIND/2040_scenarios/v05_2024_04_09_rev1/output/05_Nzero_57_bio7p5_seTrade_EU27_off_2024-04-30_18.47.42`
- a = 0.1, b = 10 (current default)
`/p/projects/ecemf/REMIND/2040_scenarios/v05_2024_04_09_rev1/output/05_Nzero_57_bio7p5_seTrade_EU27_def_2024-04-30_18.47.23`
- a = 0.1, b = 20
`/p/projects/ecemf/REMIND/2040_scenarios/v05_2024_04_09_rev1/output/05_Nzero_57_bio7p5_seTrade_EU27_b20_2024-04-30_18.48.01`
- a = 0.1, b = 30
`/p/projects/ecemf/REMIND/2040_scenarios/v05_2024_04_09_rev1/output/05_Nzero_57_bio7p5_seTrade_EU27_b30_2024-04-30_18.48.20`
- a = 0.1, b = 150 (it should be very close to off)
`/p/projects/ecemf/REMIND/2040_scenarios/v05_2024_04_09_rev1/output/05_Nzero_57_bio7p5_seTrade_EU27_b150_2024-04-30_18.48.39`

`/p/projects/ecemf/REMIND/2040_scenarios/v05_2024_04_09_rev1/compScen-h2Tax-2024-05-10_10.40.15-EU27.pdf`

### DIFF
--- a/main.gms
+++ b/main.gms
@@ -1408,7 +1408,8 @@ $setGlobal cm_import_ariadne  off !! def off
 *** For example: "2030.2050.MEA.EU27_regi.seh2 0.5, 2030.2050.MEA.DEU.seh2 0.3".
 $setGlobal cm_trade_SE_exog off !! def off
 *** cm_SEtaxRampUpParam "set the logistic function parameters to describe relationship between SE electricity tax rate and share of technology in total electricity demand"
-$setGlobal cm_SEtaxRampUpParam  GLO.te.a 0.4, GLO.te.b 10    !! def = GLO.te.a 0.4, GLO.te.b 10
+*** cm_SEtaxRampUpParam = "off" disables v21_tau_SE_tax 
+$setGlobal cm_SEtaxRampUpParam  GLO.elh2.a 0.4, GLO.elh2.b 10    !! def = GLO.te.a 0.4, GLO.te.b 10
 *** cm_EnSecScen             "switch for running an ARIADNE energy security scenario, introducing a tax on PE fossil energy in Germany"
 *** switch on energy security scenario for Germany (used in ARIADNE project), sets tax on fossil PE
 *** switch to activate energy security scenario assumptions for Germany including additional tax on gas/oil
--- a/main.gms
+++ b/main.gms
@@ -1413,7 +1413,7 @@ $setGlobal cm_trade_SE_exog off !! def off
 *** The parameter a defines how fast the tax increases with increasing share, with 4/a being the percentage point range over which the tax value increases from 12% to 88%
 *** The parameter b defines at which share the tax is halfway between the value at 0 share and the maximum value (defined by a region's electricity tax and the electricity grid cost) that it converges to for high shares.
 *** Example use: 
-*** cm_SEtaxRampUpParam = "GLO.elh2.a 0.4, GLO.elh2.b 20" sets the logistic function parameter values a=0.4 and b=10 for electrolysis (elh2) to all model regions (GLO). 
+*** cm_SEtaxRampUpParam = "GLO.elh2.a 0.4, GLO.elh2.b 30" sets the logistic function parameter values a=0.4 and b=30 for electrolysis (elh2) to all model regions (GLO). 
 *** cm_SEtaxRampUpParam = "off" disables v21_tau_SE_tax 
 *** For details, please see ./modules/21_tax/on/equations.gms.
 $setGlobal cm_SEtaxRampUpParam  GLO.elh2.a 0.4, GLO.elh2.b 30    !! def = GLO.elh2.a 0.4, GLO.elh2.b 30
--- a/main.gms
+++ b/main.gms
@@ -1410,13 +1410,13 @@ $setGlobal cm_trade_SE_exog off !! def off
 *** This allows to manually adjust the ramp-up curve of the SE tax on electricity. It is mainly used for taxing electricity going into electrolysis for green hydrogen production.
 *** The ramp-up curve is a logistic function that determines how fast taxes increase with increasing share of technology in total power demand.
 *** This essentially makes an assumption about to what extend the power demand of electrolysis will be taxed and how much tax exemptions there will be at low shares of green hydrogen production.
-*** The parameter a defines how fast the tax increases with increasing share, while the parameter b defines at which share
-*** the tax is halfway between the value at 0 share and the maximum value (defined by a region's electricity tax and the electricity grid cost) that it converges to for high shares.
+*** The parameter a defines how fast the tax increases with increasing share, with 4/a being the percentage point range over which the tax value increases from 12% to 88%
+*** The parameter b defines at which share the tax is halfway between the value at 0 share and the maximum value (defined by a region's electricity tax and the electricity grid cost) that it converges to for high shares.
 *** Example use: 
 *** cm_SEtaxRampUpParam = "GLO.elh2.a 0.4, GLO.elh2.b 20" sets the logistic function parameter values a=0.4 and b=10 for electrolysis (elh2) to all model regions (GLO). 
 *** cm_SEtaxRampUpParam = "off" disables v21_tau_SE_tax 
 *** For details, please see ./modules/21_tax/on/equations.gms.
-$setGlobal cm_SEtaxRampUpParam  GLO.elh2.a 0.4, GLO.elh2.b 10    !! def = GLO.elh2.a 0.4, GLO.elh2.b 10
+$setGlobal cm_SEtaxRampUpParam  GLO.elh2.a 0.4, GLO.elh2.b 30    !! def = GLO.elh2.a 0.4, GLO.elh2.b 30
 *** cm_EnSecScen             "switch for running an ARIADNE energy security scenario, introducing a tax on PE fossil energy in Germany"
 *** switch on energy security scenario for Germany (used in ARIADNE project), sets tax on fossil PE
 *** switch to activate energy security scenario assumptions for Germany including additional tax on gas/oil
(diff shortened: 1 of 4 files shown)


main.gms
```diff
@@ -1407,6 +1407,9 @@ $setGlobal cm_import_ariadne  off !! def off
 *** then the values from the region group disaggregation will be overwritten by this region-specific value.
 *** For example: "2030.2050.MEA.EU27_regi.seh2 0.5, 2030.2050.MEA.DEU.seh2 0.3".
 $setGlobal cm_trade_SE_exog off !! def off
+*** cm_SEtaxRampUpParam "set the logistic function parameters to describe relationship between SE electricity tax rate and share of technology in total electricity demand"
+*** cm_SEtaxRampUpParam = "off" disables v21_tau_SE_tax 
+$setGlobal cm_SEtaxRampUpParam  GLO.elh2.a 0.4, GLO.elh2.b 10    !! def = GLO.elh2.a 0.4, GLO.elh2.b 10
```
fschreyer, line 1412: Let's be a bit more verbose here and add

```

*** This allows to manually adjust the ramp-up curve of the SE tax on electricity. It is mainly used for taxing electricity going into electrolysis for green hydrogen production.
*** The ramp-up curve is a logistic function that determines how fast taxes increase with increasing share of technology in total power demand.
*** This essentially makes an assumption about to what extend the power demand of electrolysis will be taxed and how much tax exemptions there will be at low shares of green hydrogen production.
*** The parameter a defines how fast the tax increases with increasing share, while the parameter b defines at which share
*** the tax is halfway between the value at 0 share and the maximum value (defined by a region's electricity tax and the electricity grid cost) that it converges to for high shares.
*** Example use: ...
*** For details, please see ./modules/21_tax/on/equations.gms.
```

fschreyer (approved):

main.gms
Renato-Rodrigues: https://github.com/remindmodel/remind/pull/1668/commits/87f9027e8b8a3ea2e07c52b9d344db20ccbefe68

Renato-Rodrigues:
Here are the results of different parametrizations of the b parameters:

run name | type | se_trade | b
-- | -- | -- | --
05_Nzero_57_bio7p5_EU27 | old reference | no | 10
05_Nzero_57_bio7p5_seTrade_EU27 | new reference | yes | 10
05_Nzero_57_bio7p5_seTrade_EU27_off | par test | yes | disabled
05_Nzero_57_bio7p5_seTrade_EU27_b30 | par test | yes | 30
05_Nzero_57_bio7p5_seTrade_EU27_b20 | par test | yes | 20
05_Nzero_57_bio7p5_seTrade_EU27 | par test | yes | 10

`/p/projects/ecemf/REMIND/2040_scenarios/v05_2024_04_09_rev1/compScen-h2Tax-2024-05-10_10.40.15-EU27.pdf`

image

@fschreyer and @robertpietzcker 
Should we disable the elh2 tax rate (05_Nzero_57_bio7p5_seTrade_EU27_off scenario) or increase its default value (05_Nzero_57_bio7p5_seTrade_EU27_b20 or 05_Nzero_57_bio7p5_seTrade_EU27_b30)?

robertpietzcker:
This seems very different from the effects that Felix observed (if I remember correctly), where adding the tax strongly decreased electrolysis-based hydrogen. 

or do I remember your results wrongly, @fschreyer ?

actually, I don't get the results - shouldn't the second bar (triangle) be with the current default tax, and the third bar (plus) with the tax turned off? why then does the one with the tax turned off have less electrolysis and more biomass hydrogen?

Renato-Rodrigues:
> actually, I don't get the results - shouldn't the second bar (triangle) be with the current default tax, and the third bar (plus) with the tax turned off? 

Yes.

> why then does the one with the tax turned off have less electrolysis and more biomass hydrogen?

I see the opposite of what you describe in EUR by 2050. 
The plus bar (tax turned off) has more electrolysis (middle tone green) than the default parametrization (second bar, triangle shape). 
The biomass (light green) values are stable, and there is no gas anymore with the tax turned off or with the reduced tax effects.

In Germany, more or less the same happens but at a lower degree due to the imports.

main.gms
```diff
@@ -1407,6 +1407,16 @@ $setGlobal cm_import_ariadne  off !! def off
 *** then the values from the region group disaggregation will be overwritten by this region-specific value.
 *** For example: "2030.2050.MEA.EU27_regi.seh2 0.5, 2030.2050.MEA.DEU.seh2 0.3".
 $setGlobal cm_trade_SE_exog off !! def off
+*** This allows to manually adjust the ramp-up curve of the SE tax on electricity. It is mainly used for taxing electricity going into electrolysis for green hydrogen production.
+*** The ramp-up curve is a logistic function that determines how fast taxes increase with increasing share of technology in total power demand.
+*** This essentially makes an assumption about to what extend the power demand of electrolysis will be taxed and how much tax exemptions there will be at low shares of green hydrogen production.
+*** The parameter a defines how fast the tax increases with increasing share, while the parameter b defines at which share
```
robertpietzcker, line 1413: I would make it more precise: "The parameter a defines how fast the tax increases with increasing share, with a*100 being the percentage point range over which the tax value increases from 12% to 88%"

robertpietzcker:
> I see the opposite of what you describe in EUR by 2050.

haha, funny, I didn't look at the legend but somehow from memory associated the light green with electrolysis. Sorry :-)

ok, I would put the default value of b to 30.
with b=30, the share of electricity used for hydrogen is ~10-20% in most regions in 2050 (although only 4% in FRA and DEU), which I think is in a reasonable range for a high-VRE power system where electrolysis still has a stabilizing impact on the power system and is thus likely treated favorably by the tax system.

robertpietzcker (approved):

main.gms
Renato-Rodrigues: https://github.com/remindmodel/remind/pull/1668/commits/1b86906386c55ecbf44005eb1b1f2266379f771d

Renato-Rodrigues:
> ok, I would put the default value of b to 30. 

https://github.com/remindmodel/remind/pull/1668/commits/1b86906386c55ecbf44005eb1b1f2266379f771d

fschreyer:
> ok, I would put the default value of b to 30.

I agree for EU that the policy scenarios look ok. I am not sure whether we should do that in general. The thing is if we have so much tax exemptions for hydrogen globally, Npi scenarios will probably see a lot of e-fuel production in MEA etc. again, which we deem unrealistic (see [here](https://github.com/remindmodel/development_issues/issues/214)). I would rather only change it to 30 for EU and leave the rest of the world as it is. 

> there is not much effect on bio based hydrogen although.

Actually, you do not see so much difference across your scenarios because the demand-side/CCS/bioenergy assumptions are more important than the supply price. If you do a low CCS/bio run that needs to phase-out fossil fuels even more, then higher hydrogen penetrations are to be expected and I would presume the model goes to higher hydrogen prices/quantities even with the same tax parameters. 

On the biogenic H2: The levels in those runs on global level are comparable to what we had before the H2 update. They actually did not change that much. However, in Ariadne we used to have lower biogenic H2 because we had low Bio/CCS assumptions that forced a fossil phase-out. In your net-zero runs, quite some fossils are still used in 2050 and after, which frees more bioenergy to be used elsewhere.

Renato-Rodrigues:
> > ok, I would put the default value of b to 30.
> 
> I agree for EU that the policy scenarios look ok. I am not sure whether we should do that in general. The thing is if we have so much tax exemptions for hydrogen globally, Npi scenarios will probably see a lot of e-fuel production in MEA etc. again, which we deem unrealistic (see [here](https://github.com/remindmodel/development_issues/issues/214)). I would rather only change it to 30 for EU and leave the rest of the world as it is. We can also make it a tax policy which is only on in policy runs, but I would not like to decide that here.
> 

Do you want me to set a NPi run to double check that effect?

I can also just change the default values to be b=10 globally, and b=30 only to EUR regions if you two agree with that. Just let me know.

robertpietzcker:
well, if you are not using the cluster otherwise, I think having an NPi to see if the synfuel-problem appears again at b=30 (and goes away at b=10)  would be useful. 

In Renatos NetZero runs, the synfuel prices in MEA are around 40-60$/GJ, while oil prices stay around 15$/GJ - so I would expect that very little synfuel comes in NPi scenarios 

(if time is scarce, I am also fine with having b=30 only in EU and 10 elsewhere, but this may bring critical questions in the future)

Renato-Rodrigues:
Below you can find a current NPi run with variations of the above mentioned parameter. 
These results should change a bit after a new calibration, as they assume the model calibrated under the previous parametrization (b=10 if I am not mistaken, and not b=30 as we are proposing). 

image

I wouldn't say that the values under b=30 are that crazy for the end of the century. 
Opinions? @fschreyer @robertpietzcker

Renato-Rodrigues:
you can find the test runs and comparison pdfs for the NPi scenario here:
`/p/projects/ecemf/REMIND/trunk/remind_14_05_2024_h2TaxTests`

robertpietzcker:
looks good to me  (given that the NPi doesn't represent a "no policy" world but rather "starting from today, with weak policies" - I think it is not unrealistic that in the long term (after 2050), more expensive oil resources may be replaced by synfuels...

it might be interesting to get a comment by @nicobauer on this :-)

PR #1670, opened 2024-05-08
don't skip tests when running 'make test-full'.

orichters:
## Purpose of this PR

- we run `make test-full` basically in the AMTs and maybe before a major release. I think we shouldn't skip tests before a previous one failed in the AMTs. For example in `/p/projects/remind/modeltests/remind/tests/test-full-2024-05-03.log` many relevant checks were skipped because of some unreasonable setting in one config file. That shouldn't happen.
- remove IKEA config that was errornously added by @lecfab and caused the issue
- increase memory to 8000 for automated cs2 calls in coupled runs to avoid out-of-memory fails

## Type of change

- [x] Bug fix 

## Checklist:

- [x] All automated model tests pass (`FAIL 0` in the output of `make test`): [ FAIL 0 | WARN 0 | SKIP 6 | PASS 82 ]


dklein-pik:
Hi, do I understand correctly that the skip was introduced because it makes no sense to carry out subsequent tests if one of the predecessors has failed? But in this particular case (wrong config) the subsequent tests would have made sense? Is that why all tests should now always be executed for long tests, even if predecessors have failed?

orichters:
I think Mikas idea was to skip testOneRegi if gamscompile fails, for example. And as `make test` is mostly run in interactive mode, people fix errors and then run again – so that still makes some sense.

But for `make test-full`, it is definitely better to try all tests to avoid that relevant fails are not noticed because they are skipped.

Thinking of it, it also might make sense to drop this "skip if previous fails" also for `make test`, but someone else should decide that.

dklein-pik (approved):

PR #1685, opened 2024-05-23
Fix minor bug in profiles.json (cs2)

fbenke-pik:
## Purpose of this PR

This bugfix ensures that the default and defaultHTML profiles actually source the markdown files from remind2.


dklein-pik (approved):
Great, thanks!

PR #1687, opened 2024-05-23
Options for regionally differentiated floor costs for learning technologies

cchrisgong:
## Purpose of this PR
Implement a scenario switch for the IKEA project, where scenarios for regional differentiation of (power sector) technology costs are called for. See https://github.com/remindmodel/development_issues/issues/230

So far, there are no regionally differentiated projections of technology costs in REMIND (other than initial historical cost values). For non-learning technologies (coal, gas, nuclear, biomass, ccs etc), p_inco0 file in core/input contains the regionally different technology costs, but only initial points of these curves are used as "historical cost values". In model runs, the costs of these technologies converge to a global average by a convergence date (currently 2070), which means some of the technologies increase in costs (if they are already below global average), such as pc for IND and CHA. Increasing coal and gas costs in developing countries without a CO2 price in the model could give an overly optimistic scenario, especially for SSP2 there is no GDP/cap convergence, and energy demand per capita also does not converge. This call into question why coal plant capital costs must be uniform across regions. 

One immediate remedy would be to use the cost trajectories provided by the IEA and other sources in p_inco0. However, the learning curves are hard coded from the input data which projects 2015 till 2040, better treatment of regional costs are possible. 

In addition, in default REMIND scenarios, for learning technologies, solar, wind, battery, etc, only their regionally differentiated "initial technology costs" are read in. In model runs, a global learning rate as well as as a global floor cost is applied. Sometimes the global floor costs are higher than some of the regional costs, causing wind onshore for example, for CHA to increase in costs over time, which there is no current reason to believe it will happen (costs continue to drop, although will flatten at some point).

Based on this, we designed two new scenarios (in addition to the default assumption where global convergence of costs for non-learning tech, and global convergence of learning rate and one global floor cost are used). 

1) price structure prevails ("pricestruc"). This is when the new floor costs for regions all equal or lie below the current global floor cost. The value of the new regional floor cost proportional to the highest among them is the same as the value of the latest historical cost of the region to the highest cost region among them. In short
new_floor_cost(REG) = old_floor_cost(REG) * cost(2015,REG) / max( reg, cost(2015,reg) )

In this way, the price structure between regions is preserved from the historical values. So whatever reasons which cause SSA's Solar PV to be more expensive than that in China (for example due to IP, due to lack of industrial learning and manufacturing expertise, or automation, despite cheaper labor costs), this reason is preserved over time. 

2) technological transfer ("techtrans"). This scenario's floor costs for regions are proportional to per capital PPP income. So this simulates a situation where each region is able to "make their own technologies". Of course, the poorest region will have the lowest floor cost, since the tech costs are entirely determined by labor costs, and not by IP rents, or other exogenous political economy factors. 
new_floor_cost(REG) = old_floor_cost(REG) * GDPPPP_cap(2020,REG) / max( reg, GDPPPP_cap(2020,reg) )

## Type of change

- [ ] Bug fix 
- [ ] Refactoring
- [x] New feature 
- [ ] Fundamental change

## Checklist:

* Test runs are here: 


core/datainput.gms
```diff
@@ -321,9 +321,49 @@ p_inco0(ttot,regi,teRegTechCosts)  = (1 + p_tkpremused(regi,teRegTechCosts) ) *
...
+
+*** In case regionally differentiated investment costs should be used, the corresponding entries are revised:
+$if %cm_techcosts% == "REG"   pm_data(regi,"inco0",teRegTechCosts) = p_inco0("2015",regi,teRegTechCosts);
+loop(teRegTechCosts$(sameas(teRegTechCosts,"spv") ),
+$if %cm_techcosts% == "REG"   pm_data(regi,"inco0",teRegTechCosts) = p_inco0("2020",regi,teRegTechCosts);
+);
+p_maxSpvCost = SMax(regi, p_inco0("2020",regi,"spv"));
+$if %cm_techcosts% == "REG"   pm_data(regi,"inco0","storspv") = p_inco0("2020",regi,"spv")/p_maxSpvCost * pm_data(regi,"inco0","storspv");
+
+$if %cm_techcosts% == "REG"   pm_data(regi,"incolearn",teLearn(te)) = pm_data(regi,"inco0",te) - pm_data(regi,"floorcost",te) ;
```
lecfab, line 367: I think there has been a merge conflict here: same concept above and below line 367, starting with comment _*** In case regionally differentiated investment costs should be used, the corresponding entries are revised:_.
But the chunk below is newer, I had modified it for the non-learning-tech convergence switch.

Could you please integrate the relevant modifications in it and remove the chunk above?

core/datainput.gms
```diff
@@ -733,7 +773,7 @@ pm_regiEarlyRetiRate(t,regi,"biohp")   = 0.5 * pm_regiEarlyRetiRate(t,regi,"bioh
 
 $ifthen.tech_earlyreti not "%c_tech_earlyreti_rate%" == "off"
 loop((ext_regi,te)$p_techEarlyRetiRate(ext_regi,te),
-  pm_regiEarlyRetiRate(t,regi,te)$(regi_group(ext_regi,regi) and (t.val lt 2035 or sameas(ext_regi,"GLO"))) = p_techEarlyRetiRate(ext_regi,te);
+  pm_regiEarlyRetiRate(t,regi,te)$(regi_group(ext_regi,regi) and (t.val lt 2065 or sameas(ext_regi,"GLO"))) = p_techEarlyRetiRate(ext_regi,te);
```
lecfab, line 776: Why is it that the early retirement zone expands?

lecfab:
Maybe we can link it with https://github.com/remindmodel/development_issues/issues/230 ?

Also, for the 2015 or 2020 price structure, couldn't we take that of a mature technology, that possibly reflects the regional costs better?

modules/32_power/IntC/datainput.gms
```diff
@@ -44,6 +44,7 @@ f32_factorStorage(all_regi,"windoff") = f32_factorStorage(all_regi,"wind");
 f32_factorStorage(all_regi,"wind")      = 1.35 * f32_factorStorage(all_regi,"wind"); 
 $ENDIF.WindOff
 p32_factorStorage(all_regi,all_te) = f32_factorStorage(all_regi,all_te);
+p32_factorStorage(all_regi,"spv") = f32_factorStorage(all_regi,"spv") *1.3;
```
lecfab, line 47: Why 1.3? is it to account for pv's lower capacity factor than wind? but then why does wind get 1.35 as well?

core/datainput.gms
cchrisgong: with "lt 2035", what happens is that if for some scenario, one would like to set for example no early retirement of pc for IND, one would think setting 'c_tech_earlyreti_rate = "..., IND_regi.pc 0"' will do the trick, but it actually only takes care of initial years up till 2030, but afterwards it will be 9% *per year* of coal that will be early retired (again, in REMIND coal is never profitable), which means in around 10 years, almost all coal can be retired. This makes simulating the political economy difficulty of early retiring coal very difficult, so one quick trick would be to extend it till middle of the century. 

in short, it allows regions to have longer periods to have designated early retirement rate (specified by c_tech_earlyreti_rate), but if one wants one can also make it a flag, @stephenbi @Renato-Rodrigues have an opinion?

cchrisgong:
> Also, for the 2015 or 2020 price structure, couldn't we take that of a mature technology, that possibly reflects the regional costs better?

price structure has to do with the supply chain structure of a particular technology in a particular region. Some regions for example might lag behind in one technological development (so rely on mostly imported parts and therefore has a price structure that can be higher than per capita income), but not in some others. There are plenty example for this, Bolivia and Argentina have national policies to build lithium batteries and EVs, but they rely on imports for say, solar PVs and high value added chemicals. Or Indonesia through export ban of Nickel has accumulated higher-value added Nickel processing facilities inside the country, which then enable their green tech which depends on Nickel to have a cost advantage against other countries who import. But Indonesia might lack other stuff, say lithium for making EVs and batteries. So we cannot just take an "ideally mature" technology that reflects national cost structure, because of the sector and supply chain specific dependencies

modules/32_power/IntC/datainput.gms
cchrisgong: ah, I'll remove this, this was left over from my trying to adjust spv downward due to CA request (from IKEA)

core/datainput.gms
cchrisgong: done

lecfab (approved):
All good on my side!

cchrisgong:
fixed by this: https://github.com/remindmodel/remind/pull/1739

PR #1688, opened 2024-05-24
Reduce CCS 2030 upper bound to only include 40% of announced projects as default. Changeable with switch.

amerfort:
## Purpose of this PR
We want to realistically constrain near-term CCS deployment in line with IEA-data on projects that are operational or under construction as well as announced projects.
From all announced projects we **now** assume by default that only 40% will be realised by 2030 (previously the upper limit on CCS was 100% of all announced projects).
This assumption can be altered with the new switch ~~c_SharenonFIDCCS2030~~ `c_fracRealfromAnnouncedCCScap2030`

Furthermore, the very stringent CCS limit for DEU in 2030 was removed, as it is no longer needed. 
The reason is that CCS is now prominently discussed in German politics and the law that previously forbid it will be changed.

**Edits:** 

- [x] I renamed the switch such that it is correct (FID was not a criterion before, just realized vs. announced projects) and such that it can be understood intuitively.
- [x] As suggested in the review, I added a conversion factor `s_MtCO2_2_GtC`

## Type of change

## Checklist:

* Test runs are here: 
`/p/projects/ecemf/REMIND/2040_scenarios/_sharedRuns/v06_2024_05_14_rev1_x_CCStest_Anne/compScen-ccs_near-term-2024-05-24_10.13.14-EU27-short.pdf`

### DIFF
--- a/core/bounds.gms
+++ b/core/bounds.gms
@@ -393,7 +393,7 @@ $endif
 *** -------------------------------------------------------------------------------------------------------------
 *AM* No geological storage before 2025 (omitting approx. 40 MtCCS/yr globally in 2020 for Enhanced Oil Recovery)
 *AM* Lower limit for 2025 and 2030 is capacities of all projects that are operational and under construction from project data base
-*AM* Upper limit for 2025 and 2030 additionally includes 40% (default, or changed by c_SharenonFIDCCS2030) announced/planned projects from project data base
+*AM* Upper limit for 2025 and 2030 additionally includes 40% (default, or changed by c_sharemaxCCScap2030) announced/planned projects from project data base
 *AM* In nash-mode regions cannot easily share ressources, therefore CCS potentials are redistributed in Europe: 
 *AM* Potential of EU27 regions is pooled and redistributed according to GDP (Only upper limit for 2030)
 *AM* Norway and UK announced to store CO2 for EU27 countries. So 50% of Norway and UK potential in 2030 is attributed to EU27-Pool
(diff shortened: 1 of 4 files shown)


core/bounds.gms
```diff
@@ -393,19 +393,16 @@ $endif
...
 *LP* if c_ccsinjecratescen=0 --> no CCS at all and vm_co2CCS is fixed to 0 before, therefore the upper bound is only set if there should be CCS!
 *** -------------------------------------------------------------------------------------------------------------
 
 if ( c_ccsinjecratescen gt 0,
-  vm_co2CCS.up(ttot,regi,"cco2","ico2","ccsinje","1")$(ttot.val ge 2005 AND ttot.val lt 2025) = 0;
-	vm_co2CCS.up("2025",regi,"cco2","ico2","ccsinje","1") = pm_boundCapCCS("2025",regi,"up")/(1000*11/3);
-	vm_co2CCS.up("2030",regi,"cco2","ico2","ccsinje","1") = pm_boundCapCCS("2030",regi,"up")/(1000*11/3);
-	vm_co2CCS.lo("2025",regi,"cco2","ico2","ccsinje","1") = pm_boundCapCCS("2025",regi,"low")/(1000*11/3);
-	vm_co2CCS.lo("2030",regi,"cco2","ico2","ccsinje","1") = pm_boundCapCCS("2030",regi,"low")/(1000*11/3);
+  vm_co2CCS.lo(t,regi,"cco2","ico2","ccsinje","1")$(t.val le 2030) = pm_boundCapCCS(t,regi,"low")$(t.val le 2030) / (1000*11/3);
```
robertpietzcker, line 404: I think it would be great to have conversion factor `s_unit1_2_unit2` for this "(1000*11/3)"

core/bounds.gms
```diff
@@ -393,19 +393,16 @@ $endif
...
+  vm_co2CCS.up(t,regi,"cco2","ico2","ccsinje","1")$(t.val le 2030) = (pm_boundCapCCS(t,regi,"low")$(t.val le 2030) + (pm_boundCapCCS(t,regi,"up")$(t.val le 2030) - pm_boundCapCCS(t,regi,"low")$(t.val le 2030)) * c_SharenonFIDCCS2030) / (1000*11/3);
```
robertpietzcker, line 405: why le2030? is it realistic that any project without FID will be built by 2025? so shouldn't we have a lower bound on 2025?

robertpietzcker (approved):

robertpietzcker:
could you copy in as a comment 2-4 plots with the main changes in the results (from a compscen with old/new version)- CCS deployment over time for EU and other world regions (in case the changes also affected other world regions, I am not totally sure about this from the description), ...
sure, the compScen has it all, but having to download it first makes it much less likely that people briefly checking what this change did will actually see the plots :-)

amerfort:
Very good point, Robert! Thanks a lot!
Here is an EU-27 plot with scenarios with the largest CCS deployment from Renatos EU2040 target analysis: red is the old version and green is with reduced 2025-2030 capacities. You can also see that Germany's very strict CCS cap was lifted and the new version now deploys some CCS (as a proxy for a shared EU-wide CCS infrastructure)
image
and here is the H12 representation of the same runs:
image

PR #1697, opened 2024-06-10
Bugfix: exogenous secondary energy trade prices 

Renato-Rodrigues:
## Purpose of this PR

- Fix bug where secondary energy trade prices are not being defined
- Add limit to secondary energy trade prices, as exogenously enforced trade can cause unreasonable monetary pressure on macroeconomic accounts. More details below:

```
Model marginal prices are not necessarily competitive when we enforce exogenously traded quantities.
This could cause the secondary energy trade flows to "eat" a bigger amount of the country trade budget than they should, which could cause infeasibilities in the equations `q23_limit_debt_growth` and `qm_budget`.
Additional secondary energy traded prices upper bounds were added to avoid this issue.
```

## Important

The current allowed range of secondary energy trade prices is quite strict in the code:
- H2 and seliqsyn exporting prices can assume values in between 1$/kg and 2$/kg.
- seliqbio exporting prices can assume values in between 5 US$2005/GJ and 10 US$2005/GJ.

These value corridors should be revised by people that are more expert on the topic if necessary to reflect reasonable cost terms in the REMIND reporting variables.

## Type of change

- [x] Bug fix 

## Checklist:

* Test runs are here: 
`/p/projects/ecemf/REMIND/2040_scenarios/v07_2024_06_04_rev1_final/output/`


robertpietzcker:
Is there a reason the limits range is so small? I would rather go for 1-10$/kg for H2 and 5-50$/GJ for bioliquids.

robertpietzcker (approved):
looks ok

robertpietzcker:
maybe add a lower bound of 0 as well?

Renato-Rodrigues:
> Is there a reason the limits range is so small? I would rather go for 1-10$/kg for H2, and 5-50$/GJ for synfuel and bioliquids.

I set the hydrogen range to be more or less compatible with IEA net-zero assumption: 
https://www.iea.org/reports/global-hydrogen-review-2021/executive-summary#:~:text=The%20potential%20is%20reflected%20in,from%20natural%20gas%20with%20CCUS.

This hydrogen exporting cost range should reflect the value on an EU net-zero and high mitigation World case, with high renewables penetration, with relevant trade levels being present mainly after 2040, and to reflect the cost range of countries with very favorable hydrogen production conditions which could turn them into exporting countries.

In practice, the model will tend always to stay in the upper bound (2$/kg in this case), which I found a reasonable assumption under the above conditions.
I would consider anything around 5$/kg to be extremely pessimistic. 
Values bigger than 5$/kg would most probably cause the issues that I mentioned of too much of the total trade budget being directed to this exogenously defined secondary energy trade flows, hindering the macro balance.

For synfuel and bioliquids, I simply scaled the upper bound to have the same size (2 times more than the lower bound), as I used for hydrogen as a simplified assumption and because the lack of more time to better micro-found the decision.

> maybe add a lower bound of 0 as well?

There was already a lower bound set for prices (see the numbers below). This was already use to avoid cases that lack equation marginal information for setting the secondary energy marginal prices:
- H2 and seliqsyn exporting prices need to be higher than 1$/kg.
- seliqbio exporting prices need to be higher than 5 US$2005/GJ.

Renato-Rodrigues:
The upper bound for the prices were increase for years before 2050 following the reviewer comments.

- H2 and seliqsyn maximum exporting prices = linear decrease from 5$/kg by 2020 to 2$/kg by 2050 and onward
- seliqbio maximum exporting prices = linear decrease from 30 US$2005/GJ by 2020 to 10$/kg by 2050 and onward

PR #1710, opened 2024-06-19
allow to select reporting and MAGICC7_AR6 in one output.R session

orichters:
## Purpose of this PR

- at the moment, for a failed run, you cannot select `reporting` and `MAGICC7_AR6` in one go in `output.R` -> `single`, because `MAGICC7_AR6` knows it needs the mif, so if you select this as `output` option, it does not allow you to select runs where the mif does not exist
- now: if `reporting` is selected as an output scripts (among others), move it to the first position to execute it first and allow to select folders where the mif does not exist (yet)
- The only reason I can imagine this might be problematic is, if there is any script in the [scripts/output/single](https://github.com/remindmodel/remind/tree/develop/scripts/output/single) where it makes sense to run it before the reporting – I doubt it.

## Type of change

- [x] improved usability

## Checklist:

- [x] I added [in-code documentation](https://github.com/remindmodel/remind/blob/develop/main.gms#L120)

### DIFF
--- a/output.R
+++ b/output.R
@@ -130,15 +130,19 @@ if (isFALSE(comp)) comp <- "single" # legacy from times only two comp modes exis
 if (isTRUE(comp)) comp <- "comparison"
 
 if (! exists("output")) {
+  # search for R scripts in scripts/output subfolders
   modules <- gsub("\\.R$", "", grep("\\.R$", list.files(paste0("./scripts/output/", if (isFALSE(comp)) "single" else comp)), value = TRUE))
+  # if more than one option exists, let user choose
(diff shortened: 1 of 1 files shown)


lecfab:
hey, thanks for this update. 
Little issue still:

I have a normal run X, and a run Y where .mif has not been generated automatically.
* When i run output.R with "reporting" option, the choice of runs include X and Y
* When i run output.R with "reporting" and "MAGICC7_AR6" options, then the list of possible runs only includes X 

Any ideas how to fix that?

orichters:
Does you repo contain this change here, which is not part of REMIND master yet? Can you send me a link to your folder?

PR #1736, opened 2024-07-11
Multiple fixes and improvements on MAGICC for damages

gabriel-abrahao:
## Purpose of this PR
Fixes some file paths in the climate_assessment python file to be more generic. We still want it to work outside the cluster @tonnrueter , but that will need some improvements in the interfacing. This change here is better than what we currently have.

Also @fpiontek , you proposed the quickfix with the epsilon value a while ago, but it really doesn't work without it. Should we merge this and make it semi-permanent for now? I found myself always having to stash and reapply this line when I change branches, which feels like a bad sign

Furthermore, this PR makes the standard, between-iteration MAGICC for non-internalized damages use the same Python script used for the TIRF. This allows us a lot more flexibility, and now all MAGICC-related GDXs have yearly data at least until 2200. 

@fpiontek this leaves the matter of what emissions to assume after 2150. Right now it's basically assuming all emissions stop after that, but we can also assume them constant with 2150 or 2100 values. It probably won't make much of a difference given the discounting, but we should have a sensible assumption here anyway.  

## Type of change

- [X] Bug fix 
- [ ] Refactoring
- [ ] New feature 
- [ ] Fundamental change

## Checklist:

* Test runs are here: 

### DIFF
--- a/scripts/input/climate_assessment_openscm_run.py
+++ b/scripts/input/climate_assessment_openscm_run.py
@@ -18,8 +18,8 @@ import json
 
 # TONN REMOVE start
 # If these are already set, it shouldn't override it. We actually may not want to have a default but just throw an error if not set
-os.environ["MAGICC_EXECUTABLE_7"]       =   "/p/projects/piam/abrahao/scratch/module_climate_tests/climate-assessment-files/magicc-v7.5.3/bin/magicc"
-os.environ["MAGICC_WORKER_ROOT_DIR"]    =   "/p/projects/piam/abrahao/scratch/methane/methane_scm/workers"
+os.environ["MAGICC_EXECUTABLE_7"]       =   "/p/projects/rd3mod/climate-assessment-files/magicc-v7.5.3/bin/magicc"
+os.environ["MAGICC_WORKER_ROOT_DIR"]    =   os.environ["PTMP"] + "/"
 
 LOGGER = logging.getLogger(__name__) # We don't need this
 # TONN REMOVE end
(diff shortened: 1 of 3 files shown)


scripts/input/climate_assessment_openscm_run.py
```diff
@@ -18,8 +18,8 @@
 
 # TONN REMOVE start
 # If these are already set, it shouldn't override it. We actually may not want to have a default but just throw an error if not set
```
tonnrueter, line 20: So the code in the following lines actually overwrites the environment variables `MAGICC_EXECUTABLE_7` & `MAGICC_WORKER_ROOT_DIR`. If you want to achieve the behavior described in this comment, change it to

```
class EnvironmentError(Exception):
    pass

for env_var in ["MAGICC_EXECUTABLE_7", "MAGICC_WORKER_ROOT_DIR"]:
    if os.environ.get(env_var, '') == '':
        # If clause covers both cases in which the env var is not set at all 
        # as well as the case in which it is set to an empty string
        raise EnvironmentError(f"{env_var} does not exist")
```

fpiontek (commented):
No problem with adding this, under the assumption that it recovers from the infeasibility in the next iteration. The SCC for the region with the infeasibility would not be calculated properly so I don't know exactly what happens then with the further calculation.

scripts/input/climate_assessment_openscm_run.py
gabriel-abrahao: Nice, thanks!

gabriel-abrahao:
Test runs are still in progress, but I don't see a reason for them not to work. I won't have a chance to check this until late in the day on Monday, so if someone needs this on Monday morning just check if this one here worked and if so feel free to merge:

`/p/projects/piam/abrahao/scratch/tirf_remind/output/R_SSP2EU-PkBudg500-internalize-ar6_2024-07-12_17.42.14`

fpiontek (approved):

gabriel-abrahao:
@tonnrueter @orichters would you mind approving this too? For some reason Franziska's approval is not counting towards letting me merge.

PR #1740, opened 2024-07-12
New cluster: explicitly request email notification for failed runs

dklein-pik:
With the slurm version on the new cluster you have to explicitly request email notification for failed runs. Providing the additional argument "FAIL" also works on the old cluster.


orichters:
What about the other cases? I would add that basically everywhere: https://github.com/search?q=repo%3Aremindmodel%2Fremind+mail-type%3DEND&type=code

dklein-pik:
Yes, of course. You should not work on your day off.

orichters (approved):

PR #1741, opened 2024-07-12
Add OAE technologies and their parameters

katarkow:
## Purpose of this PR

Adding OAE (ocean alkalinity enhancement) technologies to `all_te`  to generate new input data with OAE.  A PR to add it as a CDR option will follow.

### DIFF
--- a/core/input/generisdata_tech.prn
+++ b/core/input/generisdata_tech.prn
@@ -155,14 +155,14 @@ eta                 1.00        1.00        1.00
 omf                 0.02        0.02        0.02
 lifetime              45          45          45
 
-+             weathering         dac
++             weathering         dac      oae_ng      oae_el
 tech_stat                          4
-inco0               0.01       18800
-mix0                   0           0
-eta                 1.00        1.00
-omf                 0.84       0.025
+inco0               0.01       18800         200         400
+mix0                   0           0           0           0
+eta                 1.00        1.00        1.00        1.00
+omf                 0.84       0.025        0.35        0.17
 omv                    0
-lifetime              20          20
+lifetime              20          20          25          25
 incolearn                      14000
 ccap0                         0.0008
 learn                           0.15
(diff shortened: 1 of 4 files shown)


LaviniaBaumstark (commented):
please aso run make test to besure that everyting is working

katarkow:
image

LaviniaBaumstark (approved):

core/input/generisdata_tech.prn
```diff
@@ -155,14 +155,14 @@ eta                 1.00        1.00        1.00
...
 lifetime              45          45          45
 
-+             weathering         dac
++             weathering         dac      oae_ng      oae_el
 tech_stat                          4
-inco0               0.01       18800
-mix0                   0           0
-eta                 1.00        1.00
-omf                 0.84       0.025
+inco0               0.01       18800         200         400
```
strefler, line 160: Should inco0 not be converted to $/tC from $/tCaO as calculated below? And why is inco0 for DACCS a factor of 100 higher?

katarkow: It should, but this inco0 is converted inside the code as the cost USD/tC depends on efficiency which a user can set. The capex is only 15$/tCaO, hence the difference between OAE and DACCS. You can see the calculations below in the same file.

strefler: could you give an example of an efficiency, or even better the whole range we use, so that one can follow these conversions and compare them with DAC?

strefler (approved):

LaviniaBaumstark:
I think, this can be merged. right @strefler @katarkow ?

katarkow:
Almost :) let me just update the input data version and test it quickly and I'll merge it

katarkow:
Change of plans, I will merge this first and then update the input data version in a separate [PR](https://github.com/remindmodel/remind/pull/1758), as some little fixes are need for that input data to work

PR #1760, opened 2024-07-26
Updated documentation on `cm_prtpScen` switch that sets the pure rate of time preference

dbantje:
## Purpose of this PR

Just a small correction of the inline documentation for the switch `cm_prtpScen` which sets the pure rate of time preference. The description of the default option said "1%", but it is actually set to 1.5%:

(from `core/datainput.gms`)
```
*** define pm_prtp according to cm_prtpScen:
if(cm_prtpScen eq 1, pm_prtp(regi) = 0.015);
if(cm_prtpScen eq 3, pm_prtp(regi) = 0.03);
```

## Type of change

- [x] Documentation fix 

## Checklist:

- [x] I adjusted the reporting in [`remind2`](https://github.com/pik-piam/remind2) where it was needed (not needed)
- [x] I adjusted `forbiddenColumnNames` in [readCheckScenarioConfig.R](https://github.com/remindmodel/remind/blob/develop/scripts/start/readCheckScenarioConfig.R) in case the PR leads to deprecated switches (not needed)

### DIFF
--- a/main.gms
+++ b/main.gms
@@ -685,7 +685,7 @@ parameter
   cm_prtpScen               "pure rate of time preference standard values"
 ;
   cm_prtpScen         = 1;         !! def = 1  !! regexp = 1|3
-*' *  (1): 1 %
+*' *  (1): 1.5 %
 *' *  (3): 3 %
 *'
 parameter


main.gms
```diff
@@ -685,7 +685,7 @@ parameter
   cm_prtpScen               "pure rate of time preference standard values"
 ;
   cm_prtpScen         = 1;         !! def = 1  !! regexp = 1|3
-*' *  (1): 1 %
```
LaviniaBaumstark, line 688: I am confuesed...isn't it 1 in the line above?

dbantje: That had me confused too, but I think it is supposed to be just an index for the switch options. Because later it clearly says
```
*** define pm_prtp according to cm_prtpScen:
if(cm_prtpScen eq 1, pm_prtp(regi) = 0.015);
if(cm_prtpScen eq 3, pm_prtp(regi) = 0.03);
```
But I guess we could also just make this switch into a numerical one, so the switch value actually corresponds to the value of the pure rate of time preference. That would also a nice development feature if one wants to vary the rate.

LaviniaBaumstark: ah, thanks. Let's keep it as is

LaviniaBaumstark (approved):

PR #1762, opened 2024-08-02
Penalty for secondary energy share deviation in sectors

Renato-Rodrigues:
## Purpose of this PR

- Add support for penalization methods to incentivize the model to maintain similar secondary energy shares across sectors.
- Addresses issue [#322](https://github.com/remindmodel/development_issues/issues/322)

## Features:

- Implemented support for three penalization methods that work independently of sector-specific formulations regarding fuel use:
  - Square share penalty
  - Square deviation from average share penalty
  - Min-max deviation from average share penalty

Detailed documentation can be found within the code:
```
*** cm_seFeSectorShareDevMethod "Switch to enable an optimization incentive for sectors to have similar shares of secondary energy fuels and determine the method used for the incentive." 
*** Possible values: off or the method name (sqSectorShare, sqSectorAvrgShare, or minMaxAvrgShare)
***  off               "The model can freely allocate bio/syn/fossil fuels between sectors. If not off, a penalization term is added so sectors are incentivized to apply similar shares of bio-fuels, synfuels, and fossils in each sector."
***  sqSectorShare     "Square share penalty"
***  sqSectorAvrgShare "Square deviation from average share penalty"
***  minMaxAvrgShare   "Min-max deviation from average share penalty"
*** The relative effect of the penalization term in the objective function is scaled to avoid affecting optimization results. This scaling factor can be defined using the switch c_seFeSectorShareDevScale.
$setglobal cm_seFeSectorShareDevMethod  sqSectorAvrgShare !! def = sqSectorAvrgShare  !! regexp = off|sqSectorShare|sqSectorAvrgShare|minMaxAvrgShare
*** c_seFeSectorShareDevUnit "Defines if the penalization term is applied over fuel shares or energy units." 
***  share,  "The square penalization is applied directly to the share values. This results in different-sized regions having varying relative penalization incentives, but the range of penalization values will be more consistent from the solver's perspective."
***  energy, "The square penalization is applied to the share values multiplied by the energy demand. This approach scales penalizations better across different-sized regions, but there is a higher risk of the penalizations being ignored and the shares not being enforced if the value range is too small."
$setglobal c_seFeSectorShareDevUnit  share !! def = share  !! regexp = share|energy

parameter
  c_seFeSectorShareDevScale "scale factor in the objective function of the penalization to incentive sectors to have similar shares of secondary energy fuels."
;
  c_seFeSectorShareDevScale = 1e-3;  !! def = 1e-3
*'
```

- The default option is set to `sqSectorAvrgShare`, as it provides the most stable solution so far with acceptable solving time.

- The `sqSectorShare` and `sqSectorAvrgShare` deviations can be defined using either share units or energy units by setting the option with the switch `c_seFeSectorShareDevUnit`. The default is set to `share`.

- The penalization term in the objective function is scaled to avoid affecting optimization results. This scaling factor can be defined using the switch `c_seFeSectorShareDevScale`.

## Additional Changes:

- Replaced `vm_demFeSector` with `vm_demFeSector_afterTax` in the core share equations, as the marginal effect of the constraints on these shares should not be reflected in final energy marginal prices reported by REMIND.
- Disabled `cm_INCONV_PENALTY_FESwitch` to avoid competitive behavior between both penalization methods

## Type of change

- [x] New feature 
- [x] Bug fix 
- [x] Refactoring
- [x] Fundamental change

## Checklist:

* Test runs are here: 
`/p/projects/ecemf/REMIND/2040_scenarios/v09_2024_08_02_REMIND_3.3.1/output`
`/p/projects/ecemf/REMIND/2040_scenarios/v11_2024_09_02`


JakobBD:
Looks good to me. This may be a problem in process-based chemicals. But it seems to be important to the overall model, and I now know that it's there, so we can either live with it, or find a solution for this implementation and chemicals-feedstocks-related SE differences to live alongside each other. Maybe I'll approach you for that in the coming months. 

On an unrelated note:
From the documentation, I'd guess that 
$\text{sqSectorShare}=\sum_{i=1}^n s_i^2\quad$ with the shares $s_i$ and $\quad\sum_{i=1}^n s_i=1,\quad$ and
$\text{sqSectorAvrgShare}=\sum_{i=1}^n (s_i-\bar{s_i})^2=\sum_{i=1}^n (s_i-\frac{1}{n})^2=\sum_{i=1}^n (s_i^2-\frac{2}{n}s_i+\frac{1}{n^2})=\text{sqSectorShare}-\frac{2}{n}+\frac{1}{n}$, 
so they'd only differ by a constant, which shouldn't make a difference to the solver.

Is that so, or did I get the documentation wrong, or did I mess up in my calculations?

core/equations.gms
```diff
@@ -1054,32 +1054,52 @@ q_shBioTrans(t,regi)..
...
-        vm_demFeSector(t,regi,entySe,entyFe2,sector,emiMkt)))
+        vm_demFeSector_afterTax(t,regi,entySe,entyFe2,sector,emiMkt)))
   =e=
   sum(emiMkt$sector2emiMkt(sector,emiMkt),
       sum(se2fe(entySe,entyFe,te),
-        vm_demFeSector(t,regi,entySe,entyFe,sector,emiMkt)))
+        vm_demFeSector_afterTax(t,regi,entySe,entyFe,sector,emiMkt)))
+;
+
+q_shSeFe(t,regi,entySe)$(entySeBio(entySe) OR entySeSyn(entySe) OR entySeFos(entySe)).. !!secondary energy carriers shares in final energy demand calculated only for entySeBio, entySeSyn and entySeFos
```
robertpietzcker, line 1071: I would add an example to reduce the chance of confusion, so maybe write 

"!!secondary energy carriers shares in final energy demand (eg 'the share of bio-based FE liquids in all FE liquids'), calculated only for entySeBio, entySeSyn and entySeFos,

core/equations.gms
```diff
@@ -1054,32 +1054,52 @@ q_shBioTrans(t,regi)..
...
+  v_shSeFe(t,regi,entySe) 
+  * sum(seAgg$seAgg2se(seAgg,entySe), !!controlling the list of secondary energy carriers to sum up
```
robertpietzcker, line 1073: I would change the comment to 
"!! summing over the bio/fos/syn variants of the chosen SE carrier"

or, even more explicit

"!! summing over the bio/fos/syn variants of the chosen SE carrier (if entySe is seliqbio, summing over seliqbio, seliqfos, seliqsyn)"

robertpietzcker: sorry, on second reading, I would switch the next two lines, so that the seAgg sums are right underneath each other. 

so I would write as follows: 
* sum((sector,emiMkt)$sector2emiMkt(sector,emiMkt),
    sum(seAgg$seAgg2se(seAgg,entySe), !! determining the aggregate SE carrier type (liquids, gases, ...) 
        sum(entySe2$seAgg2se(seAgg,entySe2), !! summing over the bio/fos/syn variants of the chosen SE carrier"

core/declarations.gms
```diff
@@ -424,6 +426,8 @@ v_shGreenH2(ttot,all_regi)   "share of green hydrogen in all hydrogen by 2030 [0
 v_shBioTrans(ttot,all_regi)    "Share of biofuels in transport liquids from 2025 onwards. Value between 0 and 1."
 
 v_shfe(ttot,all_regi,all_enty,emi_sectors)           "share of final energy in sector total final energy [0..1]"
+v_shSeFe(ttot,all_regi,all_enty)                      "share of secondary energy on total secondary energy type (liquids, solids and gases) [0..1]"
```
robertpietzcker, line 429: I would add an example to reduce the chance of confusion, so maybe write

"share of energy carrier subtype in final energy demand of the aggregated carrier type (eg 'the share of bio-based FE liquids in all FE liquids')"

core/declarations.gms
```diff
@@ -424,6 +426,8 @@ v_shGreenH2(ttot,all_regi)   "share of green hydrogen in all hydrogen by 2030 [0
...
+v_shSeFeSector(ttot,all_regi,all_enty,all_enty,emi_sectors,all_emiMkt)    "share of secondary energy on secondary energy type (liquids, solids and gases) per final energy, sector and emission market [0..1]"
```
robertpietzcker, line 430: I would add an example to reduce the chance of confusion, so maybe write

"share of energy carrier subtype in final energy demand of the aggregated carrier type per sector/emiMarket combination (eg 'the share of bio-based FE liquids in all FE liquids used in ETS-covered transport')"

core/equations.gms
```diff
@@ -1129,4 +1149,92 @@ q_shbiofe_lo(t,regi,entyFe,sector,emiMkt)$(pm_secBioShare(t,regi,entyFe,sector)
...
+  * (1$sameas("%c_seFeSectorShareDevUnit%","share") + ( vm_demFeSector_afterTax(t,regi,entySe,entyFe,sector,emiMkt) )$(sameas("%c_seFeSectorShareDevUnit%","energy")) ) !!define deviation in share or energy units 
+;
+$elseIf.seFeSectorShareDev "%cm_seFeSectorShareDevMethod%" == "sqSectorAvrgShare"
+q_penSeFeSectorShareDev(t,regi,entySe,entyFe,sector,emiMkt)$(
+    ( sefe(entySe,entyFe) AND entyFe2Sector(entyFe,sector) AND sector2emiMkt(sector,emiMkt) ) AND !!only create the equation for valid cobinations of entySe, entyFe, sector and emiMkt
+    ( (entySeBio(entySe) OR entySeSyn(entySe)) ) AND !!share incentives only need to be applied to n-1 secondary energy carriers
+    (    (NOT(sameas(sector,"build") AND (t.val le 2020))) 
+      OR (NOT(sameas(sector,"indst") AND (t.val le 2020)))
+      OR (NOT(sameas(sector,"cdr") AND (t.val le 2020)))
+    ) AND !!disable historical year share incentives to buildings, industry and CDR
```
robertpietzcker, line 1178: isn't it the opposite?  so it is not that we disable the historical year shares, but rather 
"!! disable incentives as shares are anyway fixed to historical values in equation/parameter XXX"?

robertpietzcker, line 1178: isn't it the opposite?  so it is not that we disable the historical year shares, but rather 
"!! disable incentive equation until 2020, as shares are anyway fixed to historical values in equation/parameter XXX"?

robertpietzcker:
> Is that so, or did I get the documentation wrong, or did I mess up in my calculations?

@JakobBD I am not sure I understand your calculation. 

From briefly looking at it, why should 
$\text{sqSectorAvrgShare}=\sum_{i=1}^n (s_i-\bar{s_i})^2=\sum_{i=1}^n (s_i-\frac{1}{n})^2$ 
hold? 

Anyway, I think here is a numeric example that shows that the two differ by more than a fix offset: 

if the deviation of the sector-specific bioliq share from the overall average bioliq share is 0, and similarly for fos, and syn, then the squared differences from the sector values to the average values will always be 0.

however, the sum of the squared shares will be very different if you are in a situation with all bio/fos/syn-shares = 0.33  (where it would be number_sectors * 3 * 0.33^2, so number_sectors * 0.326) and a situation where bio and fos shares are 0 and syn-share is 1, where it would be number_sectors * 1^2 = number_sectors

scripts/output/single/reporting.R
```diff
@@ -76,7 +76,7 @@ edgetOutputDir <- file.path(outputdir, "EDGE-T")
 if(file.exists(edgetOutputDir)) {
   if (! file.exists(file.path(edgetOutputDir, "4_Output", "vehSalesAndModeShares.RDS"))) {
     message("EDGE-T reporting files are missing, probably because the run was killed.")
-    message("Rerunning toolIterativeEDGETransport().")
+    message("Rerunning iterativeEdgeTransport().")
```
robertpietzcker, line 79: is this change on purpose?

core/declarations.gms
```diff
@@ -548,6 +562,8 @@ q_shGreenH2(ttot,all_regi)  "share of green hydrogen in all hydrogen"
 q_shBioTrans(ttot,all_regi)  "Define the share of biofuels in transport liquids from 2025 on."
 
 q_shfe(ttot,all_regi,all_enty,emi_sectors)            "share of final energy carrier in the sector final energy"
+q_shSeFe(ttot,all_regi,all_enty)                      "share of secondary energy on total secondary energy type (liquids, solids and gases)"
```
robertpietzcker, line 565: I would change to 
"share of energy carrier subtype in final energy demand of the aggregated carrier type (eg 'the share of bio-based FE liquids in all FE liquids')" 
or something similar

core/declarations.gms
```diff
@@ -548,6 +562,8 @@ q_shGreenH2(ttot,all_regi)  "share of green hydrogen in all hydrogen"
...
+q_shSeFeSector(ttot,all_regi,all_enty,all_enty,emi_sectors,all_emiMkt) "share of secondary energy on secondary energy type (liquids, solids and gases) per final energy, sector and emission market"
```
robertpietzcker, line 566: I would change to 
"share of energy carrier subtype in final energy demand of the aggregated carrier type for each sector/emiMarket combination (eg 'the share of bio-based FE liquids in all FE liquids within ETS-covered transport'"

core/declarations.gms
```diff
@@ -157,6 +157,8 @@ pm_shFeCes(ttot,all_regi,all_enty,all_in,all_teEs)   "Final energy shares for CE
 
 pm_shfe_up(ttot,all_regi,all_enty,emi_sectors)       "Final energy shares exogenous upper bounds per sector"
 pm_shfe_lo(ttot,all_regi,all_enty,emi_sectors)       "Final energy shares exogenous lower bounds per sector"
+p_shSeFe(ttot,all_regi,all_enty)                     "initial share of secondary energy on total secondary energy type (liquids, solids and gases) [0..1]"
+p_shSeFeSector(ttot,all_regi,all_enty,all_enty,emi_sectors,all_emiMkt) "initial share of secondary energy on secondary energy type (liquids, solids and gases) per final energy, sector and emission market [0..1]"
```
robertpietzcker, line 161: I would change to 
"Initial share of energy carrier subtype in final energy demand of the aggregated carrier type for each sector/emiMarket combination (eg 'the share of bio-based FE liquids in all FE liquids within ETS-covered transport')"

core/declarations.gms
robertpietzcker, line 429: I would change to 
"share of energy carrier subtype in final energy demand of the aggregated carrier type for each sector/emiMarket combination (eg 'the share of bio-based FE liquids in all FE liquids')"

core/declarations.gms
robertpietzcker, line 430: I would change to 
"Initial share of energy carrier subtype in final energy demand of the aggregated carrier type for each sector/emiMarket combination (eg 'the share of bio-based FE liquids in all FE liquids within ETS-covered transport')"

core/preloop.gms
```diff
@@ -189,4 +189,35 @@ if (cm_startyear gt 2005,
 Execute_Loadpoint 'input_ref' vm_capEarlyReti.l = vm_capEarlyReti.l;
 );
 
+*** initialize secondary energy shares in sectors if deviations are penalized, i.e., if cm_seFeSectorShareDevMethod is enabled
```
robertpietzcker, line 192: I would change the explanation: 
*** initialize the carrier subtype shares in final energy demand such that the starting point for the model is "all sectors have the same bio/fos/syn shares for a given carrier type" when cm_seFeSectorShareDevMethod is enabled.

robertpietzcker:
> The penalty sqSectorShare not only penalizes unequal distribution over sectors, but also unequal overall shares of SE types.

yes, which is one of the reasons I prefer the (currently default) "sqSectorAvrgShare " :-)

but you are right, some info about this should be added in the description of "sqSectorShare" - like "this method ALSO creates an unwanted incentive for the model to have equal fos/syn/bio shares"

robertpietzcker (approved):

scripts/output/single/reporting.R
Renato-Rodrigues: it was a merge overwrite that did not affect the runs. I reverted it back to avoid the change.

JakobBD:
> which is one of the reasons I prefer the (currently default) "sqSectorAvrgShare " :-)

To me, that seems like the only difference to `sqSectorAvrgShare`. 

I.e.: `sqSectorShare` can be shown to be `sqSectorAverageShare` plus a penalty on unequal overall SE shares. 

You're saying this only difference is unwanted (and I agree). 

Which begs the questions: Is there any advantage to it? If not, then I'd be in favour of deleting it. 

(As always, if I'm not messing up). 
 
Calculations: 

$\bar{s_i} = \frac{1}{n_j} \sum_j s_{ij}$

and then

$\sum_{i,j} (s_{i,j}-\bar{s_i})^2$ 

$= \sum_{i,j} (s_{i,j}^2- 2 s_{i,j} \bar{s_i} + \bar{s_i}^2) $

$= \sum_{i,j} s_{i,j}^2 - 2 \sum_{i} \bar{s_i} \sum_{j} s_{i,j}  + \sum_{i,j} \bar{s_i}^2$

$= \sum_{i,j} s_{i,j}^2 - 2 n_j \sum_{i} \bar{s_i}^2  + n_j \sum_{i} \bar{s_i}^2 $

$= \sum_{i,j} s_{i,j}^2 - n_j \sum_{j} \bar{s_i}$

Renato-Rodrigues:
> some info about this should be added in the description of "sqSectorShare"

https://github.com/remindmodel/remind/pull/1762/commits/39c264187f1cf120fd9c340c842de1843f20a8ab

Renato-Rodrigues:
The alternative formulations are present to compare different levels of complexity and allow runtime comparisons.

The default option selected considers the pros and cons observed so far about these alternatives, but as these formulations should be further improved in subsequent PRs, I prefer to keep the current options in the code.

Renato-Rodrigues:
@robertpietzcker 
I disabled the shares incentives in this PR so we can merge this code quickly without causing any changes to REMIND.
This will allow me to do quicker tests with the latest REMIND versions in a more effective way.

Please let me know if you are ok with merging in this way so I can go ahead with the merge.

robertpietzcker:
to make sure I understand correctly - the merged version would have as default that the model sees NONE of the penalizing terms, but just the share calculation terms?
If yes, then I would say please go ahead with the merge - but just as a safeguard, please also set up a quick before/after comparison cascade in SSP2 so we are sure that the share-calculation doesn't have an unwanted consequence that we didn't think of :-)

Renato-Rodrigues:
> please also set up a quick before/after comparison cascade in SSP2 so we are sure that the share-calculation doesn't have an unwanted consequence that we didn't think of :-)

Here is a comparison scenario of the NPi runs (1) without any code from the PR and (2) with the code from the PR but with `cm_seFeSectorShareDevMethod` disabled (scenarios with the `PR` suffix in their names) .

`/p/projects/ecemf/REMIND/2040_scenarios/_mergeTest/remind_seFeSectorSharePenalization/compScen-NPI-2024-10-22_15.48.18-*.pdf`

There is juts one small difference that is active between the runs. I initialized the .l share values for `p_shSeFe`, `v_shSeFe`, `p_shSeFeSector` and `v_shSeFeSector` to use the total share of bio/syn/fossil carriers to see if the initial point alone would make much difference to the results.

The results are almost the same, but there are some differences on the biomass trade and the initial point change also affects bunker emissions. Besides that I think most of the other changes are either non-existent or residual.

The policy runs aren't finished yet.

Renato-Rodrigues:
@robertpietzcker

Since both the SSP2 scenarios and the 2040 paper scenarios are using the latest setup we tested, do I have the green light to proceed with merging this into the trunk with our current test configuration?

The current tested setup uses `cm_seFeSectorShareDevMethod = sqSectorAvrgShare` for all runs and `cm_limitSolidsFossilRegi = EUR_regi` for all non-calibration scenarios, including `NPi` and policy scenarios.

If you need any additional test runs, please let me know.

If approved, I'll update this PR to the current tested configuration and create a new PR to incorporate the `cm_limitSolidsFossilRegi` implementation.

The 2040 paper test runs are available at:  

johannah-pik:
Hi all :)
Its a bit hard to follow your discussions quickly.
Can someone summarize the implications of this fix for me?
In the transport reporting we read vm_demFeSector to get the fossil/syn/bio shares and apply a fix I discussed with @fschreyer to harmonize the shares for fepet and fedie as REMIND does not follow a reasonable mechanism to allocate them.

Renato-Rodrigues:
Without the code in this PR, REMIND was free to allocate bio, syn, and fossil fuels across sectors.
If there is no additional sector specific cost to use a given carrier type of bio, syn or fossil origin, the model simply could choose randomly where it places more biofuels for example, making a sector emit more or less by pure chance.

To guide the model towards a more balanced distribution, the PR introduces a small monetary incentive (in the form of a cost penalization) to equalize the fuel shares across sectors. This incentive is designed to be minimal, so if there’s a sector-specific economic advantage to using a particular fuel type, the incentive should remain negligible. 
In the absence of any other specific cost differences the shares of biofuels, synfuels, and fossil fuels in each sector should roughly match their overall share in the economy.

This adjustment takes effect from 2025 onward. 
Post-merge, fedie and fepet should reflect bio, syn, and fossil fuel shares close to their overall proportions in secondary energy for these years. 
If you intend to control these fuel use shares in transport based on some dynamics from edge-t, you would need to also include your logic at the REMIND equation side. Otherwise, if you simply use a equal share split between these carriers fuel origin types, you don't need anymore to do any post-processing as the split will be done endogenously by the model.

Renato-Rodrigues:
the shares on years before 2025 shoudl be controlled by the `q_shbiofe_up` and `q_shbiofe_lo` equations.

johannah-pik:
Thanks @Renato-Rodrigues.
this is very much appreciated. Currently, EDGE-T only sees liquids and a corresponding price for liquids. 
Therefore, the share of fossil, syn- and biofuels is not influenced by the decision-making process in EDGE-T itself and remains on the REMIND side.

I will check once the fix is in place and remove the postprocessing step in our reporting.

Renato-Rodrigues:
This PR is merging the changes discussed here in a disabled state.
Nothing will change from the REMIND results perspective.

A follow up PR will be created to include the default behavior change in REMIND, address the issue that motivated these code changes and point to test runs.

PR #1763, opened 2024-08-03
Enable bioigcc early retirement

Renato-Rodrigues:
## Purpose of this PR

- Enable `bioigcc` early retirement as discussed in ARIADNE scenarios.

- Comments from Felix on the subject:
```
I have always wondered a bit why we do not allow for early retirement of biomass technologies as there could be benefits of shifting biomass supply to other sectors / technologies. 
I had tested this already, but this time it actually makes a difference as the model shifts biomass out of power plants without carbon capture into either bioh2 or bioliquids plants with carbon capture. 
This higher availability of captured biogenic CO2 (with the same biomass feedstock) makes DAC superfluous and decreases carbon prices of a near fossil-free 2050 system notably to ~650. 
As we do not have retrofits of current biomass power plants for carbon capture, we might underestimate this flexibility. 
Also REMIND builds some of these plants quite recently and in the real world they might be older and reach their end of lifetime earlier. 
At the same time, I vaguely remember that @amerfort made some assumption in her CDR potential study on Germany that not all biomass will be available for carbon capture plants. 
So, I am raising the question of whether we want to allow for flexibility for shifting biomass. 
```

## Type of change

## Checklist:

* Test runs are here: 
`/p/projects/ecemf/REMIND/2040_scenarios/v09_2024_08_02 `

### DIFF
--- a/core/bounds.gms
+++ b/core/bounds.gms
@@ -348,6 +348,8 @@ vm_capEarlyReti.up(ttot,regi,te) = 1e-6;
 vm_capEarlyReti.up(ttot,regi,te)$(teFosNoCCS(te)) = 1;
 *** FS: allow nuclear early retirement (for nucscen 7)
 vm_capEarlyReti.up(ttot,regi,"tnrs") = 1;
+*** allow early retirement of biomass used in electricity
+vm_capEarlyReti.up(ttot,regi,"bioigcc") = 1;
 
 ***restrict early retirement to the modeling time frame (to reduce runtime, the early retirement equations are phased out after 2110)
 vm_capEarlyReti.up(ttot,regi,te)$(ttot.val lt 2009 or ttot.val gt 2111) = 0;
--- a/core/datainput.gms
+++ b/core/datainput.gms
@@ -758,7 +758,7 @@ pm_regiEarlyRetiRate(t,regi,"biochp")  = 0.5 * pm_regiEarlyRetiRate(t,regi,"bioc
 pm_regiEarlyRetiRate(t,regi,"gashp")   = 0.5 * pm_regiEarlyRetiRate(t,regi,"gashp") ; !! chp should only be phased out slowly, as district heating networks/ industry uses are designed to a specific heat input
 pm_regiEarlyRetiRate(t,regi,"coalhp")  = 0.5 * pm_regiEarlyRetiRate(t,regi,"coalhp"); !! chp should only be phased out slowly, as district heating networks/ industry uses are designed to a specific heat input
 pm_regiEarlyRetiRate(t,regi,"biohp")   = 0.5 * pm_regiEarlyRetiRate(t,regi,"biohp") ; !! chp should only be phased out slowly, as district heating networks/ industry uses are designed to a specific heat input
-
+pm_regiEarlyRetiRate(t,regi,"bioigcc")   = 0.25 * pm_regiEarlyRetiRate(t,regi,"bioigcc") ; !! reduce bio early retirement rate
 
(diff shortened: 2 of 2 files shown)


Renato-Rodrigues:
@fschreyer
@robertpietzcker

robertpietzcker (approved):
sounds good

fschreyer:
Can I ask why only ``bioigcc`` and not all other bioTe technologies as well like ``biotrmod``, ``biochp``, ``biohp``?

Renato-Rodrigues:
> Can I ask why only `bioigcc` and not all other bioTe technologies as well like `biotrmod`, `biochp`, `biohp`? Captured biogenic CO2 becomes relatively valuable in 2050 net-zero scenarios and these technologies can partially not capture so it can make quite a difference.

In my net-zero runs with Germany (2045) and EU (2050) net-neutrality targets:

- **`biotrmod`**: After the enforced historical capacities this technology is not installed, i.e. no new installations after 2010-2015. With a 35-year lifetime, it phases out before the ambitious targets of 2050, with most of it gone by 2045. If we consider early retirement for this technology, we'll need to realistically limit the early retirement rate. The model would otherwise try to phase it out as quickly as possible, which could be unrealistic in a real-world scenario if we're too lenient.

- **`biohp`**: under extremely limited biomass scenarios (e.g. EU27 primary energy biomass limited to 4EJ by 2050), indeed the model wants to get this tech early retired. If biomass is not extremely limited, the model wants to install meaningful capacities of this tech even after 2035.

- **`biochp`**: The behavior of `biochp` is very similar to `bioigcc`.

In summary, 
- I would be 100% ok with extending the early retirement to `biohp` and `biochp`.
- I would only do the same to `biotrmod` if somebody has the time to test and set the early retirement rate to reasonable numbers to it to avoid a too unrealistic sudden drop. The model will want to get ride of `biotrmod` the fast as it can in any scenario, and not only the limited biomass cases.

fschreyer:
I think it would make sense to allow early retirement for all bio-technologies at a low rate of, for instance, 0.05%/yr. In the context of 2050 net-zero targets, that would mean that it is slow but still everything can be retired until 2050. However, we will probably also start to explore scenarios that start in 2030 (or delayed action) so the question of these parameters will become even more relevant. From a systems perspective, it becomes important for ambitious 2050 targets, as the model will try to capture as much biogenic CO2 as possible and would not want to "waste" biomass somewhere where it cannot be captured. But I can also imagine cases where we would like to limit that. ``biotrmod``, for instance, includes buildings heating which is something which people do not want to / cannot phase-out quickly even if prices rise a lot. Similarly, biochp plants might want to be kept due to local preferences even if uneconomical from a system level. Overall, I think we should open an issue to discuss this with more people and bring it up in a REMIND validation meeting to agree what is the default in REMIND. Apart from that, we can still do scenario/project-specific assumptions.

robertpietzcker:
What about a) opening an issue to put it on the discussion part of the validation meetings, and b) still implementing the changes now/merging the PR? 
I would say the proposal (allow bioigcc/biochp/biohp to be retired, albeit slowly) is better than the current setting ( I don't have a clear opinion on biotrmod), and we still may change the default after the discussion, but until then we are sure that the scenarios use the "somewhat better version")

Renato-Rodrigues:
@robertpietzcker I created an issue as suggested above: 
https://github.com/remindmodel/development_issues/issues/341

If you could please revise [PR 1787
](https://github.com/remindmodel/remind/pull/1787). The PR enables the early retirement for biochp and biohp as the suggested initial step.

PR #1819, opened 2024-09-10
harmonise options format for output.R script

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
- lucode2 0.47.11 (echt Köllnisch Wasser) accepts both option=value and --option=value formats, so does output.R
- this enables bash completion for all these options
- bumped help up to start without loading pointless packages

### DIFF
--- a/output.R
+++ b/output.R
@@ -26,31 +26,50 @@ helpText <- "
 #'
 #' [options] can be the following flags:
 #'
-#'   --help, -h:      show this help text and exit
-#'   --test, -t:      tests output.R without actually starting any run
-#'   --renv=<path>    load the renv located at <path>, incompatible with --update
-#'   --update         update packages in renv first, incompatible with --renv=<path>
+#'   --help, -h   show this help text and exit
+#'   --test, -t   tests output.R without actually starting any run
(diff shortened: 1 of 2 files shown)


output.R
```diff
@@ -26,32 +26,51 @@ helpText <- "
...
-#'   output=           output=compareScenarios2 directly selects the specific script
-#'   outputdir=        Can be used to specify the output directories to be used.
-#'                     Example: outputdir=./output/SSP2-Base-rem-1,./output/NDC-rem-1
-#'   remind_dir=       path to remind or output folder(s) where runs can be found.
-#'                     Defaults to ./output but can also be used to specify multiple
-#'                     folders, comma-separated, such as remind_dir=.,../otherremind
-#'   slurmConfig=      use slurmConfig=priority, short or standby to specify slurm
-#'                     selection. You may also pass complicated arguments such as
-#'                     slurmConfig='--qos=priority --mem=8000'
+#'   --comp=              comp=single means output for single runs
```
orichters, line 36: should we either state that `comp=single` is also supported (maybe in line 34) or adjust the examples as well?

orichters (approved):

PR #1831, opened 2024-09-13
CCS bounds update

pweigmann:
## Purpose of this PR

Update to the input-data for CCS bounds. The processing of the IEA CCUS 2024 data is now harmonized between input data and reference/validation data. The resulting changes to the bounds are minor. There is still flexibility to adjust the bounds from within REMIND via the switch ``c_fracRealfromAnnouncedCCScap2030``.

This PR is merged together with https://github.com/pik-piam/mrremind/pull/531.

## Type of change

## Checklist:

* Test runs are here: ``/p/tmp/pascalwe/remind/output/SSP2-EU21-NPi_2024-09-13_13.40.21``


amerfort (approved):
Thanks for taking care of this! I approve.
As a remark: REMIND default near-term CCS assumptions now align with "max_yel" values calculated for the validation tool. Calculation of near term CCS limits is now more transparent and can be understood with REMIND code alone. 
But heads up: if assumptions change in the validation logic, they would have to be changed in the REMIND code as well, if they should continue to align with each other.

PR #1835, opened 2024-09-19
Fix REMIND for running with no industry CCS

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
fix remindmodel/development_issues#349

### DIFF
--- a/core/presolve.gms
+++ b/core/presolve.gms
@@ -170,6 +170,7 @@ vm_macBase.fx(ttot,regi,"ch4wstl")$(ttot.val ge 2005) = p_emineg_econometric(reg
 vm_macBase.fx(ttot,regi,"n2owaste")$(ttot.val ge 2005) = p_emineg_econometric(regi,"n2owaste","p1") * pm_pop(ttot,regi) * (1000*pm_gdp(ttot,regi) / (pm_pop(ttot,regi)*pm_shPPPMER(regi)))**p_emineg_econometric(regi,"n2owaste","p2");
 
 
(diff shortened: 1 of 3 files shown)


JakobBD (approved):
Looks good. 
[`q37_limitOutflowCCPrc`](https://github.com/remindmodel/remind/pull/1835/files#diff-f2b7ce9b56bd63ff7c3e6805c6fc79f71d7a2900df47baeef0aae1a868f34775R453) could also be adapted. I tried myself, but getting domain violation errors. Still a GAMS syntax newbie, it seems.

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
> [`q37_limitOutflowCCPrc`](https://github.com/remindmodel/remind/pull/1835/files#diff-f2b7ce9b56bd63ff7c3e6805c6fc79f71d7a2900df47baeef0aae1a868f34775R453) could also be adapted.


Adapted how?  I do not see the need.  Still a PBS newbie, it seems ;)

JakobBD:
No need in terms of functionality, just code quality. 

The dollar condition on the equation domain served to define the equation only for actual CC tech, by checking that the tech appeared as a CC tech in `tePrc2teCCPrc`. Now that we have `teCCPrc` explicitly, it shouldn't be needed anymore. 
However, when I just delete the condition, I get domain violations in the sums below, which I don't understand the reason for.

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
Never change a running system :p

PR #1873, opened 2024-11-01
add support for switch to force monotonously decreasing solids fossil use per region

Renato-Rodrigues:
## Purpose of this PR

- Add switch to force monotonously decreasing solids fossil use per region to avoid zig-zag temporal behavior in the model for decarbonization scenarios

## Type of change

- [x] New feature 

## Checklist:

* Test runs are here: 

### DIFF
--- a/core/declarations.gms
+++ b/core/declarations.gms
@@ -565,6 +565,9 @@ $IFTHEN.sehe_upper not "%cm_sehe_upper%" == "off"
 q_heat_limit(ttot,all_regi)  "equation to limit maximum level of secondary energy district heating and heat pumps use"
 $ENDIF.sehe_upper
 
+$ifthen.limitSolidsFossilRegi not %cm_limitSolidsFossilRegi% == "off"
+  q_fossilSolidsLimitReg(ttot,all_regi,all_enty,all_enty,emi_sectors,all_emiMkt)  "limit solids fossil to be lower or equal to previous year values"
+$endif.limitSolidsFossilRegi
 ;
 ***----------------------------------------------------------------------------------------
 ***                                   SCALARS
--- a/core/declarations.gms
+++ b/core/declarations.gms
@@ -566,7 +566,7 @@ q_heat_limit(ttot,all_regi)  "equation to limit maximum level of secondary energ
 $ENDIF.sehe_upper
 
 $ifthen.limitSolidsFossilRegi not %cm_limitSolidsFossilRegi% == "off"
-  q_fossilSolidsLimitReg(ttot,all_regi,all_enty,all_enty,emi_sectors,all_emiMkt)  "limit solids fossil to be lower or equal to previous year values"
+  q_fossilSolidsLimitReg(ttot,all_regi,all_enty,all_enty,emi_sectors,all_emiMkt)  "limit solids fossil to be lower or equal to previous year values in each (sector x emiMkt) combination"
 $endif.limitSolidsFossilRegi
 ;
 ***----------------------------------------------------------------------------------------
--- a/core/equations.gms
+++ b/core/equations.gms
@@ -1133,7 +1133,7 @@ q_shbiofe_lo(t,regi,entyFe,sector,emiMkt)$(pm_secBioShare(t,regi,entyFe,sector) 
 $ifthen.limitSolidsFossilRegi not %cm_limitSolidsFossilRegi% == "off"
 q_fossilSolidsLimitReg(ttot,regi,entySe,entyFe,sector,emiMkt)$(limitSolidsFossilRegi(regi) and (ttot.val ge max(2020, cm_startyear)) AND sefe(entySe,entyFe) AND sector2emiMkt(sector,emiMkt) AND (sameas(sector,"indst") OR sameas(sector,"build")) AND sameas(entySe,"sesofos"))..
   vm_demFeSector_afterTax(ttot,regi,entySe,entyFe,sector,emiMkt)
-  =l=
+  =le=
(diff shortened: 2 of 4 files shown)


core/declarations.gms
```diff
@@ -565,6 +565,9 @@ $IFTHEN.sehe_upper not "%cm_sehe_upper%" == "off"
 q_heat_limit(ttot,all_regi)  "equation to limit maximum level of secondary energy district heating and heat pumps use"
 $ENDIF.sehe_upper
 
+$ifthen.limitSolidsFossilRegi not %cm_limitSolidsFossilRegi% == "off"
+  q_fossilSolidsLimitReg(ttot,all_regi,all_enty,all_enty,emi_sectors,all_emiMkt)  "limit solids fossil to be lower or equal to previous year values"
```
robertpietzcker, line 569: I would make the description more concrete: "limit solids fossil to be lower or equal to previous year values in each (sector x emiMkt) combination "

core/equations.gms
```diff
@@ -1127,4 +1127,14 @@ q_shbiofe_lo(t,regi,entyFe,sector,emiMkt)$(pm_secBioShare(t,regi,entyFe,sector)
   sum((entySeBio,te)$se2fe(entySeBio,entyFe,te), vm_demFeSector_afterTax(t,regi,entySeBio,entyFe,sector,emiMkt))
 ;
 
+***---------------------------------------------------------------------------
+*' Limit solids fossil to be lower or equal to previous year values  
+***---------------------------------------------------------------------------
+$ifthen.limitSolidsFossilRegi not %cm_limitSolidsFossilRegi% == "off"
+q_fossilSolidsLimitReg(ttot,regi,entySe,entyFe,sector,emiMkt)$(limitSolidsFossilRegi(regi) and (ttot.val ge max(2020, cm_startyear)) AND sefe(entySe,entyFe) AND sector2emiMkt(sector,emiMkt) AND (sameas(sector,"indst") OR sameas(sector,"build")) AND sameas(entySe,"sesofos"))..
+  vm_demFeSector_afterTax(ttot,regi,entySe,entyFe,sector,emiMkt)
+  =l=
```
robertpietzcker, line 1136: maybe use =le= instead of =l=? 
or is it basically the same for the model?

main.gms
```diff
@@ -1762,6 +1762,9 @@ $setglobal cm_INCONV_PENALTY_FESwitch  on !! def = on  !! regexp = off|on
 *** *JH/LB* Activate MOFEX partial fossil fuel extraction cost minimization model
 *** * Warning: Use a well-converged run since the model uses vm_prodPe from the input GDX
 $setGlobal cm_MOFEX  off        !! def = off  !! regexp = off|on
+*** cm_limitSolidsFossilRegi off   !! def=off
+*** ext_regi value, e.g. "EUR_regi, USA"   "solids fossil in industry and buildings for regions within EUR_regi and USA are forced to be monotonuosly decreasing from 2020 or cm_startyear onward."
```
robertpietzcker, line 1766: "starting in max(2020, cm_startyear), fossil solids use in each (sector x emiMkt) has to decrease compared to the previous time step for each region within EUR_regi and USA"

core/equations.gms
```diff
@@ -1127,4 +1127,14 @@ q_shbiofe_lo(t,regi,entyFe,sector,emiMkt)$(pm_secBioShare(t,regi,entyFe,sector)
   sum((entySeBio,te)$se2fe(entySeBio,entyFe,te), vm_demFeSector_afterTax(t,regi,entySeBio,entyFe,sector,emiMkt))
 ;
 
+***---------------------------------------------------------------------------
+*' Limit solids fossil to be lower or equal to previous year values  
+***---------------------------------------------------------------------------
+$ifthen.limitSolidsFossilRegi not %cm_limitSolidsFossilRegi% == "off"
+q_fossilSolidsLimitReg(ttot,regi,entySe,entyFe,sector,emiMkt)$(limitSolidsFossilRegi(regi) and (ttot.val ge max(2020, cm_startyear)) AND sefe(entySe,entyFe) AND sector2emiMkt(sector,emiMkt) AND (sameas(sector,"indst") OR sameas(sector,"build")) AND sameas(entySe,"sesofos"))..
```
robertpietzcker, line 1134: this is on for ttot = 2020, so that in 2020 fossil solids use already has to go down. 
I guess we are positive that the average of 2018-2022 is already lower than the average 2013-2017 in each EU region and US historically?

main.gms
robertpietzcker, line 1766: I personally wouldn't be 100% sure how to interpret the new line starting with "ext_regi value" - maybe it would be clearer to start the line with "replace 'off' by ext_regi value, e.g. " ?

robertpietzcker (approved):
looks good

core/equations.gms
Renato-Rodrigues: I don't think it makes a difference for the solution, but indeed `le` is a better formulation. Thks!

core/equations.gms
Renato-Rodrigues: This could be an small issue for FRA, and it would be a real issue if anybody enforces this to India.

image

I will check if we can control the historical data already by some of our other equations instead.

Renato-Rodrigues: If you are ok with it, I would merge as is for now as I only enable this for EU27 in my paper runs, and I will check on enforcing historical values after I submit the paper revision.

robertpietzcker: sure, merge for now. 
just wondering: what would speak against setting the starting point to 2025? I don't think the runs ever produced a solids spike in 2020, only in 2025?

PR #1876, opened 2024-11-08
Updating bioftcrec tech parameters

Renato-Rodrigues:
## Purpose of this PR

- Small change to bioftcrec technology parameters for investment and operation costs to make them compatible with EU 2040 scenario runs.
- These values should be revised and update again if necessary when people more involved on these technology developments have the time to evaluate the results: at @strefler @nicobauer @fschreyer @robertpietzcker 

## Type of change

## Checklist:

* Test runs are here: 


fschreyer (approved):
We discussed this and this is an (incomplete) change in the direction we discussed here: https://github.com/remindmodel/development_issues/issues/335

It is not ideal, but I don't think results will not change a lot. It serves to have a version consistent with Renato's paper version. I will soon prepare the PR with the full change of X-to-Liquids/Gas parameters with a before and after comparison, which will overwrite this.

PR #1880, opened 2024-11-11
MAgPIE coupling: use defaults from MAgPIE's default.cfg for list of MAgPIE outputs

dklein-pik:
## Purpose of this PR

So far the couplig did not use the list of output scripts as defined in MAgPIE's default.cfg, instead only `rds_report` was specified explicitely overwriting the defaults. Since the list of required/useful output scripts change from time to time we switch to using the defaults from MAgPIE's default.cfg, now including `extra/disaggregation`.


laurinks (approved):
Thanks, @dklein-pik, for the quick implementation! Looks good to me. If I understand it right, this means that REMIND runs a limited reporting between iterations and full reporting after the last iteration. MAgPIE runs the full reporting after each iteration. I suppose that this does not cost too much run time, i.e. no need to reduce the MAgPIE reporting between iterations. Agree?

PR #1883, opened 2024-11-13
fix BECCS capacities in 2020 to zero

tabeado:
## Purpose of this PR
Currently, all operating CCS projects are set as bounds for 2020 and REMIND decides freely about which capture technologies to use. globally, the split is about 50:50 for BECCS and fossil CCS, leading to 20Mt BECCS in 2020 and 27 Mt in 2025. 
However, in 2020 there were 3 BECCS plants in the USA (ethanol+CC, which we don't model), which total a maximum of 1 Mt captured, and one in Japan (bioigccc with 0.18 Mt).
We thus set vm_cap for all BECCS technologies to 0 in 2020.

For policy runs, this does not make a difference. However, in NPis the head-start for BECCS makes a big difference.

## Type of change

## Checklist:

See this [development issue](https://github.com/remindmodel/development_issues/issues/402) for effects on NPi

### DIFF
--- a/core/bounds.gms
+++ b/core/bounds.gms
@@ -288,7 +288,7 @@ loop(te$(sameas(te,"ngcc") OR sameas(te,"ngt") OR sameas(te,"gaschp")),
   vm_cap.lo("2020",regi,te,"1")$pm_histCap("2020",regi,te) = 0.95 * pm_histCap("2020",regi,te);
 );
 
-*** fix capacities for BECCS technologies to  zero in 2020
+*** fix capacities for advanced bio carbon capture technologies to zero in 2020 (i.e. no BECCS in 2020)
 vm_cap.fx("2020",regi,te,rlf)$(sameas(te,"bioigccc") OR sameas(te,"bioftcrec") OR sameas(te,"bioh2c") OR sameas(te,"biogasc")) = 0;
 
 *** fix emissions to historical emissions in 2010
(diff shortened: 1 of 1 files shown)


core/bounds.gms
```diff
@@ -288,6 +288,9 @@ loop(te$(sameas(te,"ngcc") OR sameas(te,"ngt") OR sameas(te,"gaschp")),
   vm_cap.lo("2020",regi,te,"1")$pm_histCap("2020",regi,te) = 0.95 * pm_histCap("2020",regi,te);
 );
 
+*** fix capacities for BECCS technologies to  zero in 2020
```
amerfort, line 291: I would suggest to be even more explicit here, as we technically constrain only the carbon capture part:
`fix capacities for advanced bio carbon capture technologies to zero in 2020 (carbon capture part of BECCS)`
What do you think?

tabeado: fair enough, I adjusted the text

core/bounds.gms
```diff
@@ -288,6 +288,9 @@ loop(te$(sameas(te,"ngcc") OR sameas(te,"ngt") OR sameas(te,"gaschp")),
...
+vm_cap.fx("2020",regi,te,rlf)$(sameas(te,"bioigccc") OR sameas(te,"bioftcrec") OR sameas(te,"bioh2c") OR sameas(te,"biogasc")) = 0;
```
amerfort, line 292: We could avoid listing all BECCS technologies individually by using the already existing sets teBio and teCCS. Then this bound will automatically be applied also to BECCS technologies that might be added in the future. This would read then:
`vm_cap.fx("2020",regi,te,rlf)$(teBio(te) AND teCCS(te)) = 0;`

amerfort (commented):
Thank you so much for doing this!

core/bounds.gms
tabeado: thanks for figuring this out!

amerfort (approved):
Thanks Tabea!

PR #1904, opened 2024-12-02
Remove CDR from SE-FE share equations and EW upscaling bug fix

amerfort:
## Purpose of this PR
**This PR does not affect standard scenarios such as NPi or peak Budget policy runs.** It only affects runs that intentionally switch on the CDR options OAE and EW (which are off by default), which is mainly overshoot scenarios for ScenarioMip or the AMT run SSP2-EcBudg400.
Background:
The work on equal SE-FE shares across sectors seems to cause problems for the CDR sector (see also [this issue](https://github.com/remindmodel/development_issues/issues/416)). 
Enhanced Weathering (EW) and Ocean Alkalinity Enhancement (OAE) are both CDR technologies that are off by default and use `fedie` as fuel. With the new SE-FE equations, EW and OAE never show up, even though they should given the very high carbon prices. 
For now this PR removes the CDR sector from the equal SE-FE shares concept such that EW and OAE are not suppressed anymore, as the detailed sectoral distribution is not of high priority for the overshoot scenario and the relevant CDR technologies don't show up in policy runs, where equal FE shares across demand sectors are desired.

We plan to find a clean solution how to include CDR in the equal se-fe shares across all demand sectors as soon as possible, but not within this validation.

Furthermore, this PR contains a bug fix on EW upscaling, which is also necessary for it's deployment by the model.

## Type of change

- [x] Bug fix 

## Checklist:

* Test runs are here: /p/tmp/amerfort/remind_SeFeShares_noCDR/output


robertpietzcker:
ok, I can imagine the solver being to stupid to realize he has to increase ALL types of fedie liquids at the same time . 

So I think it is not great to take CDR out of the "equal shares" equations, but the total amounts should hopefully remain limited, so fine as a quick fix to get this release done.

Renato-Rodrigues:
I agree with implementing this quick fix for now, but we should plan a follow-up with further tests, such as:

- Modifying the units in the `q_shSeFeSector` equation exclusively for the CDR sector (e.g., multiplying the equation by 1^e3) to verify if CONOPT is ignoring the equation due to the small magnitude of the CDR variables.

- Running a test with a higher solver tolerance to determine if CONOPT starts factoring in the shares as they currently are.

If either of these tests solves the issue, we should consider introducing an auxiliary variable for CDR final energy demand that uses a different unit. This could help ensure the solver does not disregard CDR equations due to their relatively small scale compared to other demand sectors.

Additionally, we need to investigate the infeasible runs when forcing CDR values, as this likely points to a separate issue.

Renato-Rodrigues (approved):

amerfort:
FYI: @LaviniaBaumstark and maybe @laurinks 
`make test` raises an error:

> Failure (test_01-readCheckScenarioConfig.R:14:5): perform readCheckScenarioConfig with scenario_config_SSP3.csv
Expected `readCheckScenarioConfig(csvfile, remindPath = "../../", testmode = TRUE)` to run without any warnings.
ℹ Actually got a <simpleWarning> with text:
  For module carbonprice, the undefined realizations diffLin2Lin are used by these scenarios: SSP3-PkBudg1000

laurinks:
No need to worry about this failure, @amerfort ! Nico's scenario_config_SSP3.csv is outdated, and will be removed with [this PR](https://github.com/remindmodel/remind/pull/1906). Sorry for the inconvenience in the meantime.

PR #1907, opened 2024-12-03
MAgPIE coupling: update coupled config to new scenarios

dklein-pik:
## Purpose of this PR

Correctly configure coupled runs

- using the correct scenario settings for SSP1 in MAgPIE
- adding SSP3
- removing SDP scenarios
- adding SSP2_lowEn
- referencing updated Pkbudgets


laurinks:
Thanks, @dklein-pik! As the new standard in the `scenario_config.csv` is NPi2025 (particularly, for SSP1, SSP3 and SSP5, we only run NPi2025, not the old NPi), I would suggest updating all scenario names in the coupled config to SSPx-NPi2025 instead of SSPx-NPi. The new MAgPIE release coming later this week will also include an update of NPi policies, thus be consistent with this setting.

laurinks (approved):
Looks good, thanks!

PR #1917, opened 2024-12-06
Up-date to include changes of techno-economic assumptions in SSP3 and…

nicobauer:
- [x ] Minor change (default scenarios show only small differences)


laurinks (approved):
Thanks, Nico!

PR #1929, opened 2024-12-12
rename and improve description of weathering rates

tabeado:
## Purpose of this PR
p33_co2_rem_rate and s33_co2_rem_rate currently have the same description, albeit meaning different things. 

To make it more accessible and have different names, I suggest to rename the latter to s33_co2_rem_fraction, as it actually provides the (theoretical) fraction weathering each year. Alternatively, we could call it p33_co2_rem_exponent? 

The PR also changes the description in the declarations.

Happy to further improve the description or names based on your suggestions. 

If you agree to the renaming, I will also need to adjust it in remind2.

## Type of change
- [x] Bug fix 

## Checklist:

@strefler @amerfort 

for reference @VerenaHof @robertsalzwedel

### DIFF
--- a/modules/33_CDR/portfolio/datainput.gms
+++ b/modules/33_CDR/portfolio/datainput.gms
@@ -38,7 +38,10 @@ s33_step = 2.5;
 s33_costs_fix = 0.0422;
 s33_co2_rem_pot = 0.3 * 12/44;       !! default for basalt, for Olivine 1.1
 
-*' rock weathering rate at ambient temperature (25 degree C): eq 2 in strefler, amann et al. (2018): wr = grain surface area based WR (10^-10.53 mol m^-2 s^-1) * molar weight of basalt/forsterite (140.7 g/mol) * 3.155^7 s/a * SSA(gs)
+*' rock weathering rate (i.e. fraction of rock weathering per year) at ambient temperature (25 degree C), based on 
+*' eq 2 in strefler, amann et al. (2018): 
+*' wr = grain surface area based weathering rate (10^-10.53 mol m^-2 s^-1) * molar weight of basalt/forsterite (140.7 g/mol) 
+*'      * 3.155^7 s/a * specific surface area(depending on grain size cm_gs_ew)
 s33_rock_weath_rate_ambientT = 10**(-10.53) * 125 * 3.155*10**7 * 69.18*(cm_gs_ew**(-1.24));
 *' rock weathering rate for different climate grades: SI Tab F-1 of strefler, amann et al. (2018)
 p33_rock_weath_rate("1") = s33_rock_weath_rate_ambientT * 0.94;
--- a/modules/33_CDR/portfolio/datainput.gms
+++ b/modules/33_CDR/portfolio/datainput.gms
@@ -38,10 +38,11 @@ s33_step = 2.5;
 s33_costs_fix = 0.0422;
 s33_co2_rem_pot = 0.3 * 12/44;       !! default for basalt, for Olivine 1.1
 
-*' carbon removal rate: eqs 2+c1 in strefler, amann et al. (2018): wr = grain surface area based WR (10^-10.53 mol m^-2 s^-1) * molar weight of basalt/forsterite (140.7 g/mol) * 3.155^7 s/a * SSA(gs)
-s33_co2_rem_fraction = 10**(-10.53) * 125 * 3.155*10**7 * 69.18*(cm_gs_ew**(-1.24));
-p33_co2_rem_rate("1") = -log(1-s33_co2_rem_fraction * 0.94);
-p33_co2_rem_rate("2") = -log(1-s33_co2_rem_fraction * 0.29);
+*' rock weathering rate at ambient temperature (25 degree C): eq 2 in strefler, amann et al. (2018): wr = grain surface area based WR (10^-10.53 mol m^-2 s^-1) * molar weight of basalt/forsterite (140.7 g/mol) * 3.155^7 s/a * SSA(gs)
+s33_rock_weath_rate_ambientT = 10**(-10.53) * 125 * 3.155*10**7 * 69.18*(cm_gs_ew**(-1.24));
+*' rock weathering rate for different climate grades: SI Tab F-1 of strefler, amann et al. (2018)
+p33_rock_weath_rate("1") = s33_rock_weath_rate_ambientT * 0.94;
+p33_rock_weath_rate("2") = s33_rock_weath_rate_ambientT * 0.29;
 
 *' JeS FE demand fit from Thorben: SI D in strefler, amann et al. (2018)
 p33_fedem("weathering", "feels") = 6.62 * cm_gs_ew**(-1.16);
--- a/modules/33_CDR/portfolio/datainput.gms
+++ b/modules/33_CDR/portfolio/datainput.gms
@@ -39,9 +39,9 @@ s33_costs_fix = 0.0422;
 s33_co2_rem_pot = 0.3 * 12/44;       !! default for basalt, for Olivine 1.1
 
 *' carbon removal rate: eqs 2+c1 in strefler, amann et al. (2018): wr = grain surface area based WR (10^-10.53 mol m^-2 s^-1) * molar weight of basalt/forsterite (140.7 g/mol) * 3.155^7 s/a * SSA(gs)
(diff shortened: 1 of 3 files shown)


modules/33_CDR/portfolio/declarations.gms
```diff
@@ -9,7 +9,7 @@ scalars
 s33_capture_rate            "CO2 capture rate for capturing emissions, e.g., from burning natural gas" / 0.9 /
 
 s33_co2_rem_pot             "specific carbon removal potential [Gt C per Gt ground rock]"
-s33_co2_rem_rate            "carbon removal rate [fraction of annual reduction of total carbon removal potential]"
+s33_co2_rem_fraction        "fraction of stone weathering per year [fraction of annual reduction of total carbon removal potential], independent of climate grade"
```
amerfort, line 12: @strefler I find the description still hard to interpret and don't really understand the rationale behind this scalar. Why would it reduce the total carbon removal potential annually?

modules/33_CDR/portfolio/declarations.gms
```diff
@@ -22,7 +22,7 @@ s33_OAE_glo_limit           "global limit for OAE [tC / a]"
 parameters
 p33_fedem(all_te,all_enty)               "final energy demand of each technology [EJ/GtC] (for EW the unit is [EJ/Gt stone])"
 p33_LimRock(all_regi)                    "regional share of EW limit [fraction], calculated ex ante for a maximal annual amount of 8 Gt rock in D:\projects\CEMICS\paper_technical\supply_curve_transport_remind_regions.m"
-p33_co2_rem_rate(rlf)                    "carbon removal rate [fraction of annual reduction of total carbon removal potential], multiplied with grade factor"
+p33_co2_rem_rate(rlf)                    "stone weathering rate, exponent in exponential decay function (based on percent action of stone weathering and including climate grade)"
```
amerfort, line 25: Thanks for spotting this Tabea! As it seems to be the stone weathering rate rather than the actual removal rate, I would be in favour of renaming the parameter accordingly.
Maybe p33_rockew_rate? What do you think?

modules/33_CDR/portfolio/declarations.gms
tabeado: In my understanding it follows from the  calculation of the removal per time step:
rocks_on_field * share_weathering * cum_removal_potential 
So the share weathering is equal to the reduction of the remaining removal potential. But I agree that it is confusing, and would be in favor or removing it.

modules/33_CDR/portfolio/declarations.gms
tabeado: How about p33_rock_weath_rate?
We should also rename it to s33_rock_weath_fraction then.
Depends on whether we want to shift the focus from the removal to the weathering (as per your other comment)

modules/33_CDR/portfolio/declarations.gms
amerfort: I looked it up in Jess paper. 
Our friend `s33_co2_rem_rate` is d(x) from Equation (1) in the [paper](https://iopscience.iop.org/article/10.1088/1748-9326/aaa9c4/pdf). That is **the weathering rate at ambient temperature (25°C)**. But I don't know what that means in terms of humidity. Maybe there is an established standard in this field of science.
In any case, [in the SI](https://content.cld.iop.org/journals/1748-9326/13/3/034010/revision2/ERL_034010_SD.pdf?Expires=1734966574&Signature=ddFnWUZqCnXpE5HLHikPUOvNPLNtLM1fbQidbIkDySR6j~YCetjfM74G-KvOl2xsxUVnriVGMTXUyYHkvNOgwj7ihT879osBGZxCoUqm1Pq9NLibXz9y6IN-KlIGPviYJj0x6exijz9-39h6OqIdOiXWR0iNdDy9TpPgcKOUDAOCYGvgJfy1HAMJvA71LTA3ooLblU01bqHgTDVl7oUweW3gXUbxJFieUHXo9hDm8h-s8ZqKYYXlpkzQn-gPzYfSzHCYrEOi5esvJuTIlnLAojgpEjIPkOIM~N7tC0drApL9Qi3ekZxUeLVipYt3LlO4P8Ai3BVtB1o-CaXafjCaGA__&Key-Pair-Id=KL1D8TIY3N7T8), there is a table with weathering rate correction factors to account for temperature deviations from the ambient case in Eq.1.
And these numbers are also multiplied to get from s33_co2_rem_rate to p33_co2_rem_rate.
But some of the calculation on how to calculate the weathered rocks is also hidden in the calculation of p33_co2_rem_rate and cannot be understood by looking at only either the datainput or the equations.
Is there a reason to have the logarithm in the calculation of p33_co2_rem_rate and then revert that kinda in the equations.gms with the exponent? Yes it shortens the code in equations-gms, but it is less transparent and also naming p33_co2_rem_rate becomes quite tricky.

tabeado:
Test run with the new formulation: /p/tmp/tabeado/remind_clean_new/remind/output/SSP2-PkBudg650-EW_2024-12-18_19.04.40

image

strefler:
> Updated suggestion to be approved by @strefler:
> 
>     * We renamed `s33_co2_rem_rate `and  `p33_co2_rem_rate `to  `s33_rock_weath_rate_ambientT` and  `p33_rock_weath_rate`:
>       - The scalar provides the fraction of stone weathering each year at ambient temperature. This is now reflected in the name.
>       - The parameter provides the fraction of stone weathering each year for the two climate grades.
> 
>     * We simplify the equations `q33_EW_onfield_tot `and `q33_EW_emi`:
>       - The logarithmus in the old `p33_co2_rem_rate` and the exponential functions in these equations cancel out. We suggest to avoid the exponential formulation. Or was there a computational GAMS reason to write it this way?
>       - The change means that the content of `p33_rock_weath_rate ` has a direct simple interpretation (changed from the exponent of the exponential function to simply the rock fraction weathering per year)

Thanks! I like the renaming. Regarding the simplification, I'm not sure whether it is easier for GAMS to deal with an expression like c * exp(x), or x**c. Both of them involve power functions, and my gut feeling is, that exponential might be easier, but I'm not sure. However, in the last change of the equations file, there is one exponential less, so maybe that helps. Might make sense to involve RSE on that question.

The test runs show quite a bit of change, I would have expected pretty much the same outcome. Do we know where the deviation comes from?

tabeado:
@strefler we consulted @LaviniaBaumstark who shared your concern. 

The compromise is to move the log calculation to the equation. This keeps the accessible interpretation of p33_rock_weath_rate and makes also q33_EW_emi simpler to read. The calculation of v33_EW_onfield_tot may look a bit more complicated at first, but I added an explanation in the description which should support understanding. As all exp & log are at the same place now, it also doesn't require to look in 2 documents to understand the equation.

modules/33_CDR/portfolio/datainput.gms
```diff
@@ -38,10 +38,11 @@ s33_step = 2.5;
 s33_costs_fix = 0.0422;
 s33_co2_rem_pot = 0.3 * 12/44;       !! default for basalt, for Olivine 1.1
 
-*' carbon removal rate: eqs 2+c1 in strefler, amann et al. (2018): wr = grain surface area based WR (10^-10.53 mol m^-2 s^-1) * molar weight of basalt/forsterite (140.7 g/mol) * 3.155^7 s/a * SSA(gs)
-s33_co2_rem_rate = 10**(-10.53) * 125 * 3.155*10**7 * 69.18*(cm_gs_ew**(-1.24));
-p33_co2_rem_rate("1") = -log(1-s33_co2_rem_rate * 0.94);
-p33_co2_rem_rate("2") = -log(1-s33_co2_rem_rate * 0.29);
+*' rock weathering rate at ambient temperature (25 degree C): eq 2 in strefler, amann et al. (2018): wr = grain surface area based WR (10^-10.53 mol m^-2 s^-1) * molar weight of basalt/forsterite (140.7 g/mol) * 3.155^7 s/a * SSA(gs)
```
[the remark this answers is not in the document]
tabeado: very helpful! I added it.

tabeado:
I made a final check comparing the previous and current formulation, just to be sure. There are very minor deviations that I don't understand. My guess would be it is due to some minor numerical differences in the calculation? 
image

It wouldn't stop me from merging, but @amerfort could you make a final check that I didn't make a random mistake in the reformulation?

PR #1934, opened 2024-12-18
Bugfix: Where 2030 capacity bounds are distributed to regions by GDP share, use 2025 GDP instead of 2030 GDP to avoid potential infes in non-SSP2 policy runs

fschreyer:
## Purpose of this PR

Where global capacity bounds are distributed to regions by GDP share, use GDP 2025 share instead of later years as GDP can be different across SSPs leading to infes.

## Type of change

- [x] Bug fix 

## Checklist:

* Test runs are here: 


laurinks (approved):
Looks good to me. GDP is indeed equal for all SSPs in 2025 so this option should be safe. Thanks, @fschreyer, for fixing this quickly!

PR #1940, opened 2025-01-06
 fail if coupled config has path_gdx_ref column but not path_gdx_refpolicycost

orichters:
## Purpose of this PR

- When reading a config file and `path_gdx_ref` is specified but `path_gdx_refpolicycost` isn't, it just copies it over, which makes sense. But that means: if you specify `path_gdx_ref` and `path_gdx_refpolicycost` separately in the normal config file, but only `path_gdx_ref` in the coupled config file, then `path_gdx_ref` from the coupled config is used to overwrite `path_gdx_refpolicycost`, which I didn't expect and it created an error in the NGFS scenarios. The problem is that the way the code is written, the coupled config cannot know at this point whether the column exist in the other config file. 
- Therefore, `start_bundle_coupled.R` now fails if `path_gdx_ref` exists in the coupled config, but not the `path_gdx_refpolicycost` column.
- Adapted all the existing config files
- removed GCS coupled config that @gabriel-abrahao said was not needed anymore.
- reduce default runtime in tests to 20 hours to make sure the test slurm of `make test-full-slurm` is not running into a timeout because the default takes too long.

## Type of change

- [x] Bug fix 

## Checklist:

- [x] I did not adjusted the reporting in [`remind2`](https://github.com/pik-piam/remind2) as it wasn't needed
- [x] I did not adjust `forbiddenColumnNames` in [readCheckScenarioConfig.R](https://github.com/remindmodel/remind/blob/develop/scripts/start/readCheckScenarioConfig.R)
- [x] All automated model tests pass (`FAIL 0` in the output of `make test` and `make test-coupled`)


dklein-pik (approved):
Thank you

PR #1951, opened 2025-01-16
Bugfix: change fixing to 0 of coalh2 and gash2 to (very low) upper bound

robertpietzcker:
## Purpose of this PR

currently, there are conflicting bounds on vm_deltacap and vm_cap for coalh2 and gash2, which sometimes make the model infeasible. (it is unclear to me WHY the model can be feasible at all at the moment) 

## Type of change

- [x] Bug fix 

## Checklist:


fschreyer (approved):
Ok, if you say that this caused issues. 

Generally, I would think we should remove upper bounds also for vm_cap of time steps where technologies are not supposed to be built. I did that for bioh2, for instance.

robertpietzcker:
I'll merge for now to make sure this doesn't hinder the current runs

PR #1958, opened 2025-01-24
Remove the explicit version setting for renv and update to renv 1.1.1

dklein-pik:
## Purpose of this PR

Remove code from .Rprofile that explicitly defines a renv version to avoid jumping back and forth with versions, which occurres when a newer version of renv becomes available. Update to renv 1.1.1


pascal-sauer:
I'd suggest to now also remove `exclude = "renv"` from `Makefile`, so renv is updated when running `make update-renv-all` just like all other packages. Alternatively users can also run `renv::install('renv')` to update only renv to get rid of the startup message

pascal-sauer:
Yes :)

pascal-sauer (approved):
Looks good now. The PR title still says we upgrade to renv 1.0.11 :shrug:

pascal-sauer:
I was able to just change the title

PR #1962, opened 2025-01-28
Fix biotr trajectory instead of upper bound 

tabeado:
## Purpose of this PR
Phaseout of biotr was prescribed through an upper bound on vm_deltaCap. However, REMIND no longer exploited this bound, leading to inconsistencies with cm_maxProdBiolc. We replace .up with .fx because the prescribed path should be followed unless a faster phaseout is explicitly chosen.

## Type of change
- [x] Bug fix 

## Checklist:


merfort:
Thanks, Tabea, for fixing this (haha).
I guess it makes sense to also do that for the non-SSP2 scenarios [here](https://github.com/remindmodel/remind/blob/a915fe9dbb28648f5ee195b46b137af0ab1ef520/core/bounds.gms#L130-L135), right? (actually only relevant for SSP1 and SSP5)

tabeado:
good catch! I added it quickly. 
But just to check: for SSP1 and SSP5 this would mean that there is no complete phaseout, is that intended?

merfort:
Thanks for adding it! 
No, this is implemented a bit differently. Here the original trajectory of the bound (that is defined above in the loop) is simply adjusted afterwards for each time step `t` and multiplied by 0.5. That is, it is not defined based on the reference time step `t = 2005`. 
Interestingly, it is refers to the lower bound, though. That means, that it was not working properly, because before the lower bound was never applied (only `.up`) and was always 0 (as defined via the `.fx` statement [here](https://github.com/remindmodel/remind/blob/develop/core/bounds.gms#L110)). 
Now with the changes it will also work with using `.lo`, but I would suggest, to also change the line to
```
$if %cm_GDPscen% == "gdp_SSP5" vm_deltaCap.fx(t,regi,"biotr","1")$(t.val gt 2020) = 0.5 * vm_deltaCap.fx(t,regi,"biotr","1");
```
because it is more explicit.

core/bounds.gms
```diff
@@ -131,8 +131,8 @@ $if %cm_GDPscen% == "gdp_SDP" vm_deltaCap.up(t,regi,"biotr","1")$(t.val gt 2020)
 $if %cm_GDPscen% == "gdp_SDP_EI" vm_deltaCap.up(t,regi,"biotr","1")$(t.val gt 2020) = 0;
 $if %cm_GDPscen% == "gdp_SDP_MC" vm_deltaCap.up(t,regi,"biotr","1")$(t.val gt 2020) = 0;
 $if %cm_GDPscen% == "gdp_SDP_RC" vm_deltaCap.up(t,regi,"biotr","1")$(t.val gt 2020) = 0;
-$if %cm_GDPscen% == "gdp_SSP1" vm_deltaCap.up(t,regi,"biotr","1")$(t.val gt 2020) = 0.5 * vm_deltaCap.lo(t,regi,"biotr","1");
-$if %cm_GDPscen% == "gdp_SSP5" vm_deltaCap.up(t,regi,"biotr","1")$(t.val gt 2020) = 0.5 * vm_deltaCap.lo(t,regi,"biotr","1");
+$if %cm_GDPscen% == "gdp_SSP1" vm_deltaCap.fx(t,regi,"biotr","1")$(t.val gt 2020) = 0.5 * vm_deltaCap.lo(t,regi,"biotr","1");
+$if %cm_GDPscen% == "gdp_SSP5" vm_deltaCap.fx(t,regi,"biotr","1")$(t.val gt 2020) = 0.5 * vm_deltaCap.lo(t,regi,"biotr","1");
```
merfort, line 135:
```suggestion
$if %cm_GDPscen% == "gdp_SSP1" vm_deltaCap.fx(t,regi,"biotr","1")$(t.val gt 2020) = 0.5 * vm_deltaCap.fx(t,regi,"biotr","1");
$if %cm_GDPscen% == "gdp_SSP5" vm_deltaCap.fx(t,regi,"biotr","1")$(t.val gt 2020) = 0.5 * vm_deltaCap.fx(t,regi,"biotr","1");
```
See comment.

merfort: Or wait, @dklein-pik can you actually refer to a `.fx` value? Is the reason maybe, that you have to either choose `.lo` or `.up`?

merfort (approved):

tabeado:
Thanks for the clarification. Taking some time to thoroughly read the code clearly helps :)

PR #1964, opened 2025-01-28
Update documentation of calculations that limit BECCS to 50% of total PE

dklein-pik:
Update documentation of calculations that limit BECCS to 50% of total PE.

Closes https://github.com/remindmodel/development_issues/issues/485


modules/30_biomass/magpie_40/equations.gms
```diff
@@ -110,7 +110,13 @@ q30_limitXpBio(t,regi)..
          vm_fuExtr(t,regi,"pebiolc","1")
 ;
 
-*** Limit BECCS in policy runs to 35% of total PE in BAU or to 130% of bioenergy demand in BAU
+*' ** Limit BECCS **
+*' To prevent regions with large carbon storage potential (RUS and MEA)
```
merfort, line 114:
```suggestion
*' To prevent regions with large carbon storage potential
```
We want to get rid of explicit region names, even in comments, right?

modules/30_biomass/magpie_40/presolve.gms
```diff
@@ -7,7 +7,13 @@
 
 *** SOF ./modules/30_biomass/magpie_40/presolve.gms
 
-*** save level value of all taxes
+*** Calcualte total primary energy to limit BECCS (see q30_limitTeBio)
+*** This must be calculated outside the optimization and stored in a 
+*** parameter so as not to create an incentive to increase the total
+*** PE demand just to increase the BECCS limit (see also presolve.gms).
```
merfort, line 13:
```suggestion
*** PE demand just to increase the BECCS limit (see also postsolve.gms).
```
Is it actually still necessary to have it in the pre- and in the postsolve?

modules/30_biomass/magpie_40/postsolve.gms
```diff
@@ -7,6 +7,9 @@
 
 *** SOF ./modules/30_biomass/magpie_40/postsolve.gms
 
+*** Calcualte total primary energy to limit BECCS (see q30_limitTeBio)
```
robertsalzwedel, line 10: Is this the easiest way to calculate pe demand?

robertsalzwedel (approved):
Hey @dklein-pik @merfort, looks great, thanks for fixing this :) 
The only thing I am still confused about is the sum to calculate p30.. I thought there is an easyier way to calculate primary energy but I probably have to get used to gams coding :)

robertsalzwedel (approved):

merfort:
> Thank you, Leon. I deleted postsolve.gms

Thanks, David! But did you check that is really not needed? (I didn't I just raised the question, haha 🙈 )

dklein-pik:
I didnt run REMIND to be honest, but I checked, that the parameter `p30_demPe` that is calculated in presolve.gms is only used in equations.gms. Thus, postsolve is not necessary, is it?

merfort:
Great! I also checked, it is not used in any reporting script, so it does not need to be updated.
Thanks for taking care of that, David! :)

PR #1881, opened 2024-11-12
Update and document CO2 capture rates and costs of X-to-Liquids / X-to-Gases technologies

fschreyer:
## Purpose of this PR

This addresses the [issue on carbon flows in the energy system](https://github.com/remindmodel/development_issues/issues/335). It is related to #1876. 

In particular, this PR adapts the CO2 capture rates of different X-to-Liquids and X-Gases technologies assuming that only 10% of the input carbon are lost (vented) during the gasification / syngas production step (consistent with X-to-H2 w/  capture technologies) and only 5% are lost during the follow-up step of FT-synthesis or methanation. This changes co2 capture rates as follows:

* increase bioftcrec: capture rate to 14.6 GtC/ZJ: 58% of carbon input 
* increase biogasc capture rate to 14.1 GtC/ZJ: 56% of carbon input 
* decrease coalftcrec capture rate to 15.8 GtC/ZJ (60%)
* decrease gasftcrec capture rate to 0.2 GtC/ZJ (1.3%)

CCU technologies (synthetic liquids and synthetic gas production) now also see carbon efficiency of 95%, i.e. 5% of captured carbon does not end up in output fuel.

Moreover, costs of some of the technologies are modified. 

* CAPEX/OPEX of synthetic liquids and synthetic gas technologies back to original parameterization based on Agora (2018)
* slightly lower bioftcrec CAPEX in line with Liu et al. 
* lower gasftrec/gasftcrec CAPEX  in line with assumptions for Coal-to-Liquids / Gas-to-H2 technologies

Please see [issue](https://github.com/remindmodel/development_issues/issues/335) for comparison plots of results. 

## Type of change

(Make sure to delete from the Type-of-change list the items not relevant to your PR)

- [x] Bug fix 

## Checklist:

- [x] My code follows the [coding etiquette](https://github.com/remindmodel/remind/blob/develop/main.gms#L80)
- [x] I performed a self-review of my own code
- [x] I explained my changes within the PR, particularly in hard-to-understand areas
- [x] I checked that the [in-code documentation](https://github.com/remindmodel/remind/blob/develop/main.gms#L120) is up-to-date
- [x] I adjusted the reporting in [`remind2`](https://github.com/pik-piam/remind2) where it was needed
- [x] I adjusted `forbiddenColumnNames` in [readCheckScenarioConfig.R](https://github.com/remindmodel/remind/blob/develop/scripts/start/readCheckScenarioConfig.R) in case the PR leads to deprecated switches
[...]

PR #1985, opened 2025-02-13
Unified SSP techno-economic data

lecfab:
## Purpose of this PR

It was hard to maintain different generisdata_tech files different SSPs:
- https://github.com/remindmodel/development_issues/issues/414

### Major elements of this PR:
- [x] go through all the technologies to choose how their inco0 (and other parameters) should change with SSP narratives. This [google sheet](https://docs.google.com/spreadsheets/d/1DaVj9MwYrytIDKnuRK6A1k5KB3m1DPhMSbqbJPSfjKs/edit?gid=241833046#gid=241833046) shows the reasoning.
- [x] turn all SSP-specific technoeconomics into in-code relative modifications and remove files generisdata_tech for SSP1, 3, 5
- [x] rehabilitate `cm_VRE_supply_assumptions`: it affects inco0 and floorcost for storage as well as learning rate for storage and VRE:
   - _1 optimistic_ improves values by around 10%
   - _2 pessimistic_ worsens values by around 10%
   - _3 very pessimistic_ worsens values by around 30%

### Minor elements:
- [x] merge generisdata_flexibility into generisdata_tech
- [x] replace incolearn by floorcost
- [x] remove absolute in-code modification of inco0, learn etc
- [x] rename the values of `c_techAssumptScen` to SSP1..5 instead of numbers without meaning
- [x] add gasftcrec to teCCS as it was missing
- [x] forbids CCS technologies to be cheaper than non-CCS equivalent (using new set te2teCCS) 
- [x] document https://github.com/remindmodel/development_issues/issues/486#event-16297434367 (side modification)
- [x] remove old code involving `ccscomp`, `ccspipe` etc
- [x] clean stuff; please review in "Hide whitespace mode":

## Type of change

- [x] Bug fix 
- [x] Refactoring
- [x] New feature 

## Checklist:

## Comparison of results
Compare scenarios are here: `/p/tmp/fabricel/202502-remind-learningDoc`.
The former techno-economic assumptions had a lot of bugs due to missed update etc. Hence significant differences after the update:
- SSP1: more gas, less biomass and hydrogen
- SSP2: few differences, less wind
- SSP3: gas replaces coal, lower emissions
- SSP5: less coal, rerouting of gas, lower emissions


core/datainput.gms
```diff
@@ -127,139 +127,189 @@ pm_shfe_lo(ttot,regi,entyFe,sector)=0;
...
-*** ccsinje cost scenarios
+***---------------------------------------------------------------------------
+*** Other technological assumptions (VRE, CCS...)
+***---------------------------------------------------------------------------
+*** cm_ccsinjeCost cost scenarios
+*** TODO: make these costs relative to default values
 *** low estimate: ccsinje cost prior to 03/2024; i.e. ~11 USD/tCO2 in 2025, decreasing to ~7.5USD/tCO2 as of 2035
 $if "%cm_ccsinjeCost%" == "low" fm_dataglob("tech_stat","ccsinje") = 2;
-$if "%cm_ccsinjeCost%" == "low" fm_dataglob("inco0","ccsinje") = 220;
+$if "%cm_ccsinjeCost%" == "low" fm_dataglob("inco0","ccsinje") = 220 / 350 * fm_dataglob("inco0","ccsinje");
```
tabeado, line 240: Please keep the cm_ccsinjeCost at the absolute levels, i.e. 220 and 550, for now, as these had meaning at the actual level and were not chosen on a relative basis. The explanation is [in the PR](https://github.com/remindmodel/remind/pull/1592) and should have been included in the switch description (sorry!). 
Instead of making the switch relational, we can limit it to SSP2 scenarios, to be noted in the switch description + possibly by adding an $if statement here.

Reason why we want to avoid spending too much time on analyzing the cost changes now: We are planning on updating the implementation of ccsinje more generally (e.g. splitting offshore & onshore storage) this year, which will anyways lead to another revision of the cost.

lecfab: Thanks a lot for the explanations, I reverted and added documentation (nothing forbidding to use the combination, but clear warning comments).

tabeado:
thanks a lot for accepting to keep the ccsinjeCost switch as it was and improving the description!

core/datainput.gms
```diff
@@ -127,139 +127,189 @@ pm_shfe_lo(ttot,regi,entyFe,sector)=0;
...
+    if (cm_VRE_supply_assumptions eq 1, p_VRE_assumption_factor = 0.9); !! optimistic assumptions
+    if (cm_VRE_supply_assumptions eq 2, p_VRE_assumption_factor = 1.1); !! pessimistic assumptions
+    if (cm_VRE_supply_assumptions eq 3, p_VRE_assumption_factor = 1.3); !! very pessimistic assumptions
+    fm_dataglob("learn",teVRE) =  1 / p_VRE_assumption_factor * fm_dataglob("learn",teVRE);
+    fm_dataglob("learn",teStor) = 1 / p_VRE_assumption_factor * fm_dataglob("learn",teStor);
+    fm_dataglob("inco0",teStor) =     p_VRE_assumption_factor * fm_dataglob("inco0",teStor); 
+    fm_dataglob("floorcost",teStor) = p_VRE_assumption_factor * fm_dataglob("floorcost",teStor);
+);
+
+*RP* rescale the global CSP investment costs in REMIND: Originally we assume a SM3/12h setup, while the cost data from IEA for the short term seems rather based on a SM2/6h setup (with 40% average CF)
```
lecfab, line 257: @robertpietzcker After reading in the data, the csp costs are all reduced to 70%. Is that still a reasonable modification??

```js
# *RP* rescale the global CSP investment costs in REMIND:
# Originally we assume a SM3/12h setup, while the cost data from IEA for the short term
# seems rather based on a SM2/6h setup (with 40% average CF)
# Accordingly, also decrease long-term costs in REMIND to 0.7 of the current values
fm_dataglob("inco0","csp")     = 0.7 * fm_dataglob("inco0","csp");
fm_dataglob("incolearn","csp") = 0.7 * fm_dataglob("incolearn","csp");
```

core/input/generisdata_tech.prn
```diff
@@ -2,300 +2,306 @@
...
+eta           0.42       0.33     0.41        0.38      0.75      0.95      0.60        0.40         0.40      0.57       0.53
+omf           0.03       0.03     0.03        0.03      0.03      0.03      0.03        0.05         0.06      0.05       0.06
+omv             38         61       30          72        36                              43           43         6          6
+flexibility   0.85        0.7      0.9        0.75
+
+
+*** biomass technologies
++             biotr   biotrmod   biochp   biohp   bioigcc   bioigccc   biogas   biogasc   bioh2   bioh2c   bioethl   bioeths   biodiesel   bioftrec   bioftcrec
+tech_stat                                               1          3        1         3       2        3         2                                3           3
+inco0            10        300     3000     550      2450       3150     1200      1230    1680     2040      2860       590         160       2800        3100
```
lecfab, line 60: @robertpietzcker @nicobauer 

Why are biogas and biogasc investment costs so close?
I noticed because with the SSP5 assumptions of cheap CCS, biogasc even becomes cheaper than biogas.

nicobauer (approved):

core/datainput.gms
robertpietzcker: yes

lecfab: ok, i added it directly into generisdata_tech

core/input/generisdata_tech.prn
robertpietzcker: That surprises me!

I just checked - it was briefly discussed in this issue https://github.com/remindmodel/development_issues/issues/335#issuecomment-2606978934 and @fschreyer implemented it in this commit: https://github.com/remindmodel/remind/pull/1881

I think Felix wanted to have a citeable source for it, and there is a 2019 paper (Michailos et al. (2019), https://doi.org/10.1016/j.bej.2019.107290) that says they cost pretty much the same. 

In my eyes this is inconsistent with the overall markup of ~200-400$/kW that we otherwise apply when adding CCS to a xxx->h2 or xxx->liquids route (eg bioh2, bioftrec,  - so I would find it more consistent if we again increase the biogasc costs to biogas+250$/kW, and similarly coalftcrec to coalftrec + 250$ or so

lecfab: There is an explanation here: https://www.biosng.com/fileadmin/biosng/user/documents/reports/The_Economy_of_Large_Scale_Biomass_to_Bio-Methane.pdf

> ### BioCCS
> BioSNG production will always result in a large flow high-purity CO2. The volume of CO2
roughly equals the bioSNG production on a volume basis. Contrary to Carbon Capture
and Storage (CCS) applied to reduce CO2-emissions from fossil fuel processes, CCS
combined with bioSNG production hardly involves an energy penalty, since CO2
separation is an integral part of the bioSNG process. Implementation of capture and
storage of CO2 in bioSNG plants is therefore a relatively efficient and cheap method to
reduce CO2 concentrations in the atmosphere, allowing negative CO2 emissions.
Moreover, costs of bioSNG might be reduced in cases where there is a market for the
produced CO2.


But I agree with you that this is strange. In the meantime, I enforced that CCS has to be 1% more expensive than non-CCS equivalent, even when SSP favours CCS. Do you think that's ok until we maybe revise the costs?

robertpietzcker: I now started a new issue: https://github.com/remindmodel/remind/issues/2000
so we can follow up on this after this PR is merged

robertpietzcker: thanks for the link to the ECN report - it would be interesting to find out why this "CO2 separation" would not be part of a coal2gas process?!

robertpietzcker: and yes, I would say enforcing that CCS always has to be at least 1% more expensive definitely makes sense

cchrisgong (commented):
Thanks a lot!
Why is there a lot of coal to liquid without CC in SSA under SSP3? There is less after the change but rather strange...

lecfab:
> Thanks a lot! Why is there a lot of coal to liquid without CC in SSA under SSP3? There is less after the change but rather strange...

Indeed, it's a separate known issue: 
- https://github.com/remindmodel/development_issues/issues/433

This PR can influence it because the Fischer Tropsch costs changed significantly.

core/datainput.gms
```diff
@@ -127,139 +127,189 @@ pm_shfe_lo(ttot,regi,entyFe,sector)=0;
...
+    fm_dataglob("inco0",teStor) =     p_VRE_assumption_factor * fm_dataglob("inco0",teStor); 
+    fm_dataglob("floorcost",teStor) = p_VRE_assumption_factor * fm_dataglob("floorcost",teStor);
+);
+
 *** Overwrite default technology cost parameter values based on specific scenario configs
-$if not "%cm_incolearn%" == "off"       parameter p_new_incolearn(all_te) / %cm_incolearn% /;
-$if not "%cm_incolearn%" == "off"                 fm_dataglob("incolearn",te)$p_new_incolearn(te) = p_new_incolearn(te);
+$ifthen.cm_incolearn not "%cm_incolearn%" == "off"
+    parameter p_new_incolearn(all_te) "value to overwrite incolearn parameters" / %cm_incolearn% /;
+    fm_dataglob("floorcost",te) $ p_new_incolearn(te) = fm_dataglob("inco0",te) - p_new_incolearn(te);
```
robertpietzcker, line 260: @lecfab I don't understand this GAMS syntax?

what is the conditional in this $ condition? is it that p_new_incolearn(te) is <>0? 
I think it is clearer if the conditional is encapsulated in brackets to show clearly what the conditional is, and I think this is also how it is implemented in most of the REMIND code

lecfab: Correct, when p_new_incolearn is zero it evaluates to False, otherwise True.
Noted for the brackets! i thought we only put them to wrap a "or" or "and" etc.
In that very example, I just copied the line shown in red that contained this syntax, and put it over several lines instead of one.

core/datainput.gms
```diff
@@ -127,139 +127,189 @@ pm_shfe_lo(ttot,regi,entyFe,sector)=0;
...
-               fm_dataglob(char,te) = f_dataglob_SSP1(char,te)
-);
-if (c_techAssumptScen eq 3,
-               fm_dataglob(char,te) = f_dataglob_SSP5(char,te)
-);
-if (c_techAssumptScen eq 4,
-               fm_dataglob(char,te) = f_dataglob_SSP3(char,te)
-);
 
-*RP* include global flexibility parameters
```
robertpietzcker, line 199: did you have a specific reason to merge the flexibility file with generisdata_tech? 

I would prefer to keep them separate, as the stuff in generisdata_tech is really fundamental techno-economic properties, while the values in _flexibility are hand-tuned parameters to create reasonable technology mixes for a variety of power system configurations. 

so if you don't have a very strong preference, I would propose to keep it split, @lecfab

lecfab: Sorry Robert I had merged already as I thought you had finished commenting.

For flexibility data, I found it clearer to add it into the general tech file, as it did not add too much clutter there. I have to admit that I still don't fully understand this parameter.
So it's perfectly ok with me if you re-separate them, and maybe add some documentation to the file?

lecfab:
> why there are changes in SSP2 results in the power sector?
from what I see, you didn't implement any changes to input data in the power sector in SSP2, only in SSP1/SSP3?

That's a puzzle for me too, and I started digging into it: it could come from using floorcost instead of incolearn.
At some point there is a formula $$inco0' = \alpha * inco0$$ where $$\alpha = 1 + tkpremused$$.
- Formerly, incolearn was fixed, so we had $$floorcost = inco0' - incolearn = \alpha * inco0 - incolearn$$
- Now, floorcost is fixed using the former value of incolearn: $$floorcost = inco0 - incolearn$$

So this $$\alpha$$ could be responsible for different dynamics for learning technologies in the new version. However, I have a feeling it should actually _increase_ their deployment. So this needs further investigation!

laurinks:
Amazing PR, @lecfab ! It is so nice that the assumptions across SSPs are well-structured, clean and easy-to-maintain by now! :)

cchrisgong:
Do we really need this turnkey parameter? I feel like it anyhow makes only a tiny difference to the result, it's regionally differentiated and is a bit hard to justify from empirics since different regions have such a diversity of financing... I also suspect this could be double coding if we also consider waccs which would have taken into account the different construction times of plants

lecfab:
@cchrisgong i would agree with you, but I have no idea about such topics, so maybe we can open an issue about that for the expert colleagues to decide?

lecfab:
Discussion on turnkey costs continues here:
- https://github.com/remindmodel/development_issues/issues/534

PR #1967, opened 2025-01-29
Update (optionally) data and base year used to calculate non-CO2 emissions

gabriel-abrahao:
## Purpose of this PR
Requires #1955 . REMIND represents non-CO2 emissions in various different ways for different emission sources. The general rationale is:
 - Calculate a baseline emissions trajectory, either completely exogenous or based on some endogenous activity level
 - Using the endogenous carbon price, apply MACCs on top of those baseline trajectories
 
 ### Current state
This PR addresses just the first step, the calculation of baseline trajectories, for several emission sources. Below is a complete-ish summary on how non-CO2 emissions were calculated before this PR:
```
# Emission factors calculated from EDGAR 2005 emissions and REMIND activities (updated in #1955 )
        ch4coal    "fugitive emissions from coal mining"
        ch4gas     "fugitive emissions from gas production"
        ch4oil     "fugitive emissions from oil production"
# Econometric estimate, "calibrated" based on EDGAR 2005 and REMIND population and GDP
        ch4wstl    "ch4 emissions from solid waste disposal on land"
        ch4wsts    "ch4 emissions from waste water"
        n2owaste   "n2o emissions from waste (domestic sewage)"
# Exogenous baseline trajectories (IMAGE, Lucas 2007), scaled to EDGAR2005
        n2otrans   "n2o emissions from transport"
        n2oadac    "n2o emissions from adipic acid production"
        n2onitac   "n2o emissions from nitric acid production"
# Exogenous (Recent MAgPIE)
        ch4rice    "ch4 emissions from rice cultivation (rice_ch4)"
        ch4peatland "ch4 emissions from peatlands (no MAC available)"
        ch4animals "ch4 emissions from enteric fermentation of ruminants (ent_ferm_ch4)"
        ch4anmlwst "ch4 emissions from animal waste management(awms_ch4)"
        n2ofertin  "n2o emissions from Inorganic fertilizers (inorg_fert_n2o)"
        n2ofertcr  "n2o emissions from decay of crop residues (resid_n2o)"
        n2ofertsom "n2o emissions from soil organic matter loss (som_n2o)"
        n2oanwstc  "n2o emissions from manure applied to croplands (man_crop_n2o)"
        n2oanwstm  "n2o emissions from animal waste management (awms_n2o)"
        n2oanwstp  "n2o emissions from manure excreted on pasture (man_past_n2o)"
        n2oagwaste "n2o emissions from agricultural waste burning (no MAC available)"
[...]

PR #1987, opened 2025-02-13
Change default for cm_emifacs_baseyear

gabriel-abrahao:
## Purpose of this PR

Changes  the default for `cm_emifacs_baseyear` from 2005 to 2020. This changes both the base year and the data sources used to derive non-CO2 emissions baselines. 

Test runs and comparisons between the old (2005) and new (2020) defaults can be found under #1967


gabriel-abrahao:
@dklein-pik this is what I was talking about in the coupled validation meeting, this PR changes the default so you don't have to worry about setting it for the next runs.

PR #1994, opened 2025-02-18
Update CH4 and N2O MACCs

gabriel-abrahao:
## Purpose of this PR

This PR introduces new Marginal Abatement Cost Curves (MACCs) for CH4 and N2O from [Harmsen et al. 2019](https://www.sciencedirect.com/science/article/abs/pii/S146290111830889X), including the Pessismistic, Optimistic and Default versions used in [Hamsen et al. 2022](https://www.nature.com/articles/s41467-023-38577-4). These can be selected with `c_nonco2_macc_version` (`PBL_2022` for the the new ones and `PBL_2007 for the original Lucas et al. 2007 ones) and `c_nonco2_macc_scenario` for  `Pessismistic`, `Optimistic` or `Default`. It also changes the defaults to `PBL_2022` and `Default`.  

These new MACCs (`PBL_2022`) include abatement measures with costs all the way up to ~4000 $/tCO2eq, while the old ones (`PBL_2007`) would only go up to ~1000 $/tCO2eq. Note that the monetary units and steps in the implementation vary. See the PDF below for a comparison of the implemented MACCs for all periods up to 2100, with compatibilized costs as implemented. `PBL_2022` has a somewhat stronger time dependence reflecting the time for implementation of different measures considered reasonable, and actual mitigation only begins in 2025.  

[maccs_allyears.pdf](https://github.com/user-attachments/files/18853660/maccs_allyears.pdf)

macs_huge_2050

Note that, for most of the CH4 sectors, `PBL_2022` is a *downwards* revision of maximum abatement potentitals, even in the `Optimistic` variant.

Also note that a version of `PBL_2022` is already the default for abatement of agricultural sources in MAgPIE. However, they were able to implement it differently as there's more granular control of activity levels, so do not expect an exact match in coupled runs.  

## Type of change
- [X] New feature 

## Checklist:

### Runs with these changes are here:

The mitigation scenario with the `Optimistic` variant had some trouble converging, but it seems to be a general REMIND difficulty around now. So SSP2 tests for NPi and PkBudg500 can be found archived at:
`/p/projects/piam/abrahao/scratch/updatenonco2/remind/output/bkp_pr1`

but should be enough for review.

New runs after pulling from develop are underway at:
`/p/projects/piam/abrahao/scratch/updatenonco2/remind/output/bkp_pr1`

### Comparison of results (what changes by this PR?): 

Below are emissions results from the scenarios mentioned above. They also include the step change from using EDGAR2005 to CEDS2020 and the new IMAGE scenarios to adjust baseline emissions as implemented in #1967 . More detailed breakdowns and compareScenarios can be found in the REMIND folder. 

GHGCH4N2O_mit_global

Note that the new default baseline setting (CEDS2020-PBL2007) leads to slightly less GHG CO2eq emissions in NPi 2050 relative to the old baseline (EDGAR2005-PBL2007). However, as its abatement is weaker, the new default MACC (CEDS2020-PBL2022Default) leads to higher GHG emissions than the old MACC with the new baseline (CEDS2020-PBL2007) and higher even than the recently old default (old baseline old MACCs EDGAR2005-PBL2007). The difference is about ~1 GtCO2eq/yr in the mitigation scenario.

| Emissions in 2050 | Emi\|CH4  (MtCH4/yr)| Emi\|N2O  (MtN2O/yr) | Emi\|GHG   (GtCO2eq/yr)|
| :-- | :-- | :-- | :-- |
| SSP2-NPi2025-EDGAR2005-PBL2007 | 311.9 | 10833.5 | 43.2 |
| SSP2-NPi2025-CEDS2020-PBL2007 | 295.0 | 10357.0 | 42.7 |
| SSP2-NPi2025-CEDS2020-PBL2022Default | 331.4 | 10014.0 | 43.5 |
| SSP2-NPi2025-CEDS2020-PBL2022Pessimistic | 335.0 | 10552.5 | 43.7 |
| SSP2-NPi2025-CEDS2020-PBL2022Optimistic | 325.2 | 9114.0 | 43.0 |
| SSP2-PkBudg500-EDGAR2005-PBL2007 | 168.3 | 9163.0 | \-9.2 |
| SSP2-PkBudg500-CEDS2020-PBL2007 | 168.0 | 9004.5 | \-9.2 |
| SSP2-PkBudg500-CEDS2020-PBL2022Default | 231.4 | 7179.2 | \-8.0 |
| SSP2-PkBudg500-CEDS2020-PBL2022Pessimistic | 245.7 | 9167.3 | \-7.2 |

## ScenarioMIP
@laurinks @nicobauer you might want to consider choosing the Optimistic and Pessimistic variants for some scenarios depending on the narrative. These variants were based on a Monte Carlo selection across case studies in the literature. In the case of the rollback, I don't think this will make a difference.


laurinks:
Great summary! Thanks a lot for your work, @gabriel-abrahao !

> @laurinks @nicobauer you might want to consider choosing the Optimistic and Pessimistic variants for some scenarios depending on the narrative. These variants were based on a Monte Carlo selection across case studies in the literature. In the case of the rollback, I don't think this will make a difference.

Interesting point! The only scenario narrative, where we might want to choose `Pessimistic`, is the high scenario (rollback), but there is likely no difference as you write since the CO2 price goes down to 0$. `Optimistic` could be a good fit for the Very Low with Limited Overshoot (VLLO) scenario. However, we already see a very strong CH4 reduction due to the diet changes in MAgPIE in that scenario, and the question would be if we want to add the optimistic assumptions for MACCs in REMIND on top of it. Add this point, I would say that we could simply test it in the next round of runs for the VLLO. 

More generally, would you see a good reason to link the choice to SSPs instead of emission scenarios? For example, Optimistic in SSP1, Default in SSP2, and Pessimistic in SSP3. To judge this, I would need to understand where the differences come from. Is it some fundamental uncertainty (e.g. linked to physical processes)? Or is it an uncertainty related to the ability to successfully implement certain measures? In the first case, I would opt for always choosing the default (in the same way that we look at the Median Temperature outcome for all scenarios, not distinguishing across scenarios). In the latter case, I would opt for linking it to the SSPs or the specific ScenarioMIP scenarios.

laurinks:
And two simple question for my understanding. First question: What is the exact meaning of `Abatement fraction`? 

1. Relative to the historical CEDS data for 2020: A CO2 price of 500 $ in year 2050 implies that only a fraction X of the value given by the historical emissions in 2020 is emitted in 2050.
2. Relative to baseline emissions trajectory (agrees with historical CEDS for 2020, but differs thereafter): A CO2 price of 500 $ in year 2050 implies that only a fraction X of the value given by the baseline emissions is emitted in 2050.
3. Relative to baseline value for each activity, say CH4 emissions per ton of produced rice: A CO2 price of 500 $ in year 2050 implies that only a fraction X of the baseline value is emitted for producing a ton of rice in 2050.
4. Something else?

Second question: Is there any inter-temporal relationship in the implementation? To give a concrete example, does it make a difference for CH4 emissions in 2050 whether (a) the CO2 price was 0$ until 2045 and 1000$ in 2050, or (b) the CO2 price was 1000$ from 2030 to 2050.

core/datainput.gms
```diff
@@ -897,6 +897,13 @@ $offdelim
 /
 ;
 
+
+* GA Read choice of non-CO2 MACs version
```
LaviniaBaumstark, line 901: please use 3*s for comments

core/datainput.gms
```diff
@@ -911,6 +918,34 @@ $include "./core/input/p_abatparam_N2O.cs4r"
...
 ;
+$ELSEIF.MaccVersion %c_nonco2_macc_version% == "PBL_2022"
+* PBL_2022 MACs are discretized in steps of 20 $/tC
+* PBL_2022 includes three scenarios, read the big file into p_all_abatparam_CH4 and subset based on c_nonco2_macc_scenario
+* Create the set here, as it's not really used anywhere else
+set macc_scens "MAC scenarios for PBL_SSP2_2022 MACCs" /Default, Optimistic, Pessimistic/;
+parameter p_all_abatparam_CH4(tall,all_regi,all_enty,macc_scens,steps)        "MAC costs for CH4 by source for different scenarios, c_nonco2_macc_scenario is chosen for p_abatparam_CH4"
+/
+$ondelim
+$include "./core/input/p_abatparam_SSP22022_CH4.cs4r"
```
LaviniaBaumstark, line 929: Do this files already exist in the current input data?

gabriel-abrahao:
> Great summary! Thanks a lot for your work, @gabriel-abrahao !
> 
> > @laurinks @nicobauer you might want to consider choosing the Optimistic and Pessimistic variants for some scenarios depending on the narrative. These variants were based on a Monte Carlo selection across case studies in the literature. In the case of the rollback, I don't think this will make a difference.
> 
> Interesting point! The only scenario narrative, where we might want to choose `Pessimistic`, is the high scenario (rollback), but there is likely no difference as you write since the CO2 price goes down to 0$. `Optimistic` could be a good fit for the Very Low with Limited Overshoot (VLLO) scenario. However, we already see a very strong CH4 reduction due to the diet changes in MAgPIE in that scenario, and the question would be if we want to add the optimistic assumptions for MACCs in REMIND on top of it. Add this point, I would say that we could simply test it in the next round of runs for the VLLO.
> 
> More generally, would you see a good reason to link the choice to SSPs instead of emission scenarios? For example, Optimistic in SSP1, Default in SSP2, and Pessimistic in SSP3. To judge this, I would need to understand where the differences come from. Is it some fundamental uncertainty (e.g. linked to physical processes)? Or is it an uncertainty related to the ability to successfully implement certain measures? In the first case, I would opt for always choosing the default (in the same way that we look at the Median Temperature outcome for all scenarios, not distinguishing across scenarios). In the latter case, I would opt for linking it to the SSPs or the specific ScenarioMIP scenarios.

Very good points, and I think I generally agree that for the ScenarioMIP exercise we should be using "Default" across the board. From what I understand from their paper, the different variants come from different technoeconomic/physical assumptions, not assumptions on policy or goodwill strength. So I like your parallel with the 50th percentile of the climate parametrizations, they shouldn't change between scenarios.

gabriel-abrahao:
> And two simple question for my understanding. First question: What is the exact meaning of `Abatement fraction`?
> 
>     1. Relative to the historical CEDS data for 2020: A CO2 price of 500 $ in year 2050 implies that only a fraction X of the value given by the historical emissions in 2020 is emitted in 2050.
> 
>     2. Relative to baseline emissions trajectory (agrees with historical CEDS for 2020, but differs thereafter): A CO2 price of 500 $ in year 2050 implies that only a fraction X of the value given by the baseline emissions is emitted in 2050.
> 
>     3. Relative to baseline value for each activity, say CH4 emissions per ton of produced rice: A CO2 price of 500 $ in year 2050 implies that only a fraction X of the baseline value is emitted for producing a ton of rice in 2050.
> 
>     4. Something else?
> 
> 
> Second question: Is there any inter-temporal relationship in the implementation? To give a concrete example, does it make a difference for CH4 emissions in 2050 whether (a) the CO2 price was 0$ until 2045 and 1000$ in 2050, or (b) the CO2 price was 1000$ from 2030 to 2050.

About the `Abatement fraction`, it's a combination of 2 and 3 for different sources. See #1967 for more details, but the general process is:

-  Derive, for all timesteps, a baseline emissions trajectory for each sector/gas. This will vary from applying emission factors to REMIND endogenous activity values (e.g. X tCH4 per ton of oil extracted) to applying more complex functions to GDP and population, to fully exogenous trajectories.
- The abatement fraction is then applied, and the MACC is the amount of those baseline emissions that is *abated*, so only a fraction 1-x will still be emitted. i.e. and Abatement fraction of 1 at a given price means that, at that price, nothing will be emitted from that sector.

About the intetemporal relationship, the answer is technically no to your example, there's no difference. That said, there are bounds implemented on the maximum year-to-year increase in abatement (that I consider updating BTW), in `p_macAbat_lim`. So, if the increase in price is too abrupt and leads to a large difference in abatement between timesteps, it will be supressed by this bound but will eventually happen. That limit is currently 5% of the maximum abatement fraction of that sector (the maximum value in the MACC) per year (a confusing number indeed).

Also keep in mind that the MACCs are time-dependent, that is, a C price of 500$/t in 2030 will not lead to the same abatement fraction as the same C price in 2050.

nicobauer:
Hi @gabriel-abrahao thanks a lot for the revisions and the PR!

I have mostly two comments: 
1. SSPs are also related to the development of techno-economics. Thus, matching the MACCs to SSPs makes sense.
2. I have some concerns about the MACCs as such that you show in Figure at the very top. I see that for fossil fuel extraction the MACCs are shaped very differently across regions. For example coal CH4 MACCS for India and EUR are very, very pessimistic, but in the US much more optimistic. This does not make so much sense I guess. It seems to me that the Indian MACC is much too pessimistic. There are other candidates. We can discus this after the REMIND meeting. We can also get Mathijs Harmsen on zoom asap. We may need to revise the PR regarding some of the MACCs [please be aware of this @LaviniaBaumstark ]
3. Following from this, how is the calibration for the NPi done exactly? We calibrate the baseline given state-of-policy in the base year. In the baseliine we assume that this policy prevails and is already associated with a certain mitigation level consistent with the MACC. Thus, if we roll back the policies we see higher emissions. Is that correct?

nicobauer:
@gabriel-abrahao I looked at the results now. The main difference seems to come from waste. ALso there I see lots of differences btw regional MACCs that make me wonder, whether they are correct. Some regions like EUR seem to have only very little mitigation potential, others are much higher. However, the sealing of and treatment of waste dumps is common practice. [Actually, one of my bike courses was blocked for the better part of 2024 because they upgraded the CH4 mitigation there. I could not go riding there over the summer :-/ ]

gabriel-abrahao:
> Hi @gabriel-abrahao thanks a lot for the revisions and the PR!
> 
> I have mostly two comments:
> 
>     1. SSPs are also related to the development of techno-economics. Thus, matching the MACCs to SSPs makes sense.
> 
>     2. I have some concerns about the MACCs as such that you show in Figure at the very top. I see that for fossil fuel extraction the MACCs are shaped very differently across regions. For example coal CH4 MACCS for India and EUR are very, very pessimistic, but in the US much more optimistic. This does not make so much sense I guess. It seems to me that the Indian MACC is much too pessimistic. There are other candidates. We can discus this after the REMIND meeting. We can also get Mathijs Harmsen on zoom asap. We may need to revise the PR regarding some of the MACCs [please be aware of this @LaviniaBaumstark ]

All this might be related to how these MACCs were derived, they are relative to a specific IMAGE SSP2 scenario. I've thought of a couple more elegant and less SSP-specific ways to implement them, but they would take substantially more work. 

>     3. Following from this, how is the calibration for the NPi done exactly? We calibrate the baseline given state-of-policy in the base year. In the baseliine we assume that this policy prevails and is already associated with a certain mitigation level consistent with the MACC. Thus, if we roll back the policies we see higher emissions. Is that correct?

Yes, as far as I understand that's correct. I would still check though if the implementation of `p_macAbat_lim` that I mentioned to Laurin won't bug in case of a rollback, please keep me informed when you do the next round of tests for the rollback for me to check.

nicobauer:
OK, if we look at the MACCs for waste, I see that they are generally more pessimistic than the older Lukas version. Is this reasonable? Do you have any idea why that changed?

gabriel-abrahao:
@nicobauer 
> OK, if we look at the MACCs for waste (solide and liquid waste), I see that they are generally more pessimistic than the older Lukas version. Is this reasonable? Do you have any idea why that changed?

Well, it is reasonable in the sense that it was a result from the MACC construction itself. Mathijs [Harmsen et al. 2019](https://doi.org/10.1016/j.envsci.2019.05.013) says of the comparison between the Lucas and the new MACCs, for example:

>The new set of MAC curves represents comparable total non-CO2 maximum reduction potentials (MRPs), notably at higher carbon prices (Note that the maximum price in Lucas et al. = 1000 $(2000), which translates to 1266 $(2010)). The overall MRPs in the new set are 59% in 2050 and 75% in 2100, compared to 60% and 77% in the old set, respectively. Although the aggregated GHG reduction potential is similar to Lucas et al., the gas specific MRPs are quite different. 

>CH4 reduction is projected to be much lower in the new set (in 2050, new: 51%, old: 67%, in 2100, new: 68%, old: 79%). Notably, projected MRPs are currently lower for enteric fermentation, rice, coal and natural gas production. N2O reduction, however, is projected to be substantially higher. MRP, old vs new: in 2050: 44% vs 53%, in 2100: 52% vs 62%. (in 2050, new: 53%, old: 44%, in 2100, new: 62%, old: 52%), due to higher MRPs for fertilizer application, animal waste and domestic sewage. [...]

Note that he doesn't specifically mention CH4 waste, so there may be be some baseline-dependency playing a role. The discussion and supplement also doesn't seem to get into why the MRPs for CH4 are lower in general (and more expensive).

All in all, there are some things worth investigating, and especially talking to Mathijs about. But our implementation seems to be, at least in broad strokes, following the results of his paper.

nicobauer:
OK, I guess it would be useful to talk with him about some of the nuts-and-bolts. Some of the numbers in the waste sector seem not make much sense (at least for me). When do you have time next week?

PR #2027, opened 2025-03-07
Ensure that technology bounds are non decreasing

RahelMA:
## Purpose of this PR
After technology targets are met, they should stay at least at the same level.
E.g., EU wind offshore target:

Before:
image

After:
image

Still don't really understand the spike in 2060!

Overview on policy updates:
https://docs.google.com/document/d/1ZQrfXY6zuSqb9ieI6JhoO_AsCvH1oikk96kwPbAqVCM/edit?addon_store&tab=t.0

## Type of change

- [x] Bug fix 
- [ ] Refactoring
- [ ] New feature 
- [ ] Fundamental change

## Checklist:

- [ ] All automated model tests pass, executed after my final commit (`FAIL 0` in the output of `make test`)

### DIFF
--- a/modules/40_techpol/NPi2025/datainput.gms
+++ b/modules/40_techpol/NPi2025/datainput.gms
@@ -14,16 +14,12 @@ $offdelim
 $onlisting
 ;
 
-p40_TechBound(ttot,all_regi,te) = f40_TechBound(ttot,all_regi,"%cm_NPi_version%",te);
+*** ensure that technology bounds are not decreasing
+p40_TechBound(ttot,all_regi,te) = smax(ttot2$(ttot2.val le ttot.val) , f40_TechBound(ttot2,all_regi,"%cm_NPi_version%",te));
 
 *** windoffshore-todo: separate NDC targets for windon and windoff
 p40_TechBound(ttot,all_regi,"wind") = f40_TechBound(ttot,all_regi,"%cm_NPi_version%","wind");
 p40_ElecBioBound("2030",regi) = p40_TechBound("2030",regi,"bioigcc");
-
-*** ensure that technology bounds are not decreasing
-loop(ttot$(ttot.val > 2025),
-    p40_TechBound(ttot, all_regi, te) = max(p40_TechBound(ttot, all_regi, te), p40_TechBound(ttot-1, all_regi, te));
-);
 
 *** In scenarios with 2nd generation bioenergy technology phaseout,
 *** switch-off biomass capacity targets of NDC
--- a/modules/40_techpol/NPi2025/datainput.gms
+++ b/modules/40_techpol/NPi2025/datainput.gms
@@ -14,7 +14,8 @@ $offdelim
 $onlisting
 ;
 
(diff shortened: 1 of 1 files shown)


modules/40_techpol/NPi2025/datainput.gms
laurinks, line 1: For better readability and easier debugging, I would suggest to adapt the code as follows: Instead of the suggested loop, I would replace the row 

`p40_TechBound(ttot,all_regi,te) = f40_TechBound(ttot,all_regi,"%cm_NPi_version%",te);`

by 

`p40_TechBound(ttot,all_regi,te) = smax(ttot2$(ttot2.val le ttot.val) , f40_TechBound(ttot2,all_regi,"%cm_NPi_version%",te) );`

The result is the same but I find it easier to read (also because it is applied directly when p40_TechBound gets values assigned for the first time).

laurinks (changes requested):
Thanks for the changes, @RahelMA ! 

Please see my proposal for a different code formulation (same result).

I also have a general question: Are all technology bounds at the moment and in the future **lower bounds**? 

- If yes, I agree that it makes more sense to enforce the lower bound also in future time steps, leading to the maximum in the code. In this case, could be make this clear by renaming `p40_TechBound` to `p40_TechLowerBound` ? 
- If not, the maximum cannot be applied in this way because it would also affect upper bounds. A fictive example: Imagine EUR having an upper bound on fossil related technologies that is decreasing over time (say <= 100 in 2030 and <= 50 in 2050). Then taking the max over previous time steps would deactivate the <= 50 bound in 2050 and always keep the <= 100 bound instead. One would rather take a minimum in this case, I think, but you could better judge this with the various technology targets in mind.

modules/40_techpol/NPi2025/datainput.gms
lea-hayez: Thanks both for your suggestions.

Are all technology bounds at the moment and in the future lower bounds?
> At the moment yes, in the future maybe not. We have excluded coal-related technology bounds so far, but it would require a upper bound, as it is usually about lowering the capacity or even a phase out.

Regarding the peak in 2060, I unfortunately don't know the modules well enough to know where this would come from. Maybe linked to the lifetimes of wind parks? In 2060 it might be cheaper to not replace end-of-life parks because another techno got cheaper?

laurinks (approved):
Thanks! It looks good to me.

However, I would still like to insist on the fact that this will cause problems for upper bounds. I understand if there is no satisfactory short-term implementation fix for now.

Maybe add a comment to the code like

`*** this only makes sense for lower bounds and needs to be revised once upper bounds are introduced.`

RahelMA:
> Thanks! It looks good to me.
> 
> However, I would still like to insist on the fact that this will cause problems for upper bounds. I understand if there is no satisfactory short-term implementation fix for now.
> 
> Maybe add a comment to the code like
> 
> `*** this only makes sense for lower bounds and needs to be revised once upper bounds are introduced.`

Yes, maybe it makes sense to have upper bounds in a separate input file.

PR #2028, opened 2025-03-07
add REMIND-MAgPIE compatibility test, remove last traces of SSP2EU

orichters:
## Purpose of this PR

- remove last traces of `SSP2EU` in order not to confuse our magpie collagues
- add a check that check consistency across REMIND and MAgPIE:
  - population and GDP scenarios
  - if REMIND climate damages are on, then also MAgPIE needs climate damages
  - bioenergy phaseout is in both models simultaneously either on or off
- close https://github.com/remindmodel/remind/pull/2028

It does the following: In MAgPIE, there seems to be `cfg_mag$gms$c09_pop_scenario` and `...$c09_gdp_scenario` with those options:
```
SDP, SDP_EI, SDP_MC, SDP_RC, SSP1, SSP2, SP3, SSP4, SSP5, SSP2EU
```
In REMIND, we had `cfg_rem$gms$cm_GDPpopScen` with those options:
```
SDP_EI, SDP_MC, SDP_RC, SSP1, SSP2, SSP2EU, SSP3, SSP4, SSP5, SSP2IndiaMedium, SSP2IndiaHigh
```
First of all, I removed the last traces of `SSP2EU` as it is not in any input file anyhow. `grep -d skip SSPEU * */* */*/* */*/*/* */*/*/*/* */*/*/*/*/*` does not show any trace of it anymore, so it can be deleted.

This check checks whether if those things start with `SSP?` or `SDP_??`, then they must all be identical. So SSP2IndiaHigh in REMIND, SSP2EU in MAgPIE -> works. But `SDP_EI` in REMIND and `SDP_MC` in MAgPIE doesn't.

For bioenergy, I followed @merfort's suggestions here: https://github.com/remindmodel/development_issues/issues/521

For damages, I agreed with @piontek that currently all REMIND damage implementations need MAgPIE damages as well. In case this might change, I added an explanation how to exclude it from the check

At the moment, the ScenarioMIP coupled config is erroneous with this check, but @laurinks is aware and wants to fix it on the MAgPIE side.

## Type of change

- [x] New feature 

## Checklist:

- [x] My code follows the [coding etiquette](https://github.com/remindmodel/remind/blob/develop/main.gms#L80)
- [x] I performed a self-review of my own code
- [x] I explained my changes within the PR, particularly in hard-to-understand areas
- [x] I checked that the [in-code documentation](https://github.com/remindmodel/remind/blob/develop/main.gms#L120) is up-to-date
[...]

PR #2033, opened 2025-03-12
Automatically detect in prepare.R which variables + equations can be fixed

orichters:
## Purpose of this PR

- avoid the problem at its core that REMIND tries to fix variables or equations that are not declared, maybe because they were renamed or because a different realization was chosen
- instead of a hand-picked list of variables to be dropped depending on realization choices, just get the variables and equations that are declared from the code and only keep marginals, levels and fixings for the variables that are declared
- print which variables that do exist in path_gdx_ref but have been dropped as they do not exist in the current run
- print which variables couldn't not be fixed because they did not exist in path_gdx_ref. They might then still have non-zero values from `path_gdx`
- this lead to me finding [this problem where negative variables were not fixed at all](https://github.com/pik-piam/gms/pull/109)
- I removed all the manual renaming from prepare.R from old variable names that existed only long ago. The feature of renaming might still be used in the future, but I also adjusted it such that you only have to specify it once, not three times for a single renaming. Dropping stuff manually should not be necessary anymore.
- fixed an inconsistency in bioenergy phaseout and population/GDP settings detected by the [new tests](https://github.com/remindmodel/remind/pull/2028) in coupled test config file and ScenarioMIP as agreed with Laurin
- add `make test-coupled` log files to `.gitignore`
- example run: `/p/tmp/oliverr/remindmagpie/output/SSP2-PkBudg1000_2025-03-13_08.41.05`

## Type of change

- [x] Bugfix including a bigger refactoring of this part with minor new features

## Checklist:

- [x] `make test-coupled-slurm` works apart from a from settings in ScenarioMIP config is fixed in [this commit](https://github.com/remindmodel/remind/pull/2033/commits/99849dcadae030634f218174f545dc01268dc73a).


scripts/start/prepare.R
```diff
@@ -334,201 +347,31 @@ prepare <- function() {
...
-                            list(c("vm_taxrevimplEnergyBoundTax.FX", "!!vm_taxrevimplEnergyBoundTax.FX")))
+    file.copy("levs.gms", "fixings.gms", overwrite = TRUE)
+    # Replace fixings with level values
+    manipulateFile("fixings.gms", list(c(".L ", ".FX ")), fixed = TRUE)
 
     # Include fixings (levels) and marginals in full.gms at predefined position
     # in core/loop.gms.
-    full_manipulateThis <- c(full_manipulateThis,
+    full_manipulateThis <- c(
                              list(c("cb20150605readinpositionforlevelfile",
```
orichters, line 362: Where does this actually go? I couldn't find any spot where this part will be inserted.

orichters: I deleted it. It was removed here: https://github.com/remindmodel/remind/commit/a64823c01adf87d72abf789d16e6ac91861a604b

LaviniaBaumstark (approved):
that is nice, thank you!

PR #2038, opened 2025-03-17
add historic-reqs target to Makefile to easily restore renv packages to

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
See [this conversation](https://mattermost.pik-potsdam.de/rd3/pl/n6ez4jf6p3yffyb48tja4ijiuc).


fbenke-pik:
Not my area of expertise. I think @dklein-pik and @pascal-sauer would be better suited to review this PR.

pascal-sauer (approved):

dklein-pik:
I think this is very useful. How do I know which date tags are available?

pascal-sauer:
One way is to look here: https://github.com/pik-piam/lockfile-archive/tags

pascal-sauer:
But should be available daily, so just try and only check which are available if the one you need is not :)

dklein-pik (approved):
Great, thanks.

PR #2050, opened 2025-03-21
don't confuse checkSummations with more than 1 model

orichters:
## Purpose of this PR

- if not, it tries to calculate summation checks independently for REMIND and MAgPIE which makes no sense.

## Type of change

- [x] Bug fix 

## Checklist:

### DIFF
--- a/output.R
+++ b/output.R
@@ -163,7 +163,8 @@ if (! exists("output")) {
 if (! exists("outputdir")) {
   modulesNeedingMif <- c("compareScenarios2", "xlsx_IIASA", "policyCosts", "Ariadne_output",
                          "plot_compare_iterations", "varListHtml", "fixOnRef", "MAGICC7_AR6",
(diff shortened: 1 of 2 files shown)


laurinks (commented):
Thank you, @orichters! I don`t fully get it from the two lines that have changed, thus asking here for clarification. 

Let`s say that there is a summation check that partly consists of REMIND and partly of MAgPIE variables. In that case, the check is only applied if both a REMIND mif and a MAgPIE mif are available, right?

orichters:
- If you have a mif with data from REMIND and MAgPIE, the script until now did a separate check for each model, leading to failing summation checks. This is now fixed by assigning the same `model` to all data.
- If you have a mif that contains only REMIND data (standalone run), the summation checks will fail as well and this is not addressed by this PR, but I think that is also ok because this summation is not holding given the data you provide.

laurinks (approved):
Perfect, thanks!

PR #2057, opened 2025-03-25
Enable calculation of net-negative emissions across iterations

laurinks:
## Purpose of this PR

Currently, net-negative CO2 emissions are calculated within an iteration as the difference of gross CO2 emissions and CDR. Therefore, the net-negative emissions tax disincentivizes both further CO2 emission reductions and CDR beyond net-zero. 

This PR introduces the (non-default) option to calculate net-negative CO2 emissions across iterations, i.e. as the difference of gross CO2 emissions based on previous iterations and CDR of the current iteration. Thereby, gross emissions are a parameter (and no longer a variable) and only CDR is a variable in the net-negative emissions tax revenue. Consequently, the net-negative emissions tax only disincentivizes CDR beyond net-zero, but not further CO2 emission reductions. 

As of now, this updated calculation interferes with the carbon price adjustment algorithm in `45_carbonprice/functionalForm/postsolve.gms`, which leads to longer runtime or non-convergence. Therefore, it is not set as a default until this problem is solved.

Linked to https://github.com/remindmodel/development_issues/issues/493

## Type of change

- [x] New feature 
- [x] No change (default scenarios show no differences)

## Checklist:

* Runs with these changes are here: `/p/tmp/laurinko/netNegEmiTax/remind/` (new runs started on March 25 after merging develop into netNegEmiTax, see `compScen-comparison-newNNET-2025-03-26_15.49.43-H12.pdf`)

* Run time and convergence (as described above): 
image

image
image
image
image


laurinks:
@tabeado : I have not adapted the `modules/33_CDR/portfolio/equations.gms` but only added the warning. Why? 1) Leon is working on the endogenous adjustment of the CDR price slope after the peak, which will require re-adjustments here anyway (which I cannot foresee as of now). Thus, trying to avoid double work. 2) I am wondering if it would be cleaner to switch off the equation be default (instead of setting [p33_GDP_NetNeg_share](https://github.com/remindmodel/remind/blob/a5c5adfb8318b94bd47f9ee73ed1c71883b50fee/modules/33_CDR/portfolio/datainput.gms#L122) to 1) since regions sometimes spend more than 100% of their GDP (due to imports/exports). However, I admit that this is irrelevant in practice.

```
*' Limit spending on net negative emissions to a share of the region's GDP 
*' Warning: This needs to be adapted if cm_NetNegEmi_calculation = 1 is used.
q33_CDRspending(t,regi)$(t.val ge max(2035,cm_startyear))..
  v33_NetNegEmi_expenses(t,regi)
  =e=
  (1-cm_frac_NetNegEmi) * pm_taxCO2eqSum(t,regi) * vm_emiAllco2neg(t,regi)
;
```

modules/21_tax/on/equations.gms
```diff
@@ -304,21 +304,30 @@ q21_taxrevCCS(t,regi)$(t.val ge max(2010,cm_startyear))..
...
-*'  vm_emiALLco2neg and v21_emiAllco2neg_slack are defined as positive variables
+*'  Auxiliary calculation of net-negative CO2 emissions in the current iteration: 
+*'  vm_emiAllco2neg and v21_emiAllco2neg_slack are defined as positive variables
 *'  so as long as vm_emiAll is positive, v21_emiAllco2neg_slack adjusts so that sum is zero
 *'  if vm_emiAll is negative, in order to minimize tax v21_emiAllco2neg_slack becomes zero
 ***---------------------------------------------------------------------------
 q21_emiAllco2neg(t,regi)..
-  vm_emiALLco2neg(t,regi) =e= -vm_emiAll(t,regi,"co2") + v21_emiALLco2neg_slack(t,regi)
-;
+vm_emiAllco2neg(t,regi) =e= -vm_emiAll(t,regi,"co2") + v21_emiALLco2neg_slack(t,regi);
```
tabeado, line 321: does v21_emiALLco2neg_slack not throw an error due to LL being written in capital letters? If it doesn't I don't mind it

tabeado (approved):

tabeado:
Thank you so much for this great adjustment, especially all the time spent on detecting and solving the convergence issues!

tabeado:
>  2) I am wondering if it would be cleaner to switch off the equation be default (instead of setting [p33_GDP_NetNeg_share](https://github.com/remindmodel/remind/blob/a5c5adfb8318b94bd47f9ee73ed1c71883b50fee/modules/33_CDR/portfolio/datainput.gms#L122) to 1) since regions sometimes spend more than 100% of their GDP (due to imports/exports). However, I admit that this is irrelevant in practice.

thanks for pointing this out, I will discuss with Leon once he has made his adjustment. Regarding the 100% GDP limit, however, I wonder, whether we should have scenarios exceeding this? I would see it as a sanity bound. That's a question for a different context though

modules/21_tax/on/equations.gms
laurinks: It is not a problem but I have corrected it. Thanks for spotting. In any case better to have consistent notation.

PR #2068, opened 2025-03-31
turn on inconvPenalties on fossil solids in buildings

robertpietzcker:
## Purpose of this PR

Turn on inconvenience penalties on fossil solids use in buildings. These inconvenience costs directly (linearly) enter the welfare equation, thus their relative strength increases with GDP/cap because consumption only enters via a logarithm-like function. 

They represent the observed behavior that heating with coal is very inconvenient, and the richer one gets the less one wants to go down to the coal cellar and carry up coal to a stinking burner in one's home. 

It is ONLY turned on in non-calibration runs, as these inconvenience penalties make the calibration somewhat unstable. This should not be a problem to the validity of the calibration, as the main effect is to shift fossil solids from buildings to industry. This makes results more realistic, but overall FE use is not changed much - so the overall difference between calibration and normal model run is small. 

## Type of change

- [x] New feature 

## Checklist:

/p/projects/remind/users/robertp/REMIND_3px/2025-03-26-InconvPen/inconvPenPhasein

blue = new, red = old, (green: intermediate version with full inconv costs from 2005 onwards)

more biosolids in buildings: 
image

less fossil solids in buildings: 

image

overall final energy solids use has relatively little change: 

image

overall solids mix also has limited change: 

image

### DIFF
--- a/modules/02_welfare/ineqLognormal/datainput.gms
+++ b/modules/02_welfare/ineqLognormal/datainput.gms
@@ -11,7 +11,10 @@ $if %cm_less_TS% == "on"  pm_welf("2060") = 0.9;
 
 *RP* 2012-03-06: Inconvenience costs on seprod
 $IFTHEN.INCONV %cm_INCONV_PENALTY% == "on"
-p02_inconvpen_lap(ttot,regi,"coaltr")$(ttot.val ge 2005)      = sm_D2005_2_D2017 * 0;    !! for the moment, set InconvPen for fesos to ezero to make calibrations more stable
+p02_inconvpen_lap(ttot,regi,"coaltr")$(ttot.val ge 2005)      = sm_D2005_2_D2017 * 0;    !! In dollar per GJ seprod at 1.000$/cap GDP, or 10$/GJ at 10.000$_GDP/cap
+p02_inconvpen_lap(ttot,regi,"coaltr")$( (ttot.val ge 2005) AND (ttot.val le 2025) ) = 0.33 * p02_inconvpen_lap("2050",regi,"coaltr");  !! phase-in to decrease jump from historic 2020 fesob|foss to model 2025 results
+p02_inconvpen_lap("2030",regi,"coaltr") = 0.66 * p02_inconvpen_lap("2050",regi,"coaltr");  !! phase-in to decrease jump from historic 2020 fesob|foss to model 2025 results
+
 p02_inconvpen_lap(ttot,regi,"biotr")$(ttot.val ge 2005)       = sm_D2005_2_D2017 * 1.0;   
 p02_inconvpen_lap(ttot,regi,"biotrmod")$(ttot.val ge 2005)    = sm_D2005_2_D2017 * 0;    !! for the moment, set InconvPen for fesos to ezero to make calibrations more stable
 *' Transformation of coal to liquids/gases/H2 brings local pollution, which is less accepted at higher incomes -> use the inconvenience cost channel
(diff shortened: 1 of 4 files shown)


LaviniaBaumstark (approved):

robinhasse:
The results do look better at first sight. I'm a bit hesitant about introducing further inconsistencies between calibration and scenario runs though. This change makes solids more expensive (in welfare therms) overall. So I'd expect systematic deviations from the calibration towards lower solids use in NPi runs. In your plot, deviations for total solids do look small but not negligible. Have you check by how much `FE|Buildings|Solids` has dropped? Tagging @ricardarosemann here, as she has worked a lot around these issues recently.
Also it seems we are fiddling around a lot with the solids split even though we have this information in EDGE-B but don't use it.

modules/02_welfare/ineqLognormal/datainput.gms
```diff
@@ -11,7 +11,10 @@ $if %cm_less_TS% == "on"  pm_welf("2060") = 0.9;
 
 *RP* 2012-03-06: Inconvenience costs on seprod
 $IFTHEN.INCONV %cm_INCONV_PENALTY% == "on"
-p02_inconvpen_lap(ttot,regi,"coaltr")$(ttot.val ge 2005)      = sm_D2005_2_D2017 * 0;    !! for the moment, set InconvPen for fesos to ezero to make calibrations more stable
+p02_inconvpen_lap(ttot,regi,"coaltr")$(ttot.val ge 2005)      = sm_D2005_2_D2017 * 0;    !! In dollar per GJ seprod at 1.000$/cap GDP, or 10$/GJ at 10.000$_GDP/cap
```
ricardarosemann, line 14: I'm a bit confused. With this line, ```p02_inconvpen_lap``` is still zero, right? How does this switch on the inconvenience penalty?

modules/02_welfare/utilitarian/datainput.gms
```diff
@@ -11,7 +11,10 @@ pm_welf("2060") = 0.9;
 
 *RP* 2012-03-06: Inconvenience costs on seprod
 $IFTHEN.INCONV %cm_INCONV_PENALTY% == "on"
-p02_inconvpen_lap(ttot,regi,"coaltr")$(ttot.val ge 2005)      = sm_D2005_2_D2017 * 0;   !! In dollar per GJ seprod at 1.000$/cap GDP, or 10$/GJ at 10.000$_GDP/cap
+p02_inconvpen_lap(ttot,regi,"coaltr")$(ttot.val ge 2005)      = sm_D2005_2_D2017 * 0;    !! In dollar per GJ seprod at 1.000$/cap GDP, or 10$/GJ at 10.000$_GDP/cap
```
ricardarosemann, line 14: See above ...

robertpietzcker:
@robinhasse I agree that in the mid-term, the coupling between EDGE-B and REMIND should become closer with respect to the fossil/biomass split. 

In the short-term, I clearly prefer an NPi with slightly lower solids in buildings compared the calibration over an NPi where the US and Japan suddenly start building new private coal furnaces, and world-wide coal heating roughly doubles vs today.

ricardarosemann:
Thanks for tagging. I'm a bit concerned that these changes interact with my recent PR #2067, I assume that this was not included in your tests?

Moreover, I share @robinhasse concerns, I'd also expect differences between the calibration and scenario runs. In general, we have bounds and penalties especially on FE solids in quite a few different ways and locations, which makes it difficult to keep an overview of what actually influences the model behavior here. Of course it makes sense to suppress coal heating in the future, but maybe we should take a closer look at all the relevant equations and bounds in the code and come up with a comprehensive way to handle this issue.

PR #2114, opened 2025-06-02
Nuclear: reduce investment costs in SSP1, reduce NDC target for IND

lecfab:
## Purpose of this PR

In ScenarioMIP, we don't want nuclear to disappear fully in SSP1:
- https://github.com/remindmodel/development_issues/issues/557#issuecomment-2922633403
image
 
This PR:
- changes the factor from 1.7 to 1.3 of SSP1 with respect to SSP2
- reduces IND NDC targets from 72 to 17GW (following near-term analysis "yellow flag" value)

## Type of change

### Parts concerned

### Impact

## Checklist

## Further information (optional)


lecfab:
Thanks Laurin for the fast review.

> could you add a plot/explanation of the calculation to the PR? 

Indeed, as the previous and current NDC are both unrealistic for IND (and also the wording is very weak), it makes sense to lower it to the _highest plausible value_, which is written as 17.25 in the ScenarioMIP vetting.
I agree with you that the calculation of this value is not clear in the comments I added (because I don't know where it comes from 😬) .. is there a way to better describe how the calculation takes place, @RahelMA @pweigmann ?

> deactivating early retirement for Nuclear (except in EUR)

After looking into it, I think it would require a bit of testing to do it right. The reason is that there are a lot of switches and configurations for retirement, and I'm not sure how they interact.

On the other hand, it means we may use these existing switches: 
```
c_tech_earlyreti_rate  GLO.(..., tnrs 0.01), EUR_regi.(..., tnrs 0.15), ...
```
(1% retirement allowed at all times everywhere, and 15% in EUR for year < `c_earlyRetiValidYr=2035`)

laurinks:
> On the other hand, it means we may use these existing switches:
> 
> `c_tech_earlyreti_rate  GLO.(..., tnrs 0.01), EUR_regi.(..., tnrs 0.15), ...`
> 
> (1% retirement allowed at all times everywhere, and 15% in EUR for year < c_earlyRetiValidYr=2035)

Great! Indeed, this means we can directly set it in the config file. Would you recommend making these specifications only for ScenarioMIP or also in the REMIND default?

RahelMA (changes requested):
Thank you, Fabrice, for the nice implementation! For now, I think it is better than changing the input files! 

Only one thing: Is there a reason why you would assume a drop in nuclear after reaching the target in 2035? After a technology target is reached, we would stay at least at that level (that was our assumption in the NPi case). That's [how](https://github.com/remindmodel/remind/blob/develop/modules/40_techpol/NPi2025/datainput.gms#L19) we did it for NPi2025.

I'll check the description in mrremind, we need to make this more transparent!

RahelMA:
So, I just revised the description in mrremind https://github.com/pik-piam/mrremind/discussions/540 and it is actually described as calculated: 

regional: +40% ("operational" + "inactive (JPN)" + "construction") so for India it is the following: (7.55 operational + 4.77 construction) * 1.4= 17.248

@lecfab would you change something to make it better understandable?

This value is also very close to the National Electricity plan:

https://mattermost.pik-potsdam.de/rd3/pl/8mkmohzs9b84fpbjfghcsxbn7e
image(25)

lecfab:
> After a technology target is reached, we would stay at least at that level (that was our assumption in the NPi case). 

Didn't know, i added it =) this is a big conceptual change and it will add many equations, but i guess this is safe since it has been done for NPi. It would definitely remove the nuclear phase-out (though only ind CHA and IND), and maybe also the problem of decreasing hydro (https://github.com/remindmodel/development_issues/issues/580).

>> we may use these existing switches:
>> 
>> `c_tech_earlyreti_rate  GLO.(..., tnrs 0.01), EUR_regi.(..., tnrs 0.15), ...`
>> 
>> (1% retirement allowed at all times everywhere, and 15% in EUR for year < c_earlyRetiValidYr=2035)
>
> Great! Indeed, this means we can directly set it in the config file. Would you recommend making these specifications only for ScenarioMIP or also in the REMIND default?

Not sure, any thought @robertpietzcker ?
I think we should wait for tests with the reduced investment cost and sustained NDC in CHA/IND.

RahelMA:
> > After a technology target is reached, we would stay at least at that level (that was our assumption in the NPi case).
> 
> Didn't know, i added it =) this is a big conceptual change and it will add many equations, but i guess this is safe since it has been done for NPi. It would definitely remove the nuclear phase-out (though only ind CHA and IND), and maybe also the problem of decreasing hydro ([remindmodel/development_issues#580](https://github.com/remindmodel/development_issues/issues/580)).

Thank you! We should check in a compare scenario if changing to persistent targets has reasonable effects on the NDCs. Do you have a path to a compScen (I would like to double check)?

lecfab:
Checking SSP1 runs in `/p/projects/remind/runs/REMIND_2025_06_03/remind`.
This looks ok for now. I will open a new issue for the question of everlasting targets.

image

RahelMA (approved):
Looks good to me now!

PR #2323, opened 2026-03-27
Remove capacity factor flexibility in steel sector

clarabachorz:
## Purpose of this PR

Up to now, steel production could be ramped up or down by the model freely by implicitely changing the capacity factor (production was set to be less than `vm_capFac * vm_cap`). This means e.g. coal-based steel production can be scaled down (unrealistically) fast with rising carbon prices. Typically, steel plants always operate at high capacity factor as they are capital-intensive, so this parametrization is not ideal.

Production is now fixed to be equal to `vm_capFac * vm_cap` after 2020. Now, for more control over the phase-out of coal-based steel, we should use earlyReti switches. The default parameters work fine, but in a very ambitious scenario, we could consider allowing some early retirement in the steel sector, especially in Global North countries + China.

We also change the default lifetime of `bf` and `bof` to 35 years (from 20 previously). This corresponds to a BF-BOF plant undergoing one BF relining before retiring. While steel plants could technically be retired after 20yr already, this is not financially attractive in most cases. Combined with the other changes made to the steel module, this change means a bit more coal remains locked in, even in ambitious scenarios (see below). Very ambitious scenarios could switch on some measure of earlyReti for `bf` and `bof`.

## Type of change

### Parts concerned

### Impact

## Checklist


modules/29_CES_parameters/calibrate/bounds.gms
```diff
@@ -58,6 +58,29 @@ loop (pf_industry_relaxed_bounds_dyn37(in),
...
 
+
+*' For primary steel, the upper bound from 2025 is too restrictive given slight inconsistencies in input data
+*' This helps to accommodate these inconsistencies, though ECS remains difficult.
+$offOrder
+* vm_cesIO.up(t,regi,"ue_steel_primary")$(t.val gt 2005 AND NOT sameAs(regi, "ECS"))
+vm_cesIO.up(t,regi,"ue_steel_primary")$(t.val gt 2005)
+= max(
+  pm_cesdata(t-1,regi,"ue_steel_primary","quantity"),
+  (pm_cesdata(t,regi,"ue_steel_primary","quantity")
```
fschreyer, line 70: Can you comment in the code why the bound needs to depend on both the target in t and t-1? Just because it is different to the other bounds in this section.

modules/37_industry/subsectors/equations.gms
```diff
@@ -464,9 +464,9 @@ q37_mat2ue(t,regi,mat,in)$( ppfUePrc(in) ) ..
 ;
 
 ***------------------------------------------------------
-*' Definition of capacity constraints
+*' Definition of capacity constraints (historical and current)
```
fschreyer, line 467: Would be good to add a short explanation why the production is free to change historically but is fixed in the future.

fschreyer (changes requested):
Thanks for this change!

Do you have some before/after plots for H12 Npi and PkBug to see what changes?

clarabachorz:
Adding here a quick overview of how this changes H12 NPi and PkBudget runs.

## Steel production

- Expected: coal-based production (black and grey with CCS) is much higher in PkBudget runs between 2040-2050. The transition to H2-based steel in Europe is slower, which fixes [this issue](https://github.com/remindmodel/development_issues/issues/737).
- Expected but may need additional adjustments: Chinese steel production is much higher between 2040-2050 (from 600Mt to close to 900Mt). This is likely too high. Maybe we need to introduce some default early retirement in some regions for bf-bofs? @JakobBD please let me know what you think.
- Will likely change: there is a lot of BF-BOF-CCS now. This will likely change again with this PR limiting biosolids to 30% in industry: https://github.com/remindmodel/remind/pull/2335

### Short-term
image

### Long-term
image


## PE|Coal

Slight increase across all regions and scenarios (only in the short term to 2050)
image

modules/29_CES_parameters/calibrate/bounds.gms
clarabachorz: done

modules/37_industry/subsectors/equations.gms
clarabachorz: done (below)

clarabachorz:
I've now added the bf / bof lifetime change to this PR.

modules/29_CES_parameters/calibrate/bounds.gms
```diff
@@ -58,6 +58,22 @@ loop (pf_industry_relaxed_bounds_dyn37(in),
...
+= max(
+  pm_cesdata(t-1,regi,"ue_steel_primary","quantity"),
+  !! more flexible upper bound, can be set by the previous time step's data
+  (pm_cesdata(t,regi,"ue_steel_primary","quantity")
+  *(2.5 + max(0, (sm_CES_calibration_iteration - 1) / sm_tmp))
+    ))$( sm_CES_calibration_iteration le sm_tmp )
+  + INF$( sm_CES_calibration_iteration gt sm_tmp 
+  );
+$onOrder
+
```
JakobBD, line 76: Woah, nice! Looks like that took a while to find/fix.

Does this fix the calibration issue that @whitenightZhang also encountered for process-based chemicals? 
If so, it would be good to FYI @leonieschweiger about this.

JakobBD (approved):
Awesome! Will definitely be an improvement. Thanks for this!

modules/29_CES_parameters/calibrate/bounds.gms
clarabachorz: Thank you! It did :)
I am not sure (I sent this fix to @whitenightZhang when he was having issues and can't remember the outcome). But it will be a step in the right direction. Happy to check with @leonieschweiger if it's enough, or if we should adjust it further for chemicals.

JakobBD:
My apologies, but having a closer look, I am having second thoughts: 

The PR changes our messaging around steel drastically, as BF-BOF-CCS now is the dominant technology until 2050. 
Is that what we want? I am having doubts. 

I see further refinement capabilities:  

- We could soften the early retirement constraint by adding BF/BOF to earlyReti tech and setting some rates
- We could introduce adjustment costs both for CCS, but also for DRI. 
- We could experiment with extending the CCS tech lifetime as well, which would remove some of the CCS. 

I am sorry if this causes further work, but I'm really not sure if the current results are better than the previous. 
What do you think?

clarabachorz:
Maybe you missed this:

> Will likely change: there is a lot of BF-BOF-CCS now. This will likely change again with this PR limiting biosolids to 30% in industry: https://github.com/remindmodel/remind/pull/2335

I don't know to what extent this will make a difference, but it will reduce the amount of BF-BOF-CCS. 
Each of your three options would also make sense - I reckon either using earlyReti tech or adding adjustment cost is best. Using earlyReti tech is easier.

What I suggest: I will check how big a difference it makes to add the biomass fix. If we still think it's a lot of CCS, we can introduce some earlyReti.

My two cents on CCS in PkBudg750: I actually don't think it's unrealistic to have quite a bit of CCS in an ambitious scenario with high carbon prices. I know CCS in steel is difficult to scale up, but so is reaching a PkBudg750 .. that's a more philosophical question though!

clarabachorz:
## Overview of changes
@JakobBD , here are the runs with two changes:
1) Biomass limit to solids in indst (https://github.com/remindmodel/remind/pull/2335) - this doesn't affect the amount of BF-BOF-CCS.
2) bf and bof added to earlyReti technologies. I tried with a 25% and 50% lower early Reti rate than the default (the default allowed an even faster transition than before).

Overall, I would imagine you'd be happier with these results. Worth noting: in both cases, the transition to H2-DRI in EUR is even faster than before, so we probably need to add some adjustment costs to H2-DRI and CCS on top of my changes.

### Next steps
What I would need from you:
- Are you happy with these results? If yes, would you prefer 50pc or 25pc?
- Should I merge PR #2335 and this one, or would you rather keep them separate?

## CS2 results
### With 25% lower earlyReti rate

image

### With 50% lower earlyReti rate

image

JakobBD:
I don't understand what's going on: 

In the first plot you're comparing 75% defaultER (no suffix) to 0% dER (.1 suffix), and in the second plot 75 % dER  (no suffix) to 50 % dER (.1 suffix)? 

Which also means that in PkBudg750, EUR has more DRI in 2040 if less BF/BOF can be retired?? 

If you give your scenarios a custom suffix (first column in scenario_config), they are easier to distinguish. It's easiest to only do it for PkBudg (if you do it to NPi, you also have to change some later columns). 

**Important**: For a fair model, I think the bfcc tech should also be added there. Apologies if I didn't spell this out. This might change the results quite a bit. 

> We probably need to add some adjustment costs to H2-DRI and CCS on top of my changes.

Yes, I think so. Unfortunately, that requires quite some parameter fiddling. Especially the seed, since it has different units than the energy tech. let's talk. 

And just to double check: The bio limit has no effect, right? 

> #2335 [...] will reduce the amount of BF-BOF-CCS.

> #2335 [...] doesn't affect the amount of BF-BOF-CCS.

I don't have a preference on joint or separate PR's, but I'll comment my bio questions on the bio one as long as it's open.

JakobBD:
Guideline of what could be next steps from my point of view: 

- Sort out bio questions (other PR)
- Add bfcc to earlyReti
- Add idr, bfcc, idrcc to teAdj
- Fiddle parameters for adjustment costs (coeff and seed) until you're in the right ballpark (see an effect of changing either, but no complete choking of scale-up)
- Do six runs with roughly 2 earlyReti and 3 adjustment cost settings - and suffixes on PkBudg names (I trust your judgement on which ones make sense, based on the previous parameter fiddling)
- Post results here

core/input/generisdata_tech.prn
```diff
@@ -173,7 +173,7 @@ $ifthen.cm_subsec_model_steel "%cm_subsec_model_steel%" == "processes"
 tech_stat
 inco0              400.0       200.0       420.0       350.0
 omf                 0.06        0.06        0.06        0.06
```
lecfab: Should not the initial investment or the fixed maintenance costs increase to absorb the relining cost?

clarabachorz: Yes, it should. Thanks for catching this, it's now fixed.

LaviniaBaumstark:
Will this PR being merged for  the next release?

clarabachorz:
> Will this PR being merged for the next release?

I hope so. I'll know by the end of the week (is that too late?)

LaviniaBaumstark:
fine, I just checked all open PRs to make sure that nothing is overseen, thanks

core/input/generisdata_tech.prn
lecfab: If i see correctly, it's now set to 460 for bf and still 200 for bof. But the comment at the bottom of the file says 660?
Sorry if I'm missing the difference between several types of costs.

clarabachorz: no worries - BF+BOF should be inco0=660 total, and how its distributed between bf and bof doesn't matter.

core/input/generisdata_tech.prn
```diff
@@ -322,7 +322,11 @@ bf and bof:
     Figure shows 65 / 55 --> inco0 approx 700
     Agora KSV-Rechner: 38 € OPEX
     Woertler 2013 Exhibit 8: Components (no C&R) summed CAPEX 442 € (149 BF, 128 BOF, + Rest) --> approx 500 $
-    --> inco0 average is 600; distribution among bf and bof does not affect model, measurement of bf in tPI is negligible
+    --> inco0 average is 600 without relining;
+    Relining cost: IEAGHG 2013 states US$ 232 Mill => US$ 250 Mill today => $60/t.
```
fschreyer: What is Mill? How is the conversion Mill to $/t?

core/datainput.gms
```diff
@@ -1318,8 +1323,8 @@ $endif.cm_subsec_model_steel
   p_adj_coeff(ttot,regi,'oae_el')       = 0.8;
 $ifthen.cm_subsec_model_steel "%cm_subsec_model_steel%" == "processes"
 *** steel technologies
-  p_adj_coeff(ttot,regi,"bfcc")         = 1.0;
-  p_adj_coeff(ttot,regi,"idrcc")        = 1.0;
+  p_adj_coeff(ttot,regi,"bfcc")         = 4.0;
+  p_adj_coeff(ttot,regi,"idrcc")        = 4.0;
```
fschreyer, line 1327: Adjustment cost with this parameterization are now quite much higher than the standard for REMIND technologies (p_adj_coeff = 0.25 and p_adj_seed_te =1). Is BF-BOF-CC and IDR-CC so much harder to scale than, say, thermal power plants? I think, it is fine if this gives you more realistic trajectories for now, but ideally we should also review the relative scalability of technologies at some point.

fschreyer (approved):
Looks good to me. Some minor comments. Thanks for fixing this! :)

core/input/generisdata_tech.prn
clarabachorz: good you pointed this out. fixed!

core/datainput.gms
clarabachorz: BF-BOF-CC and IDR-CC currently look much harder to scale up. There are limited pilot projects, and overall, not much has changed since 2013. Happy to review this at another point if you think we should investigate further.

PR #2335, opened 2026-04-28
Add biosolids share limit to industry

clarabachorz:
## Purpose of this PR

The industry module does not currently set a limit on the amount of biosolids that can be used. However, in industry sectors that use solids (mostly cement and steel), biomass is only partly substitutable due to technical limitations. In the steel sector, this is around 30% based on current literature. We implement this in `q37_limitBioSolidsIndst` using `v37_shSolidsIndst`.

### Parts concerned

### Impact

## Checklist

## Further information

* Comparison of results:
FE in industry:
image

Co2 prices:
image


JakobBD:
How does this PR relate to [this bio-solids limit](https://github.com/clarabachorz/remind/blob/fcd3f93db3b143297da3623111f9a992139da417/modules/37_industry/subsectors/bounds.gms#L117-L125) and to [this issue](https://github.com/remindmodel/development_issues/issues/667)?

JakobBD:
To expand on my previous comment: 

**Existing bound**

There is already a limit of bio-solids in the industry sector. It's set to 25 % of the BAU (NPi) industry solids use. Which means that either of the two can become tighter: If industry solids in the PkBudg is reduced by more than 16 % compared with NPi, your new bound is tighter. Else it's the existing one. 

I find this confusing. I find it especially confusing that these two bounds are implemented in different places in the code with no reference to each other. 

I would be in favor of removing the existing bound and keeping yours, it seems to make more sense to me. But you'd have to double check cement BECCS (especially in PkBudg650 runs), because that was seemingly the reason to introduce that bound in the first place. 

If you keep it, make sure they're as close to each other as possible in the source code (possibly even joining them), and/or reference each other in the in-code doc.  

**Felix's issue**

Felix wants to get rid of the bio bound. He put a lot of thought and work into this. His arguments: 

> I don't think it is a good way to enforce these hard bounds as it forces to keep fossils in as long as there are solids. And for the CES function it is really hard to reduce solids to 0 if they are non-zero in the baseline run. I think it is especially important for fossil phase-out runs ( @lecfab).

As I wrote over there, I think a hard bound for bio in steel makes absolute sense. I lack knowledge for the other sectors, but don't see a strict reason for bounds there: It's mostly used as energy input, right? Except for some CHA chemicals feedstocks? Do energy density and achievable temperatures restrict bio use for energy inputs? If so, to what extent?

I also don't know if they want to try to remedy the performance issues.

Before you move further with this: If you haven't done so, please read the whole thread there, and coordinate with Felix (and Fabrice). Then comment on Felix's issue with your common conclusions, and possibly close it.

JakobBD:
Further point: 

Could you please either share a path to your cs2 pdf, or check and comment yourself on remindmodel/development_issues#419 whether your changes here have any effect on this.

clarabachorz:
Thanks. Here are some clarifications on the points raised above, now that @fschreyer and I have discussed the bound.

## On the existing bound
I have decided to remove it. A strict bound based on biomass substitution potential limits is more defensible (see section below). The results do not change much in the short term when replacing this old bound by the new one (cement BECCS was the cited justification for the old bound, but it is kept in check by the new bound). The new bound now better constrains biosolids usage in 2050 and beyond (potentially relevant for @lecfab).
See below the cs2 file for default develop runs (with the old bound) vs new runs (with only the new bound).

Cs2 paths:
-  `/p/tmp/claraba/remind/compScen-newBioBound_compWithOldScens-2026-05-27_10.00.39-H12.pdf`
- `/p/tmp/claraba/remind/compScen-newBioBound_compWithOldScens-2026-05-27_10.00.39-H12-to2050.pdf`

### Cement CCS (cited as the justification for the old bound) is still the same
image

### Solids biomass in Industry
image

## On using a hard bound for biosolids in Industry

@fschreyer and I discussed this. Felix is happy with a hard bound, as long as it can be justified across industry sectors (since the bound applies to steel, cement and chemicals). Cement is the other key sector that uses solids, and after quickly checking, substituting biomass is also limited to 20-30% according to [this study](https://www.sciencedirect.com/science/article/abs/pii/S1364032122004075). So I think a hard bound absolutely makes sense here.
This means we can close https://github.com/remindmodel/development_issues/issues/667.

## On https://github.com/remindmodel/development_issues/issues/419

Unfortunately, the new bound does not fix this:
image

modules/37_industry/subsectors/equations.gms
```diff
@@ -529,5 +529,16 @@ q37_emiCCPrc(t,regi,emiInd37)$(
     )
 ;
 
+***------------------------------------------------------
+*' Limit biosolids in industry (only for ETS - all sectors except otherInd)
+***------------------------------------------------------
+q37_limitBioSolidsIndst(t,regi,entyFe)$((t.val ge 2025) AND sameas(entyFe,"fesos"))..
+  v37_shSolidsIndst(t,regi)
+  *
+  sum((entySe,te)$se2fe(entySe,entyFe,te), vm_demFeSector_afterTax(t,regi,entySe,entyFe,"indst","ETS"))
```
fschreyer: I prefer line breaks in summation statements like
```
sum((entySe,te)$se2fe(entySe,entyFe,te), 
  vm_demFeSector_afterTax(t,regi,entySe,entyFe,"indst","ETS"))
...
```

fschreyer (changes requested):
Great! I think this is a good pragmatic and more transparent solution. I raised two things to check in the [issue](https://github.com/remindmodel/development_issues/issues/667). Maybe let's continue discussing there about details.

modules/37_industry/subsectors/equations.gms
clarabachorz: done!

LaviniaBaumstark:
When do you plan to merge this PR?

clarabachorz:
> When do you plan to merge this PR?

as soon as I have Felix's final review. Likely tomorrow

fschreyer (approved):

PR #2301, opened 2026-03-04
Include magpie Climate Change in NGFS-related runs

lecfab:
## Purpose of this PR

New default in MAgPIE uses climate change impacts. For NGFS, each scenario is now associated with its RCP, and Climate Change impact is taken into account.
This PR merges de NGFS switches into default.

## Type of change

*Indicate the items relevant for your PR by replacing* :white_medium_square: *with* :ballot_box_with_check:.\
*Do not delete any lines. This makes it easier to understand which areas are affected by your changes and which are not.*

### Parts concerned
- :white_medium_square: GAMS Code
- :white_medium_square: R-scripts
- :white_medium_square: Documentation (GAMS incode documentation, comments, tutorials)
- :white_medium_square: Input data / CES parameters
- :white_medium_square: Tests, CI/CD (continuous integration/deployment)
- :ballot_box_with_check: Configuration (switches in main.gms, default.cfg, and scenario_config*.csv files)
- :white_medium_square: Other (please give a description)

### Impact
- :white_medium_square: Bug fix
- :white_medium_square: Refactoring
- :white_medium_square: New feature
- :white_medium_square: Change of parameter values or input data (including CES parameters)
- :ballot_box_with_check: Minor change (default scenarios show only small differences)
- :white_medium_square: Fundamental change of results of default scenarios

## Checklist

*Do not delete any line. Leave **unfinished** elements unchecked so others know how far along you are.\
In the end all checkboxes must be ticked before you can merge*.

- [x] **I executed the automated model tests (`make test`) after my final commit and all tests pass (`FAIL 0`)**
- [x] **I adjusted the reporting in [`remind2`](https://github.com/pik-piam/remind2) if and where it was needed**
- [x] **I adjusted the madrat packages (mrremind and other packages involved) for input data generation if and where it was needed**
- [x] My code follows the [coding etiquette](https://github.com/remindmodel/remind/blob/develop/main.gms#L80)
- [x] I explained my changes within the PR, particularly in hard-to-understand areas
- [x] I checked that the [in-code documentation](https://github.com/remindmodel/remind/blob/develop/main.gms#L120) is up-to-date
- [x] I adjusted `forbiddenColumnNames` in [readCheckScenarioConfig.R](https://github.com/remindmodel/remind/blob/develop/scripts/start/readCheckScenarioConfig.R) in case the PR leads to deprecated switches
[...]

PR #2345, opened 2026-05-15
Standard settings NDC LTS

RahelMA:
## Purpose of this PR

The purpose of this PR is to correct the standard settings for the default `NDC` and the `NDC-LTS` scenario. It includes these central elements:

1. Differentiation of transport assumptions:

- NPi2025 = `Mix2ICEban`
- NDC = `Mix3ICEban`
- NDC-LTS = `Mix4ICEban`
as proposed by @robertpietzcker 

2. Correction of MAgPIE settings:

Background: In MAgPIE net zero targets are currently **NOT** covered, leading to implausible energy mitigation efforts.

The following two changes are proposed by @mishkos to account for the missing targets:

- start GHG pricing in 2030 (`no_ghgprices_land_until` = y2030)
- change rcp assumption from `SSP2|NDC|AR-natveg|nocc_hist|rcp4p5` to `SSP2|NDC|AR-natveg|cc|rcp2p6`

These changes lead to fundamental differences in **carbon pricing**, especially OAS needing less ambiguous/implausible carbon prices and GHG only changing swiftly.

Scenarios: NDC-LTS (old settings), NDC-LTS_new (new settings for transport + no_ghgprices_land_until), NDC-LTS_new_cl (all new proposed settings)
image

image

I kindly ask you for your thoughts regarding the assumption on SSP2|NDC|AR-natveg|cc|rcp2p6 @mishkos & @lecfab !

## Type of change

### Parts concerned
- :white_medium_square: GAMS Code
- :ballot_box_with_check: Configuration (switches in main.gms, default.cfg, and scenario_config*.csv files)

### Impact
- :white_medium_square: Minor change (default scenarios show only small differences)
- :ballot_box_with_check: Fundamental change of results of default scenarios

## Checklist

- [ ] **I executed the automated model tests (`make test`) after my final commit and all tests pass (`FAIL 0`)**
- [ ] **I adjusted the reporting in [`remind2`](https://github.com/pik-piam/remind2) if and where it was needed**
- [ ] **I adjusted the madrat packages (mrremind and other packages involved) for input data generation if and where it was needed**
- [ ] I updated the `CHANGELOG.md` [correctly](https://gitlab.pik-potsdam.de/rse/rsewiki/-/wikis/Standards-for-Writing-a-Changelog) (added, changed, fixed, removed, input data/calibration)

## Further information (optional)

* Runs with these changes are here: `/p/tmp/rahelma/cleaning`
* Comparison of results (what changes by this PR?): `/p/tmp/rahelma/cleaning/compScen-NewSettings-2026-05-15_09.26.03-REMIND-MAgPIE.pdf`


lecfab:
Waouh the results are super, thanks @RahelMA for digging into the OAS mystery.
> change rcp assumption from `SSP2|NDC|AR-natveg|nocc_hist|rcp4p5` to `SSP2|NDC|AR-natveg|cc|rcp2p6`

Why use cc instead of nocc? I understand it may activate some Magpie mechanics that help the issue here, but that introduces inconsistencies between scenarios (LTS has cc but Pk750 does not?)
See similar discussion that concluded in keeping nocc:
- https://github.com/remindmodel/remind/pull/2301

config/scenario_config.csv
```diff
@@ -1,7 +1,7 @@
...
-SSP2-NDC-LTS;1,AMT,2;;;;;;globallyOptimal;;;3;;;NDC;netZero;;;;;;EUR_regi 0.08, USA_regi 0.07, CHA_regi 0.07, CAZ_regi 0.07,  JPN_regi 0.07, GLO 0.06;transformative;;;;;;;;;;;;;;;;;;;;;;;;2;;;;;;;;;1;2030;;SSP2-NPi2025;SSP2-NPi2025;SSP2-NPi2025;SSP2-NDC-LTS: Long-term-strategy, this scenario follows the NDCs until 2035 and starts rescaling the carbon prices from 2040 to reach national net-zero pledges. 
+SSP2-NDC-LTS;1,AMT,2;;;;;;globallyOptimal;;;3;;;NDC;netZero;;;;;;EUR_regi 0.08, USA_regi 0.07, CHA_regi 0.07, CAZ_regi 0.07,  JPN_regi 0.07, GLO 0.06;transformative;;;;;;;;;;;;;;;;;;;;;;;;2;;;;;;;;Mix4ICEban;1;2030;;/p/tmp/rahelma/cleaning/output/C_SSP2-NPi2025_2026-05-12_16.43.17/fulldata.gdx;/p/tmp/rahelma/cleaning/output/C_SSP2-NPi2025_2026-05-12_16.43.17/fulldata.gdx;/p/tmp/rahelma/cleaning/output/C_SSP2-NPi2025_2026-05-12_16.43.17/fulldata.gdx;SSP2-NDC-LTS: Long-term-strategy, this scenario follows the NDCs until 2035 and starts rescaling the carbon prices from 2040 to reach national net-zero pledges.
```
lecfab: Absolute paths should not be in the default config

RahelMA:
> Waouh the results are super, thanks @RahelMA for digging into the OAS mystery.
> 
> > change rcp assumption from `SSP2|NDC|AR-natveg|nocc_hist|rcp4p5` to `SSP2|NDC|AR-natveg|cc|rcp2p6`
> 
> Why use cc instead of nocc? I understand it may activate some Magpie mechanics that help the issue here, but that introduces inconsistencies between scenarios (LTS has cc but Pk750 does not?) See similar discussion that concluded in keeping nocc:
> 
>     * [Include magpie Climate Change in NGFS-related runs #2301](https://github.com/remindmodel/remind/pull/2301)

Great, that was exactly the discussion I was looking for! The difference is relatively small, and I see the point of consistency. What do you think, @mishkos, could we also stay with `nocc`?

mishkos:
> > Waouh the results are super, thanks @RahelMA for digging into the OAS mystery.
> > > change rcp assumption from `SSP2|NDC|AR-natveg|nocc_hist|rcp4p5` to `SSP2|NDC|AR-natveg|cc|rcp2p6`
> > 
> > 
> > Why use cc instead of nocc? I understand it may activate some Magpie mechanics that help the issue here, but that introduces inconsistencies between scenarios (LTS has cc but Pk750 does not?) See similar discussion that concluded in keeping nocc:
> > ```
> > * [Include magpie Climate Change in NGFS-related runs #2301](https://github.com/remindmodel/remind/pull/2301)
> > ```
> 
> Great, that was exactly the discussion I was looking for! The difference is relatively small, and I see the point of consistency. What do you think, @mishkos, could we also stay with `nocc`?

yes, we should stick to `nocc`, because of consistency with other coupled runs. If a project directly asks for LU climate impacts, we can include them. For general discussion whether we can have `cc` as default for coupling (it's default in MAgPIE already), best to discus in PIAM meetings.

config/scenario_config_magpie.csv
```diff
@@ -8,6 +8,7 @@ SSP2_lowEn-NPi2025;0;;;SSP2|NPI|AR-natveg|nocc_hist;y2150;1;;;0;;;0;const2030;73
 SSP2_lowEn-PkBudg1000;0;;;SSP2|NDC|AR-natveg|nocc_hist;y2030;1;2025;2050;0.01;2025;2050;0.1;const2030;734;;0;0;2030;2030;2030
 SSP2_lowEn-PkBudg650;0;;;SSP2|NDC|AR-natveg|nocc_hist;y2030;1;2025;2050;0.03;2025;2050;0.3;const2030;734;;0;0;2030;2030;2030
 SSP2-EcBudg500;1;;;SSP2|NDC|AR-plant|nocc_hist;y2030;1;2050;2070;0.02;2050;2070;0.2;const2030;734;;0;0;2050;2040;2050
+SSP2-NDC-LTS;1;;;SSP2|NDC|AR-natveg|cc|rcp2p6;y2030;1;;;0;;;0;const2030;734;;0;1;2030;2030;2030
```
mishkos: to revert to `nocc_hist`

mishkos (changes requested):
to revert to `nocc_hist`

robertpietzcker:
I don't have access to the cluster, thus can't take a look at the compScen. 

at first glance, the EUR carbon price of 300 \$ \/ tCO2 to reach the LTS (= climate neutrality in 2050) doesn't fit to my own runs, where we need ~500\$\/tCO2.

lea-hayez:
@fschreyer @RahelMA could we look a bit into the NDC carbon prices in CAZ? The region is doing better in NDC scenario than in the NDC-LTS (both NDC and NDC_new scenarios - so not linked to Rahel's changes), any idea why?

config/scenario_config.csv
RahelMA: Thanks for noticing; I removed it!

config/scenario_config_magpie.csv
RahelMA: I changed it!

RahelMA:
> I don't have access to the cluster, thus can't take a look at the compScen.
> 
> at first glance, the EUR carbon price of 300 $ / tCO2 to reach the LTS (= climate neutrality in 2050) doesn't fit to my own runs, where we need ~500$/tCO2.

That's right, together with Léa and Fabrice, we noticed that this limit function prevents the NDC-LTS scenario from reaching the NDC targets! I will try to fix it in another PR!

[min(max(0.1**p45_adjustExponent, p45_factorRescaleCO2Tax(t,regi)), max(2-iteration.val/15,1.01-iteration.val/10000));](https://github.com/remindmodel/remind/blob/408e7ad020e869a631053c17080ac44f4cda6a7c/modules/45_carbonprice/NDC/postsolve.gms#L48)

robertpietzcker:
> That's right, together with Léa and Fabrice, we noticed that this limit function prevents the NDC-LTS scenario from reaching the NDC targets! I will try to fix it in another PR!

I don't get it - why is this a problem in the NDC-LTS scenarios but not the NDC scenarios?


actually, I am not sure this is the reason - if I see correctly, EUR actually overachieves its 2050 netZero target in all those runs:

image
 
(variable is `Emi|GHG|w/ Bunkers|LULUCF national accounting` , which is the one that the LTS is formulated in afaik)


How is the NDC for EUR formulated - does it include the 2050 climate neutrality, or not?
if yes, this means that "NDC" and "NDC+LTS" is equivalent for EUR, right? 

then, changing from Mix3 to Mix4 in transport will clearly make the transport decarbonization easier, and changing the LULUCF setting from "NDC" to "NDC+LTS" will provide lots of negative emissions from LULUCF:  
image


In principle, with the current model, transport policy mixes (and possibly LULUCF, @mishkos ?) can only be changed for all regions at the same time. This means that in order to get reasonable results for a number of scenarios with increasing stringency, the stringency needs to increase in EACH region - otherwise you will get these effects like we see in EUR. 


@lea-hayez @RahelMA

RahelMA:
> > That's right, together with Léa and Fabrice, we noticed that this limit function prevents the NDC-LTS scenario from reaching the NDC targets! I will try to fix it in another PR!
> 
> I don't get it - why is this a problem in the NDC-LTS scenarios but not the NDC scenarios?

From my understanding, in NDC the target was already reached in only a few iterations, and NDC-LTS was cut too early:

```
/p/tmp/rahelma/cleaning/output/C_SSP2-NDC-LTS_new_2026-05-12_23.15.36 (exogen)$ dumpgdx fulldata.gdx p45_factorRescaleCO2Tax_iter 2030,CAZ
10  2030  CAZ  2.8380378709336
11  2030  CAZ  2.83467324734395
12  2030  CAZ  2.82348340877467
13  2030  CAZ  2.82241881276397
14  2030  CAZ  2.81985515045827
15  2030  CAZ  2.01266703874667
16  2030  CAZ  2.00888675131607
17  2030  CAZ  2.00696553094752
18  2030  CAZ  2.01340612345111
```

```
/p/tmp/rahelma/cleaning/output/C_SSP2-NDC-LTS_new_2026-05-12_23.15.36 (exogen)$ dumpgdx fulldata.gdx p45_factorRescaleCO2TaxLtd_iter 2030,CAZ
10  2030  CAZ  1.33333333333333
11  2030  CAZ  1.26666666666667
12  2030  CAZ  1.2
13  2030  CAZ  1.13333333333333
14  2030  CAZ  1.06666666666667
15  2030  CAZ  1.0085
16  2030  CAZ  1.0084
17  2030  CAZ  1.0083
18  2030  CAZ  1.0082
```

so `p45_factorRescaleCO2Tax` would still like to increase the carbon price but is limited by `p45_factorRescaleCO2TaxLtd` as this  `min(max(0.1**p45_adjustExponent, p45_factorRescaleCO2Tax(t,regi)), max(2-iteration.val/15,1.01-iteration.val/10000)); `is too restrictive and already cuts strongly after iter 15

mishkos (approved):

RahelMA:
After our last validation meeting, I tested to allow LULUCF emission pricing only after 2040 and also tried the option to start the scenario NDC-LTS only in 2040. As this would also align with the scenario narrative I have a strong preference to change our AMT policy NDC-LTS to NDC-LTS_2040 as in the scenario here: `/p/tmp/rahelma/NDCupdate/standardSettings/compScen-AMT_policy-2026-05-28_09.19.39-REMIND-MAgPIE.pdf`

image

image


```
Changes in the rows:
~ SSP2-NDC-LTS:
    cm_EDGEtr_scen: Mix2ICEban -> Mix4ICEban (default: Mix2ICEban)
    cm_startyear: 2030 -> 2040 (default: 2005)
    path_gdx_ref: SSP2-NPi2025 -> SSP2-NDC
    path_gdx_bau: SSP2-NPi2025 -> SSP2-NDC
    path_gdx_refpolicycost: SSP2-NPi2025 -> SSP2-NDC
~ SSP2-NDC:
    cm_EDGEtr_scen: Mix2ICEban -> Mix3ICEban (default: Mix2ICEban)

$allwarnings
```

lea-hayez (approved):

lea-hayez:
For reference, we based the NDC-LTS standard run on the protocol defined in the ELEVATE project, WP6.3 2026 :

_All modelling teams should update their NDC-LTS scenario and include the recently submitted 2035 targets (NDCs 3.0). The scenario should first follow a trajectory towards 2030 targets, subsequently towards 2035 targets, and then to LTS. Current policies should be included as a baseline._ 

_The descriptions and absolute emission levels for countries and regions calculated from these targets can be found in the enclosed excel-file ‘ELEVATE T6.3 Scenario Protocol NDC and LTS information.xlsx’. Additionally, this file includes detailed information on the NDCs for major emitters in line with the Climate Policy Modelling Protocol. If a region has a conditional NDC, that NDC should be implemented._

robertpietzcker:
thanks for the update, @RahelMA 

I have a few observations and questions that you could mabye comment on: 

1) I see that EUR now has much lower CO2 prices in LTS2040 than in LTS. Given that also emissions are higher in LTS2040, do you have any explanation why this is the case? was the target strongly overachieved in the LTS scenario, but the convergence algorithm stopped the model before the carbon prices could be brought down? if yes: what can we do to ensure that the run is only taken as converged when the target is roughly met? (otherwise we may have wildly fluctuating results for the same scenario setting, depending on pure chance if the run terminated so early that the targets are not met)

1b) why does the EUR carbon price jump to very high levels and then stay constant after 2035? do you have the 2035 NDC in that drives the price? if yes: in our previous analysis, we estimated that the 2035 NDC for EU is super super weak - so I don't understand how it can drive the carbon price level like that (if I am not misinterpreting what I see) 

2) why do REF, MEA and CHA have these funny 2030 carbon price bumps even in NDC? it looks a bit weird - why would we build our main scenario such that these countries do a climate policy reversal in 2035? 
even if the NDC2035 is potentially weaker than the NDC2030, I think it would look much more plausible to simply keep the carbon price at 2030 levels. 

3) Looking at REF, I am now confused which scenario cascades off of which one. 
I previously thought the cascade would be:
NDC:                 NPi2025 -> from 2030 on: NDC
NDC-LTS:          NPi2025 -> from 2030 on: NDC ->  from 2035 onwards NDC-LTS
NDC-LTS-2040: NPi2025 -> from 2030 on: NDC -> from 2040 onwards NDC-LTS

but then I see that the carbon price in NDC-LTS in REF is on NPi level in 2030, not on NDC level? so is that the difference - the NDC-LTS fixed on NPi and starts in 2030, while the NDC-LTS-2040 is fixed on NDC and starts only in 2040? 


4) I think the 2030 targets for CAZ should be adjusted - my understanding from Elmar's comments at the discussion yesterday is that targets should only be implemented if they don't lead to completely unrealistic carbon prices, and I would say 700-1000 S/tCO2 in 2035 is completely unrealistic. 

5) I am surprised by the stringency of the OAS LTS - all OAS countries have net-zero pledges by 2070? 

6) (mainly understanding:) MEA and SSA have relative reduction targets for 2050 as their LTS, and once those relative reductions are reached, carbon prices are kept constant?

lecfab:
A few reactions to [Robert's message](https://github.com/remindmodel/remind/pull/2345#issuecomment-4564570489): 

2. Tagging @Renato-Rodrigues, who also suggested (during @fschreyer's ETL presentation) that carbon prices for NDC should go up in 2030 with a realism limit, and then **not decrease** in 2035.

3. Just adding that LTS are reached with a markup carbon price increasing linearly between an LTSstartYr and the target year. The start year is a switch, and I believe in these runs the first non-zero markup is defined as 2040.
Still weird that REF is ignoring NDCs in the NDC-LTS run...

5. For OAS, we assume the following LTS: 2055 GHG emissions are 86% lower than in 2025. The coverage of 86% corresponds to pledge-less countries or excluded gas species. We can re-assess whether this is correct (Coverage shares are calculated using PBL's Net-Zero Calculator based on https://zerotracker.net/).
The current CO2 price is already a big improvement from previous version (it went up to $3000 😆) due to land emissions updates I believe, @RahelMA @mishkos

RahelMA:
@robertpietzcker thank you for your detailed assessment! That is very helpful.

This PR is only meant to improve some already existing standard settings in transport and LULUCF pricing. I would like to take up your points in other PRs/issues:


> A few reactions to [Robert's message](https://github.com/remindmodel/remind/pull/2345#issuecomment-4564570489):
> 
>     2. Tagging @Renato-Rodrigues, who also suggested (during @fschreyer's ETL presentation) that carbon prices for NDC should go up in 2030 with a realism limit, and then **not decrease** in 2035.

We applied a 2030 realism limit with the carbon price caps motivated by maximum emission reduction in industry according to ScenarioMip. For regions where the emission targets in 2035 are less ambitious due to higher growth assumptions, I, personally, think it is helpful to reflect and not transfer a 2030 ambition level that is beyond the regional pledges. 

We will discuss this issue in our next ICP meeting and come back to you!

>     3. Just adding that LTS are reached with a markup carbon price increasing linearly between an LTSstartYr and the target year. The start year is a switch, and I believe in these runs the first non-zero markup is defined as 2040.

Do I understand you correctly that you would advise changing the markup increase from linear to exponential? If yes, then we can make some test runs with different increases in modul_46 with @lecfab.

>        Still weird that REF is ignoring NDCs in the NDC-LTS run...

That's why we suggest starting from 2040 on the NDC as new default, than all NDC targets are captured and we are in line with the narrative:

_All modelling teams should update their NDC-LTS scenario and include the recently submitted 2035 targets (NDCs 3.0). The scenario should first follow a trajectory towards 2030 targets, subsequently towards 2035 targets, and then to LTS. Current policies should be included as a baseline._



>     4. For OAS, we assume the following LTS: 2055 GHG emissions are 86% lower than in 2025. The coverage of 86% corresponds to pledge-less countries or excluded gas species. We can re-assess whether this is correct (Coverage shares are calculated using PBL's Net-Zero Calculator based on https://zerotracker.net/).
>        The current CO2 price is already a big improvement from previous version (it went up to $3000 😆) due to land emissions updates I believe, @RahelMA @mishkos

The critical point in my view is that the underlying assumption is that the relation between country emissions stays constant over time, which is hard to defend, especially in OAS with single high-growth countries. The idea is to do this calculation in the future on our own using downscaled data from NGFS. (@lea-hayez & @lecfab maybe I can help setting this up in the summer)

**OAS
Regional net-zero calculation**		
		
```
Selected countries	Afghanistan, Bangladesh, Brunei Darussalam, Bhutan, Fiji, Micronesia, Indonesia, Cambodia, Kiribati, South Korea, Laos, Sri Lanka, Maldives, Marshall Islands, Myanmar, Mongolia, Malaysia, Niue, Nepal, Nauru, Pakistan, Philippines, Palau, Papua New Guinea, North Korea, Singapore, Solomon Islands, Thailand, Timor-Leste, Tonga, Tuvalu, Vietnam, Vanuatu, Samoa	
Total emissions	5329640000	Mt CO2eq
Emissions covered by targets	4582889299	Mt CO2eq
Share of emissions covered by targets	86%	
Average target year weighed by covered emissions	2054	
		
Result	This region reduces emissions by 86% by 2054	
```

LaviniaBaumstark:
when will you merge this PR? Will NDC_LTS become our default NDC scenario?

robertpietzcker:
I am slowly changing my mind on whether it is useful to fix the NDC-LTS to the NDC or not. 

I am now thinking that we sort of force the model to do it the hard way - force a 2035 target without the knowledge that there are later LTS targets that are aligned, and then use that as a starting point to finish the LTS. 

especially for the EU, this makes the NDC+LTS_2040 harder to achieve / have a higher carbon price in 2035 than if we went directly with the NDC+LTS targets. 

(I know, the plots look different, but the run simply wasn't converged - the emissions in 2050 are MUCH too low in the NDC-LTS )

so for getting reasonable trajectories for the LTS, I would actually say "let's not fix it to NDC, but solve both at the same time from 2030 onwards"

@RahelMA

robertpietzcker:
So I would actually propose to redo the normal NDC+LTS run and make sure it converges, and then look at the results if they aren't indeed better

robertpietzcker (approved):
I am fine with merging the transport and Magpie changes; 

I think the "should NDC+LTS start in 2030 or 2040" question requires further runs and comparisons, so a new issue/PR

PR #2362, opened 2026-06-05
Dynamically calculate calibration_resuts path

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
Change `calibration_results/.Rprofile_calibration_results` to use the current directory the file is located in for the `remind_repos` option to allow for moving project directories after the `calibration_results` directory has been set up.
See https://mattermost.pik-potsdam.de/rd3/pl/iiionod59bdquxpj6io7z87bur

### DIFF
--- a/scripts/utils/set-local-calibration.sh
+++ b/scripts/utils/set-local-calibration.sh
@@ -24,10 +24,7 @@ cp ../scripts/utils/set-local-calibration/pre-commit .git/hooks
 cp ../scripts/utils/set-local-calibration/post-commit .git/hooks
 chmod u+x .git/hooks/pre-commit .git/hooks/post-commit
 
-# create additional .Rprofile (sourced through default .Rprofile)
-echo -e "options(remind_repos = c(\n" \
-	"    getOption(\"remind_repos\"),\n" \
-	"    stats::setNames(list(x = NULL), \"$PWD\")))" \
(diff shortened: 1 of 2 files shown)


robertpietzcker:
Thanks! sounds great, I just can't comment on the code - I don't know/understand these scripts. so maybe add Lavinia or David as reviewer? 

this would lead to an automatic change, or one would need to specify something in the call?

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
> this would lead to an automatic change, or one would need to specify something in the call?

With this change, when `make set-local-calibration` is run, REMIND is set up in such a way that the directory `./calibration_results/.Rprofile_calibration_results` is in will be added to the `remind_repos` option at runtime, so it can be copied around without impact.

No change to existing `./calibration_results/` directories, but one can copy
```
cp ./scripts/utils/set-local-calibration/.Rprofile_calibration_results ./calibration_results/
```
and then it will work.

fbenke-pik:
Thanks for the PR, but as this is not my area of expertise, I'll leave this to David and Lavinia.

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
Pulling this up again, because it proves useful for archiving REMIND runs for publications.
So … bump @dklein-pik @LaviniaBaumstark

PR #2397, opened 2026-07-08
Fix industry biosolids historical shares

clarabachorz:
## Purpose of this PR

With the new industry biosolids bound, historical biosolid shares were no longer enforced to allow the model to optimize freely. This, however, leads to issues in regions such as IND, where biosolids are not scaled up unless forced to. As a result, industry emissions were too high in the short term (2025).
This PR now bounds the share of industry biosolids on the total of both ETS+ES sectors (rather than separately like it was previously done).

Fix for https://github.com/remindmodel/development_issues/issues/778.

cs2: `/p/tmp/claraba/remind/compScen-IndFix_NPi2025-2026-07-08_17.10.35-H12.pdf`
### Parts concerned

### Impact
- :ballot_box_with_check: Bug fix
- :white_medium_square: Minor change (default scenarios show only small differences)

## Checklist

## Further information

- red is current develop branch
- green is an intermediate test of removing the bound for India
- blue is the final fix.
### Industry emissions
image
### Biomass in SE solids
image


core/equations.gms
```diff
@@ -1140,7 +1140,7 @@ q_limitCapFeH2BI(t,regi,sector)$(SAMEAS(sector,"build") OR SAMEAS(sector,"indst"
 
 ***---------------------------------------------------------------------------
 *' Enforce historical data biomass share per carrier in sector final energy for transport and buildings (+- 2%)
-*' Exempt industry solids as these are covered by the equation q37_limitBioSolidsIndst
+*' Exempt industry solids as they are covered below
```
fschreyer, line 1143: Why do they need to be exempt here and we need another equation? Shouldn't applying the same bound to ETS and ES lead to very similar resutls?

fschreyer (approved):
Thanks for tackling this so fast. Looks reasonable. I wasn't aware anymore that this historical share had been removed for industry solids in the PR before. Was just wondering why we need new equations for that.

core/equations.gms
clarabachorz: ### On why industry solids need to be exempt from `q_shbiofe_up/lo`: 
The industry bound enforces a maximum of 30% biosolids in industry solids, **but only for the ETS sectors** (steel, cement, chemicals). OtherInd (ES) should have no such requirement. However, `q_shbiofe_up/lo` applies the historical bioshare target to ETS and ES markets independently. This led to a situation where, e.g. in the EU, both equations forced ~50% biosolids in ETS AND ES industry until 2020. The strict 30% ETS bound starts in 2025, creating some substantial jumps in solids bioshare in some regions. 

Before this PR, I decided the best solution was to exempt industry solids here and rely exclusively on the new bounds from 2005 onwards to constrain ETS industry solids to a maximum of 30% biosolids.

### Why we need a separate equation:
Since the 30% bound applies only to ETS sectors, I expect that the net historical bioshare is still broadly respected, with ES sectors compensating by using up a bigger share of biosolids. I had hoped the model would do this by itself, but this doesn't work in e.g. India, which led to the issue you raised. The new equations therefore bound the **combined** ETS+ES bioshare in industry solids, making sure otherInd (ES) makes up for the lower bio share in ETS sectors.

Nb. These aggregated equations across `emiMkt` could actually be applied to all sectors and FE, and replace `q_shbiofe_up/lo` entirely. I don't want to make such a change close to the new REMIND release, though. I could test it in a few days.

